PR #48, opened 2015-01-03
Change question logic on non-SSL port

lwindolf:
Idea is to bail out per default (with WARNINGS=off) this makes batch processing possible
as often testssl.sh hangs for minutes or endless on non-SSL ports.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1969,8 +1969,8 @@ parse_hn_port() {
 		$OPENSSL s_client -connect "$NODE:$PORT" $SNI </dev/null >/dev/null 2>&1 
 		if [ $? -ne 0 ]; then
 			boldln "$NODE:$PORT doesn't seem a TLS/SSL enabled server or it requires a certificate"; 
-			ignore_no_or_lame "Proceed (note that the results might look ok but they are nonsense) ? "
-			[ $? -ne 0 ] && exit 3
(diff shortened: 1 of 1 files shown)


drwetter:
Maybe I currently have a wek moment but your patch is askng whether to proceed and if answered yes it exits?

lwindolf:
I just retested it and works as intented. The question is "Do you want to exit [...]?" and if I answer "y" it exits. So I believe the change to be correct :-)

drwetter:
Don't know what your use case is/

I find REALLY dangerous to continue, thus a bare "enter" needs to exit for a normal use case. Thus I added a statemnent (e816e4877a500c6a9c94e1816cc47a0b626ff8e3) for WARNINGS=batch. You should be fine with this, pls check.

lwindolf:
My use case is just batch processing. But please do not feel forced to merge this. It's really ok not to!

Thanks for the great tool BTW!!!

drwetter:
Pls just try with the lastest 

  WARNINGS=batch ./testssl.sh 

and let me know.

Personally I don't do batch processing, but I know also from other requests that others do. Of course the tool should support it. 

PS: Thx for the flowers

lwindolf:
Current master used with "WARNINGS=batch" works perfectly. Thanks!

PR #54, opened 2015-01-28
Fix GOST handling in LibreSSL

nvsofts:
In LibreSSL, GOST support is in built-in engine.


drwetter:
Please keep in mind that LibreSSL is due to it's secure defaults not good for security testing. That might change if over time as I will do more and more in bash sockets.

drwetter:
Forgot one important thing: Thx!

Issue #11, opened 2014-07-06
IPv6


drwetter:
tests fail using IPv6 addresses

drwetter:
Only sockets do work with IPv6, so unless the whole testssl.sh is using this or openssl is supporting IPv6 there's nothing I could do.

drwetter:
Good news: 

The IPv6 patch from Fedora/RH works and out of the box newer FC/CentOS/RHEL openssl versions seems  to have that included. ( @feld: Any any idea about the IPV6 status of FreeBSD openssl binaries?). Anyway: There need to be new binaries (see https://github.com/PeterMosmans/openssl/issues/31).

As I wrote testssl.sh anticipatory for IPv6 (yes, I am one of the handful IPv6 users out there) the changes to testssl.sh were minor and I could w/ 20 minutes effort get a complete check of ``ipv6.google.com``  The only thing I am worrying now is how not to bother IPv4 only users with error messages.

Stay tuned!

drwetter:
Just committed the IPv6 patch.

As the log says: IPv6 is 80% working now. You either need a recent FC/RHEL/CentOS or a manually patched  openssl tree with the patch from Fedora http://pkgs.fedoraproject.org/cgit/openssl.git/plain/openssl-1.0.2a-ipv6-apps.patch . Also you need a to set `HAS_IPv6=true``.

Example:
```
prompt% OPENSSL=/data/tmp/openssl-1.0.2d.v6/apps/openssl HAS_IPv6=true ./testssl.sh -p -U ipv6.google.com           22:52:39

###########################################################
    testssl.sh       2.7dev from https://testssl.sh/dev/
    (feaef68 2015-09-26 22:44:33 -- 1.393)

      This program is free software. Distribution and 
             modification under GPLv2 permitted. 
      USAGE w/o ANY WARRANTY. USE IT AT YOUR OWN RISK!

       Please file bugs @ https://testssl.sh/bugs/

###########################################################

 Using "OpenSSL 1.0.2d 9 Jul 2015" [~145 ciphers] on
XXXX:/data/tmp/openssl-1.0.2d.v6/apps/openssl
 (built: "Sep 26 00:30:42 2015", platform: "linux-x86_64")


Testing now (2015-09-26 22:52) ---> [2a00:1450:4007:80e::200e]:443 (ipv6.google.com) <---

 rDNS ([2a00:1450:4007:80e::200e]): --
 Service detected:       HTTP

[...]

PR #61, opened 2015-02-17
Reformatted Code and added check for IPv6 Addresses

Rechi:


drwetter:
Thx first for the wrk! And sorry for my delayed response.  I am afraid I cannot work the next few weeks on it. 

WRT to IPv6 I am hestitant as it unfortunately doesn't make much sense yet -- see #11. Welcome to 2015 ;-)

Sorry for my for now disappointing answer. Better news in a few ;-)

PR #63, opened 2015-02-23
Check result of openssl when renegotiating

PeterMosmans:
Hi @drwetter , I don't know whether you'll agree with the formatting and text (please DO change the request to your liking), but I added an extra test for the return value of openssl when doing a renegotiation check. Sometimes servers *say* they do secure renegotiation when in fact they do not, thereby increasing false positives.

Cheers,

Peter


drwetter:
Thx Peter! I'll look into it. However pls do not expect anything from me within two weeks.

drwetter:
Hi @PeterMosmans,

sorry I forgot about that one.  Can you think of a condition that here the openssl command fails? (PS: magenta is currently used for client side problems).

Cheers, Dirk

drwetter:
PPS: the function has changed a bit, see 2aa82e51646c3f42282a48d4119f7ee6c5647682

drwetter:
closing this and keep the hint for the pipe status in mind

Issue #41, opened 2014-12-10
mx record


drwetter:
Automagically determine mx record of supplied domain / host and do a starttls scan on it if requested, e.g.  ./testssl.sh --mx <domain|host>

Rechi:
Should all MX-Records be check or just one?

drwetter:
better all

drwetter:
thx. Last polish see 77ed442

PR #66, opened 2015-02-24
Check MX Records (#41)

Rechi:


drwetter:
Thx! 
The first two minor changes I don't agree with (1st: exit is a must here, 2nd:  IP4+IP6 are global variables. 

The mx contribution is highly appreciated, thx

PR #72, opened 2015-03-11
Minor optimizations to reduce unnecessary forking

feld:
This is very nit-picky but in my non-scientific tests it sometimes saves half a second! :-)

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -270,7 +270,7 @@ wait_kill(){
 	pid=$1
 	maxsleep=$2
 	while true; do
-		if ! ps ax | grep -v grep | grep -q $pid; then
+		if ! ps $pid 2>&1 >/dev/null; then
 			return 0 	# didn't reach maxsleep yet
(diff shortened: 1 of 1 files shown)


PeterMosmans:
Whoops, I'm so used to Bash 4 that I failed to notice that this feature only works in version 4 upwards.
You're probably using an older version (I'm guessing the OSX one ?). So for maximum compatibility it's better to keep using the more-keystrokes type 1> 2>
Moreover the folks at http://wiki.bash-hackers.org/scripting/obsolete seem to favour the 1> 2> syntax over the redirect-all-in-one &> version.
Once again sorry for the confusion

# echo $BASH_VERSION
4.2.25(1)-release
# ps 1
  PID TTY      STAT   TIME COMMAND
    1 ?        Ss     0:02 /sbin/init
# ps 1&>/dev/null
#

drwetter:
Thx for the cleanup

PR #73, opened 2015-03-13
Shellcheck static analysis

feld:
I'm starting to push testssl.sh through shellcheck.sh's static analysis. The 2>&1 commit was actually wrong -- I had forgotten about the order issue.

The backticks to $(..) is really a style/modernization nitpick, but there are some other things that will be coming such as unquoted variables, misuse of variables in printf, etc. Ideally we'll have testssl.sh throwing no warnings when run through shellcheck and there will fewer issues to troubleshoot in the future. :-)

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1139,7 +1139,7 @@ server_defaults() {
 		SAN=$($OPENSSL x509 -in $HOSTCERT -noout -text | grep -A3 "Subject Alternative Name" | grep "DNS:" | \
 			sed -e 's/DNS://g' -e 's/ //g' -e 's/,/\n/g' -e 's/othername:<unsupported>//g')
 #                                                               ^^^ CACert
-		[ x"$SAN" != "x" ] && SAN=$(echo "$SAN" | sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/ /g) && outln " subjectAltName (SAN)         $SAN"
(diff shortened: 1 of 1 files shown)


feld:
Closing pull request until I work a few things out -- might just be FreeBSD / Linux differences -- we'll see.

feld:
This is now stable/reliable/working on FreeBSD and Linux from what I can tell.

drwetter:
Hi Mark,

static analysis sounds excellent! Also you get the trophy mentioning that you tested ;-)

But please ideally only one pull request per issue. I am probably not fine with the replacement of backticks, others I need to test, others are no-brainer to me.

Cheers, Dirk

drwetter:
Thx Mark!

For some reaon I inititially thought that backticks don't fork a subshell and are faster. But that's wrong. expr should be changed in the future by bash internal arithmetic functions like $(()) or let.

Cheers, Dirk

feld:
Sorry about the pull request mess. I have not yet taken the time to figure out why github wants to take every single commit and add it to an existing pull request. I suppose the solution is to create endless branches? *sigh*

drwetter:
never mind. That's fine for now. Happy about the quality of what you did.

drwetter:
Thx for the hint wrt shellcheck. Cool program! I did some minor improvements too but it's time consuming to weed out the stuff which doesn't make so much sense (mostly non-quoted variables which can't have a blank)

PR #75, opened 2015-03-18
Using square brackets in tr results in trying to match/replace them

feld:


drwetter:
Historical system V stuff, yes. Not needed anymore and breaks BSD

PR #76, opened 2015-03-18
Fix variable directly referenced in printf

feld:


drwetter:
Thx for the cleanup again

PR #79, opened 2015-03-22
Refactored major parts of code

PeterMosmans:
(new pull request, based on correct upstream/master branch)

Note that due to the refactoring of some status messages, the output will be slightly different (more verbose) than previous versions

Moved specific status messages to http_header()
Moved specific status messages to breach()
Moved specific status messages to ccs_injection()
Moved specific status messages to heartbleed()
Moved specific status messages to renego()
Moved specific status messages to crime()
Moved specific status messages to tls_poodle()
Moved specific status messages to freak()
Moved specific status messages to beast()

Added some more documentation for functions

Fixed typos in help

Created new function main:
This is the main function of testssl.sh
Refactored major part of the original main function

Created new function startup:
Parses the startup options

Created new function intialize_globals:
Initializes all used global variables

Created new function scanning_defaults:
Sets default scanning options when only one parameter (URI) is given

TODO: Refactor more/duplicate parts of functions

Note: For the new functions, fixed spaces (4) are used instead of tabs

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -388,6 +388,7 @@ runs_HTTP() {
 
 #problems not handled: chunked
 http_header() {
+    outln; pr_blue "--> Testing HTTP Header response"; outln "\n"
 	[ -z "$1" ] && url="/" || url="$1"
 	if [ $SNEAKY -eq 0 ] ; then
 		referer="Referer: http://google.com/"
@@ -1449,6 +1450,7 @@ rc4() {
 # is agnostic to the version of TLS/SSL, more: http://www.breachattack.com/
 # foreign referers are the important thing here!
 breach() {
+    outln; pr_blue "--> Testing for BREACH (HTTP compression) vulnerability"; outln "\n"
 	pr_bold " BREACH"; out " (CVE-2013-3587) =HTTP Compression  "
 	url="$1"
 	[ -z "$url" ] && url="/"
@@ -1954,6 +1956,7 @@ ok_ids(){
 ccs_injection(){
 	# see https://www.openssl.org/news/secadv_20140605.txt
 	# mainly adapted from Ramon de C Valle's C code from https://gist.github.com/rcvalle/71f4b027d61a78c42607
+    outln; pr_blue "--> Testing for CCS injection vulnerability"; outln "\n"
 	pr_bold " CCS "; out " (CVE-2014-0224), experimental        "
 
 	$OPENSSL s_client $STARTTLS -connect $NODEIP:$PORT &>$TMPFILE </dev/null
@@ -2052,6 +2055,7 @@ ccs_injection(){
 }
 
 heartbleed(){
+    outln; pr_blue "--> Testing for heartbleed vulnerability"; outln "\n"
 	pr_bold " Heartbleed\c"; out " (CVE-2014-0160)                "
 	# mainly adapted from https://gist.github.com/takeshixx/10107280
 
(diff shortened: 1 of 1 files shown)


drwetter:
First feedback: Just tested it on BSD. Doesn't work.

/usr/bin/getopt is not standardized and thus it doesn't work under BSD.

Alternative would be using the internal getopts from bash. But that has the catch of not supporting long style option. This is not what I want .

Only solution I see parsing the options manually. :-(

PeterMosmans:
Hmmm. As some of the options were long options I went for the GNU getopt variant. Not all shells have long options support for getopts.
I think that using getopt(s) is still an improvement over manually parsing (string comparing) options. Getopt parses `-abc` to `-a -b -c` for example , and `-d1 -d=1 -d 1` all mean the same thing.

One possiblity is to check whether the installed getopt is the GNU variant. If it is, use this syntax. If not, tell the user that long options are not allowed on that platform, hint them to install GNU getopt, and parse only the short options.

What do you think ? Once again, I really think that using getopt(s) is an improvement over string comparisons.

drwetter:
Sorry, more clearly: 

1) Gnu getopt is no option just because there's no GNU getopt under BSD/MacOSX.. There's /usr/bin/getopt though but with your patch the script bails out. Haven't understand the difference of BSD getopt in detail though.

2) bash internal getoptS would be compatible however it breaks the long style switches. And I don't want to break backwards -- now.

A different behavior under Linux vs. BSD is ugly and thus no option.

drwetter:
Hi Peter,

maybe this helps: http://wiki.bash-hackers.org/scripting/posparams

PeterMosmans:
Do I understand it correctly that you would like to keep the current string comparison routines for now ? With short as well asl long 'options' ?
If so, I could still keep parts of the refactored code from the original pull request, and swap out the getopt routine for the previous string comparison routine.

drwetter:
Hi Peter,

Am 28. März 2015 00:54:42 MEZ, schrieb Peter Mosmans <notifications@github.com>:
>Do I understand it correctly that you would like to keep the current
>string comparison routines for now ? With short as well asl long
>'options' ?

If you mean by current your change: yes. My historic code sucks ;)  as far as that is concerned. 

>If so, I could still keep parts of the refactored code from the
>original pull request, and swap out the getopt routine for the previous
>string comparison routine.

That indeed sounds like the best solution.  

Cheers,  Dirk

PeterMosmans:
@drwetter , thanks for your answers. As soon as I rewrite the code I'll create a new pull request, and close this one.

drwetter:
Am 03/29/2015 um 11:13 AM schrieb Peter Mosmans:
> @drwetter <https://github.com/drwetter> , thanks for your answers. As soon as I rewrite the
> code I'll create a new pull request, and close this one.

Well, thank *you*. cheers! Dirk

PeterMosmans:
Removed GNU getopt, all back to original string comparisons. @drwetter , could you please test this as thoroughly as you can ?

Thanks!

Peter

drwetter:
Am 7. April 2015 10:12:31 MESZ, schrieb Peter Mosmans <notifications@github.com>:
>Removed GNU getopt, all back to original string comparisons. @drwetter
>, could you please test this as thoroughly as you can ?

Thank you, Peter.  Will do!

Cheers,  Dirk

drwetter:
Thx Peter! cool...

PR #110, opened 2015-05-28
simplify life for OSX users running gnu's coreutils...

AntonioMeireles:
hi, 

attached patch gets things to work out of the box for users that have GNU's coreutils installed under OSX, via brew or whatever.

 ( && congrats for the great work!)

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1459,10 +1459,12 @@ server_defaults() {
 		now=$(date "+%s")
 		HTTP_TIME=$(awk -F': ' '/^date:/ { print $2 }  /^Date:/ { print $2 }' $TMPFILE)
 		if [ -n "$HTTP_TIME" ] ; then
-			case $SYSTEM in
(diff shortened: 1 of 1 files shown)


drwetter:
Hi Antonio,

is BSDs date not working anymore then?

Cheers, Dirk

AntonioMeireles:
no, that's not the issue. the issue is - right now BSD's date syntax is assumed by just looking at $SYSTEM which is misleading, as an user can very well have installed GNU's date (which as a != syntax). so, instead of relying on $SYSTEM for detection we test if a flag unique to gnu's coreutils date is available - if it is we use default, if not *then* we are using a BSD's box and ence we'll use bsd's syntax...

drwetter:
Thx @AntonioMeireles .

Indentation btw is always a tab, virtually 5 spaces. Under FreeBSD / MacOSX it's better to use vim (see lines in the bottom) or any other editor which can be set to this.

PR #111, opened 2015-05-29
FIX: show header when specifying -b | --banner | -v | --version

PeterMosmans:
Another addition: show git commit hash and summary instead of non-dynamic (RCS ?) data.
This probably helps when troubleshooting issues, as you can see which version people are using.

If you want I can split these two into two separate pull requests...


drwetter:
Hi Peter!

Thx -- but as you can see looking at the diffs -- the code changed a lot since your pull. eb3227c837efcdfa62a2d666e53e1f1e7fbe56ac  is a no brainer. The other ones: I just can't tell off the top of my head whether this applies still to the actual code.

Can't you check with the latest code or you want me to go through this?

Cheers! Dirk

PeterMosmans:
closed - see new pull request

PR #132, opened 2015-07-10
Fix list-reversal (and comment)

thomassa:
Use "tac" (which reverses things) rather than "sort -r" (which will
reverse a list only if it was sorted already).

PR #133, opened 2015-07-10
Fix list-reversal

thomassa:
Running an unsorted list through "sort -r" will not reverse it, but
using both "sort" and "sort -r" will give a forward/reversed pair.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1428,19 +1428,20 @@ read_dhbits_from_file() {
 
 server_preference() {
 	local list1="DES-CBC3-SHA:RC4-MD5:DES-CBC-SHA:RC4-SHA:AES128-SHA:AES128-SHA256:AES256-SHA:ECDHE-RSA-AES128-SHA:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA:DHE-DSS-AES128-SHA:DHE-DSS-AES128-SHA:DHE-DSS-AES256-SHA:ECDH-RSA-DES-CBC3-SHA:ECDH-RSA-AES128-SHA:ECDH-RSA-AES256-SHA:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:DHE-DSS-AES256-GCM-SHA384:AES256-SHA256"
+	local list_forward=$(echo $list1 | tr ':' '\n' | sort | tr '\n' ':' | sed -e 's/:$//')
 	outln;
 	pr_blue "--> Testing server preferences"; outln "\n"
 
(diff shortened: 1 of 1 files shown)


thomassa:
This version of the fix is more cumbersome than #132 but avoids use of `tac`.

drwetter:
Thx. But If I compare in https://github.com/thomassa/testssl.sh/commit/eaf1c97d47cd8fbc13b15360fe7a307461ad72bb my line 1443 and yours I do not see your point.

thomassa:
But look at my lines 1431 and 1436:  https://github.com/thomassa/testssl.sh/commit/eaf1c97d47cd8fbc13b15360fe7a307461ad72bb#diff-2b772c01e86977452d2db64b3031f913R1431
There I sort the list forward and use the forward-sorted list in the first connection to the server.
This means that the reverse-sorted list (used in the second connection) is the reverse of the list used in the first connection.

drwetter:
ok, got that now and see your point.

What rather should be done is using tac offline and working with a manual fwd list and backward list.

thomassa:
That would work, certainly... so long as you want to use a hardcoded list
`list1="DES-CBC3-SHA:RC4-MD5:...:AES256-SHA256"`
rather than
`list1=$($OPENSSL ciphers)`
or similar. I presume there's a reason for that, but I don't see it.

drwetter:
a) it's faster
b) I'd rather prefer to leave it up to a human to prepare those two lists instead of shuffling through bubble sort ;-) and friends or some random list, i.e. list1=$($OPENSSL ciphers) which is not reproducible.

Looking at  my line 1442 now: The "sort -r" was probably not what I intended

thomassa:
> a) it's faster

Enough to matter? I don't think the slow version produces a noticeable delay.
> b)  I'd rather prefer to leave it up to a human ... instead of `$($OPENSSL ciphers)` which is not reproducible.

Fair enough then.

> Looking at my line 1442 now: The "sort -r" was probably not what I intended

Right: it seems fairly clear that you wanted to reverse the list. (I presume the comment was mangled at some point in history by a global replace of "reverse" with "pr_reverse".)

drwetter:
Am 07/13/2015 um 12:33 PM schrieb thomassa:
>     a) it's faster
> 
> Enough to matter? I don't think the slow version produces a noticeable delay.

maybe

>     b) I'd rather prefer to leave it up to a human ... instead of |$($OPENSSL ciphers)| which
>     is not reproducible.
> 
> Fair enough then.
> 
>     Looking at my line 1442 now: The "sort -r" was probably not what I intended
> 
> Right: it seems fairly clear that you wanted to reverse the list. (I presume the comment was
> mangled at some point in history by a global replace of "reverse" with "pr_reverse".)

yup, that was a victim of search and replace as I renamed the print functions  ;-)

If you feel like using tac offline for reversing the first list just let me know otherwise I'll do
it on the next occasion.

Anyway: thx for the fruitful discussions!

drwetter:
Thx, @thomassa. 

Done in 1ae4b121c4271bb0e91b118ce2d9f38dbca5d125

PR #142, opened 2015-07-21
Fix syntax error near unexpected token on Mac OS X

jonorossi:
Mac OS X ships with bash 3, not 4 by default (3.2.57 on my 10.10.4 system). The bash case statement fallthrough and continue operators were added in bash 4. This patch removes the use of the operator that was [recently added](https://github.com/drwetter/testssl.sh/commit/b742c543584db71752a09e3362176692373b7278#diff-2b772c01e86977452d2db64b3031f913R3740).

```
$ ./testssl.sh example.com
./testssl.sh: line 3980: syntax error near unexpected token `;'
./testssl.sh: line 3980: `				 ;&'
```

Since the XMPP detection is quite different I considered separating it, but decided to just merge the two `$protocol == "xmpp"` checks inside the existing case rather than separating xmpp support out completely.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3975,20 +3975,22 @@ determine_service() {
 	else
 		protocol=$(echo "$1" | sed 's/s$//')     # strip trailing s in ftp(s), smtp(s), pop3(s), imap(s), ldap(s), telnet(s)
 		case "$protocol" in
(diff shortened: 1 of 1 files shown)


drwetter:
Thx!

Issue #31, opened 2014-11-18
Output JSON/HTML


drwetter:
Json / html would be fine. Target: version 2.2.

jonorossi:
@drwetter have you put any thought into how you'd like the implementation to work?

I see two real options here since testssl.sh is written in bash and not a language with a JSON library. Either append to a handcrafted JSON variable as the script runs paired with each existing print to stdout, or perform in-place JSON modification using a CLI tool like [json](http://trentm.com/json/) that you'd add new attributes as you went.

I've had a skim read through the whole script file and it feels like there should be a bit more structure around the scan results, which would allow the implementation of ratings (#100) to be a little more formal like ssllabs. However, I'm not a bash master and don't even know if bash can do what I'm thinking.

jonorossi:
I just found [json-bash](https://github.com/ingydotnet/json-bash), which can do things like the json CLI tool I just mentioned, however this is a single bash script rather than a node.js script. The way it allows you to represent the JSON structure in a path syntax with a value should make it pretty easy to implement.

drwetter:
Am 07/21/2015 um 08:02 AM schrieb Jonathon Rossi:
> @drwetter <https://github.com/drwetter> have you put any thought into how you'd like the
> implementation to work?

a bit. I was up to now more on the user side and was planning to see who other were
doing this.

> I see two real options here since testssl.sh is written in bash and not a language with a JSON
> library. Either append to a handcrafted JSON variable as the script runs paired with each
> existing print to stdout, or perform in-place JSON modification using a CLI tool like json
> <http://trentm.com/json/> that you'd add new attributes as you went.

The latter need node.js. That makes testssl.sh too depended.

I am more in favor of the first suggestion. Roughly I was thinking on first extending the output
functions. But any output I am generating so far is for humans to read and not really adaptable
to JSON. Then the idea was to just have another output function. This one would add each check
a JavaScript object.

That sounded to me like the easiest thing, road might be a bit steep though in the beginning.

> I've had a skim read through the whole script file and it feels like there should be a bit more
> structure around the scan results, which would allow the implementation of ratings (#100
> <https://github.com/drwetter/testssl.sh/issues/100>) to be a little more formal like ssllabs.

That's somewhere planned on the road, yes.

> However, I'm not a bash master and don't even know if bash can do what I'm thinking.

Well, in software development almost everything is possible. Just a matter of will
and complexity. ;-) If you think about it before doing that might be not too
[...]

PR #144, opened 2015-07-21
Public key pinning check code causes invalid arithmetic operator (OSX)

jonorossi:
Running the HEAD of testssl.sh on Mac OS X against https://testssl.sh is broken.

```
$ ./testssl.sh -H testssl.sh
[...]
--> Testing HTTP header response @ "/"

 HTTP Status Code             200 OK
 HTTP clock skew              0 sec from localtime
 Strict Transport Security    362 days=31337000 s, just this domain
")syntax error: invalid arithmetic operator (error token is "tssl.sh: line 699: 2592000
```

I fixed the error by normalising to LF on line 691 to remove CRs from the HTTP headers before writing out the HPKP `TMPFILE`, and removed the CR handling code from line 699:
```
691:    egrep -aiw '^Public-Key-Pins|Public-Key-Pins-Report-Only' $HEADERFILE | tr -d '\r' >$TMPFILE
699:    hpkp_age_sec=$(sed -e 's/^.*max-age=//' -e 's/;.*//' $TMPFILE)
```

Then I got another error on line 713 because sed on OS X doesn't like `sed -i` without a file name extension.

After these fixes to get it running the HPKP results still look wrong. https://testssl.sh provides two keys so the counting on line 695 doesn't seem to work (at least on OS X):
```
 Public Key Pinning           One key is not sufficent, 30 days=2592000 s, just this domain
                              No matching key for pin found
```

Could you please confirm the results are correct on Linux so I know the remaining issues are caused by tool differences.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -710,7 +710,7 @@ hpkp() {
 		preload "$TMPFILE"
 
 		# get the key fingerprints
-		sed -i -e 's/Public-Key-Pins://g' -e s'/Public-Key-Pins-Report-Only://' $TMPFILE
+		sed -i '' -e 's/Public-Key-Pins://' -e 's/Public-Key-Pins-Report-Only://' $TMPFILE
 		[ -s "$HOSTCERT" ] || get_host_cert
 		hpkp_key_hostcert="$($OPENSSL x509 -in $HOSTCERT -pubkey -noout | grep -v PUBLIC | \
 			$OPENSSL base64 -d | $OPENSSL dgst -sha256 -binary | $OPENSSL base64)"
--- a/testssl.sh
+++ b/testssl.sh
@@ -688,7 +688,7 @@ hpkp() {
 	fi
 	#pr_bold " HPKP                         "
 	pr_bold " Public Key Pinning           "
-	egrep -aiw '^Public-Key-Pins|Public-Key-Pins-Report-Only' $HEADERFILE >$TMPFILE
+	egrep -aiw '^Public-Key-Pins|Public-Key-Pins-Report-Only' $HEADERFILE | tr -d '\r' >$TMPFILE
 	if [ $? -eq 0 ]; then
 		egrep -aciw '^Public-Key-Pins|Public-Key-Pins-Report-Only' $HEADERFILE | egrep -waq "1" || out "(two HPKP headers, using 1st one) "
 		# dirty trick so that grep -c really counts occurrences and not lines w/ occurrences:
@@ -696,7 +696,7 @@ hpkp() {
 		if [ $hpkp_nr_keys -eq 1 ]; then
 			pr_litered "One key is not sufficent, "
 		fi
-		hpkp_age_sec=$(sed -e 's/\r//g' -e 's/^.*max-age=//' -e 's/;.*//' $TMPFILE)
+		hpkp_age_sec=$(sed -e 's/^.*max-age=//' -e 's/;.*//' $TMPFILE)
 #FIXME: test for number!
 		hpkp_age_days=$((hpkp_age_sec / 86400))
 		if [ $hpkp_age_days -ge $HPKP_MIN ]; then


drwetter:
Thx, @jonorossi . However your patch breaks the Linux side. ;-/



I'll can have a look into it later.

jonorossi:
:disappointed: maybe I'll just give up running this on my Mac, and just set up a VM. Congrats, this is an engineering feat doing so much in bash, especially the sockets. Have you looked at writing any automated functional tests?

drwetter:
Am 07/21/2015 um 11:10 AM schrieb Jonathon Rossi:
> :disappointed: maybe I'll just give up running this on my Mac, and just set up a VM. 

of course you could do that. OTOH your efforts so far helped others too. For me it's
kind of an strategic issue to have support for Darwin but my time is limited and
I don't own a Mac.

Every real release I do testing on FreeBSD 9, that is the latest moment when
I am trying to straighten out the Darwin bugs too as both are similar.

But it cannot hurt to do that regularly and earlier to not disgruntle Darwin users ;-)

If you still want to help that would be great. A Linux VM can also
be quite handy for testing issues with your Mac OS X patch.

> Congrats,
> this is an engineering feat doing so much in bash, especially the sockets. 

Thx!

> Have you looked at
> writing any automated functional tests?

Not really but partly. I have some private testcases I run once in a while or if big
changes are pending (e.g. in the past: sockets). And I do basic+small unit
testing for new features, so that ~all cases of those new features work (TM) ;-)

Do you have something on your mind?

jonorossi:
I worked out that this broke on Linux because OSX has BSD sed not GNU sed and the `-i` option is slightly different. I'll try to work out a portable option, however testssl.sh has so much text manipulation that I sort of question if bash is the right language for this sort of thing.

If you are keen to have darwin working then I'll persist, but knowing that my production is Linux I'll have to look at that soon.

With regards to automated tests and JSON output, I was wondering if testssl.sh should be broken into separate scripts in Git that can be run in isolation that return the results for that check. The released version of testssl.sh can still be a single merged bash script. I've never used it, but could we possibly use `openssl s_server` for automated tests. Thoughts?

drwetter:
Am 07/21/2015 um 02:05 PM schrieb Jonathon Rossi:
> I worked out that this broke on Linux because OSX has BSD sed not GNU sed and the |-i| option
> is slightly different. 

That somehow rings a bell, done that in testssl.sh. Have to look at it though... later.

> I'll try to work out a portable option, however testssl.sh has so much
> text manipulation that I sort of question if bash is the right language for this sort of thing.

Trust me, it is ;-) But I see your point somewhat.

Other languages  have the point that you need to install some thing -- either
the intepreter (or compiler) itself or some module. testssl.sh doesn't need this. It
is working out of the box.

I am not good at it yet but some text manipulations you can do with bash also.

> If you are keen to have darwin working then I'll persist, but knowing that my production is
> Linux I'll have to look at that soon.

It would be awesome if you could spend a few cycles on it.

> With regards to automated tests and JSON output, I was wondering if testssl.sh should be broken
> into separate scripts in Git that can be run in isolation that return the results for that
> check. The released version of testssl.sh can still be a single merged bash script. I've never
> used it, but could we possibly use |openssl s_server| for automated tests. Thoughts?

Not sure I understand you I don't know how it would work?

openssl s_server is not a bad idea but there are limitations. Don't know e.g. whether CCS,
Heartbleed
work with a borken version. At least last week I tried insecure client renegotiation with an
old openssl version and the s_server. That did not work.

Somewhat on the horizon I had in mind to split the script anyway (includes, text functions, simple
other functions, etc.), for the sake of cleaner code but that's not what you meant?

Cheers, Dirk

jonorossi:
> Other languages  have the point that you need to install some thing -- either
> the intepreter (or compiler) itself or some module. testssl.sh doesn't need this. It
> is working out of the box.

I guess that was the point I was making, it isn't working out of the box (on OSX) :wink:. I'm setting up a fresh Debian VM at the moment so I should be able to get this PR updated.

> Not sure I understand you I don't know how it would work?

I'm not sure exactly as I've never built something so big in bash, was more thinking how I'd do it in an OO language. Currently you've got the `heartbleed()` function, which like many of the other functions performs an action and writes a bunch to stdout. I was trying to think how I'd implement #31 without making the code completely unmaintainable and considering if each check should return their results that the main program formats to stdout rather than the actual function. We should probably continue this part of the discussion under #31.

drwetter:
Am 07/21/2015 um 03:05 PM schrieb Jonathon Rossi:
>     Other languages have the point that you need to install some thing -- either
>     the intepreter (or compiler) itself or some module. testssl.sh doesn't need this. It
>     is working out of the box.
> 
> I guess that was the point I was making, it isn't working out of the box (on OSX) :wink:.

Good point. But we're not on a stable release and developer accidents can happen here.

> I'm setting up a fresh Debian VM at the moment so I should be able to get this PR updated.
> 
>     Not sure I understand you I don't know how it would work?
> 
> I'm not sure exactly as I've never built something so big in bash, was more thinking how I'd do
> it in an OO language. Currently you've got the |heartbleed()| function, which like many of the
> other functions performs an action and writes a bunch to stdout. I was trying to think how I'd
> implement #31 <https://github.com/drwetter/testssl.sh/issues/31> without making the code
> completely unmaintainable and considering if each check should return their results that the
> main program formats to stdout rather than the actual function. 

I would rather leave the output to the functions, not the low level ones. But certainly
not main. How else would you pass all the details? Heartbleed is not a good example as
basically the traffic light is red or green. Other functions output numerical values, strings
or some even change during runtime (cipher order).

Cheers, Dirk

drwetter:
Fix follows in a second. Remark: Probably that was a bit late as I wrote this. As debugging it under BSD I remember the things I shouldn't have done for BSD sed (e.g. ``sed -i``  or adding a linefeed).

drwetter:
Pls test it under Mac OS and let me know if it still hiccups.

FreeBSD 9 works, thus I assume Darwin does work too.

jonorossi:
@drwetter confirmed it is now working on OSX without the changes in this PR. Thanks.

drwetter:
@jonorossi Good! Thx for your report and your PR.

I was wondering whether you would be willing to keep an eye on Darwin compatibility and either report an issue or make a PR. As said normally I do those tests on FreeBSD9  before a stable release but in the master branch between releases I lack time.

jonorossi:
@drwetter will try to do, ping me if you want me to do a more thorough test before a release.

Issue #140, opened 2015-07-20
awk issue MSYS2


PeterMosmans:
Issue when running the latest version under msys2 (Windows 8.1):
```
  testssl.sh       2.5dev from https://testssl.sh/dev/
    (9183127 2015-07-14 09:36:23 -- 1.310)

./testssl.sh --assuming-http -H testssl.sh

--> Testing HTTP header response @ "/"

awk: warning: escape sequence `\/' treated as plain `/'
awk: cmd. line:1: /tmp/ssltester.0coOKA/http_header.txt
awk: cmd. line:1:                                  ^ syntax error
```

drwetter:
Thx, Peter. for reporting.

No clue why this is currently -- as my MSYS2 system has GNU AWK 4.1.1 and e.g. the Linux system I am currently writing from is GNU AWK 4.1.0.

PeterMosmans:
Please let me know if I can be of any help
```
 % awk --version
GNU Awk 4.1.3, API: 1.1 (GNU MPFR 3.1.3, GNU MP 6.0.0)
% uname -a
MINGW64_NT-6.3 Paradise 2.1.0(0.287/5/3) x86_64 Msys

```

drwetter:
> Please let me know if I can be of any help

yes: to understand  why this fails under MSYS2.

I should have added in my previous posting that I see the same under MSYS2, but not under Linux.

PS: also I don't need ``--assuming-http`` under Linux.

drwetter:
see 9ffdf7f5808333ec33b7fe22dd533351503adfc0.

Repaired the parser for windowish nslookup. Tested it. Worked for me. Pls reopen it if it doesn't for you.

Thx for reporting

PeterMosmans:
Sorry, issue still persists:
```
 ./testssl.sh --assuming-http -H www.some.site

    testssl.sh       2.5dev from https://testssl.sh/dev/
    (9ffdf7f 2015-07-21 14:20:15 -- 1.324)

 Using "OpenSSL 1.0.2-chacha (1.0.2e-dev)" [~181 ciphers]


--> Testing HTTP header response @ "/"

awk: warning: escape sequence `\/' treated as plain `/'
awk: cmd. line:1: /tmp/ssltester.BBmyl5/http_header.txt
awk: cmd. line:1:               ^ syntax error
awk: cmd. line:1: /tmp/ssltester.BBmyl5/http_header.txt
awk: cmd. line:1:                                  ^ syntax error

```

[...]

PR #149, opened 2015-07-27
Fix for #140

PeterMosmans:
Suppress awk warnings
Remove dot from temporary HEADERFILE filename (generates errors on MSYS2 / AWK)

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -538,8 +538,8 @@ EOF
 			#ret=3
 		fi
 	fi
-	status_code=$(awk '/^HTTP\// { print $2 }' $HEADERFILE)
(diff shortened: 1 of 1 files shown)


drwetter:
Hi @PeterMosmans,

thx! 

Especially on Windows the TXT extension is quite handy. So I have a little bit of pain just removing it -- also because of a lot of other files are using this (run .``/testssl --debug=1 <URI>`` and have a look in the directory.

Did you get why awk under (your) windows has a problem with the dot? Or does quoting help?

Cheers, Dirk

PeterMosmans:
Dirk, please see my new pullrequest. I managed to work around the issue, so you can keep those handy extensions :smile:

PR #151, opened 2015-07-27
Another fix for #140

PeterMosmans:
Suppress awk warnings
Don't try to retrieve header information from openssl stderr output


drwetter:
looks good, thx!

Issue #166, opened 2015-08-17
Subject Alternative Name is empty


jonorossi:
The Subject Alternative Name reports as an empty string, I thought it should match the others with a double dash to indicate no data, and low and behold the code reads like that:

```
$ ./testssl.sh testssl.sh 
...
 Common Name (CN)             dev.testssl.sh (CN in response to request w/o SNI: default.name)
 subjectAltName (SAN)         
 Issuer                       StartCom Class 1 Primary Intermediate Server CA (StartCom Ltd. from IL)
...
```

https://github.com/drwetter/testssl.sh/blob/5c0559042f1ea891b158b97b0afb1c2093bb56a6/testssl.sh#L1775-L1783

Platform: darwin

drwetter:
Oh, thx!

jonorossi:
I just came across a cloudflare proxied site and the SAN is also empty so I guess that it doesn't show at all. Let me know if you think this is a darwin only problem and I'll try to work out what is causing it.

drwetter:
It's certainly doesn't occur on Linux. I can't tell whether it's a Darwin and/or BSD issue (right now lacking success to me BSD hosts).

I would bet it's Darwin only as this code is older and I would have spotted it on BSD.

Thx, Dirk

jonorossi:
I've made a fix in #173, and tested on OSX 10.10 and Debian 8.1.

drwetter:
Thx, see #173

PR #173, opened 2015-08-18
Fix subject alternative name on darwin

jonorossi:
Fix for #166

On OSX the following line doesn't seem to work (i.e. the input is good the output is empty), however [others have it working on OSX with separated -e args like you've got](http://stackoverflow.com/questions/1251999/how-can-i-replace-a-newline-n-using-sed). I went to change it from sed to tr, however realised we didn't need the new line at all.

```
L1794: sans=$(echo "$sans" | sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/ /g') # replace line feed by " "
```


drwetter:
great! Thank you

PR #180, opened 2015-08-27
Fixes #164 Mac OS X Darwin binaries with zlib support

### LINKED ISSUE #164: openssl.Darwin.x86_64 lacks zlib support
I've been trying to keep up with the changes happening with the binaries lately but don't recall anything about this not being supported on darwin. Is this a bug?

```
 CRIME, TLS (CVE-2012-4929)                Local Problem: Your ./bin/openssl.Darwin.x86_64 lacks zlib support
```

If we can't get openssl zlib support on darwin then maybe just add something to the release notes, or even write out a different message about being unsupported on darwin?

Platform: darwin

jpluimers:
Enjoy (:


drwetter:
Thx, @jpluimers !

PR #205, opened 2015-09-29
fix CBC cipher selection for BEAST detection

typingArtist:
My server runs ECDHE-ECDSA ciphers only, and as documented by SSL Labs, it’s vulnerable to BEAST as it supports TLS 1.0 and Cipher 0xc00a (TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA). However, testssl.sh doesn’t detect this situation as it’s never choosing the affected CBC cipher. Not all CBC ciphers have CBC in their name **for openssl**. Their RFC counterpart does have CBC in its name, but openssl doesn’t help us find the proper set of ciphers.
Since the list of CBC ciphers isn’t going to grow in the future, I think it’s safe to hard-code the list. Nevertheless, I supplied a small script to extract the CBC cipher list from the ciphers given in openssl-rfc.mapping.html, based on their RFC name.

Please pull.


typingArtist:
Sorry for the merge mess.

This is from the Qualys SSLLabs report of my server:

Protocols
TLS 1.2 	Yes	
TLS 1.1 	Yes
TLS 1.0 	Yes
SSL 3 	No	
SSL 2 	No


Cipher Suites (SSL 3+ suites in server-preferred order; deprecated and SSL 2 suites at the end)
TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256 (0xcc14)   ECDH 384 bits (eq. 7680 bits RSA)   FS 	256
TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 (0xc02c)   ECDH 384 bits (eq. 7680 bits RSA)   FS 	256
TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 (0xc02b)   ECDH 384 bits (eq. 7680 bits RSA)   FS 	128
TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384 (0xc024)   ECDH 384 bits (eq. 7680 bits RSA)   FS 	256
TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA (0xc00a)   ECDH 384 bits (eq. 7680 bits RSA)   FS 	256

The last one is actually the one used in TLS 1.0 connections and is not tried by the old version of testssl.sh as its openssl name is ECDHE-ECDSA-AES256-SHA (no CBC in the name).

PR #206, opened 2015-09-29
Improved support for servers using client certificate authentication

Iansus:
When using **client certificates**, openssl returns exit code 1 since the handshake is not successful.
However, that does not prevent this tool from listing ciphers & searching for vulnerabilities.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1099,7 +1099,8 @@ std_cipherlists() {
      pr_bold "$2    "         # indent in order to be in the same row as server preferences
      if listciphers "$1"; then  # is that locally available??
           $OPENSSL s_client -cipher "$1" $STARTTLS -connect $NODEIP:$PORT $PROXY $SNI 2>$ERRFILE >$TMPFILE </dev/null
-          ret=$?
+	  grep -v 'NONE' $TMPFILE | fgrep 'Cipher is' &> /dev/null
+	  ret=$?
           debugme cat $ERRFILE
           case $3 in
                0)   # ok to offer
@@ -1236,7 +1237,8 @@ test_just_one(){
                fi
                if [[ $? -eq 0 ]]; then    # string matches, so we can ssl to it:
                     $OPENSSL s_client -cipher $ciph $STARTTLS -connect $NODEIP:$PORT $PROXY $SNI &>$TMPFILE </dev/null
-                    ret=$?
+		    grep -v 'NONE' $TMPFILE | fgrep 'Cipher is' &> /dev/null
+		    ret=$?
                     if [[ $kx == "Kx=ECDH" ]] || [[ $kx == "Kx=DH" ]] || [[ $kx == "Kx=EDH" ]]; then
                          if [[ $ret -eq 0 ]]; then
                               dhlen=$(read_dhbits_from_file $TMPFILE quiet)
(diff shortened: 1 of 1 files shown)


Iansus:
Pull request was made a bit too soon, it still needs some corrections.

drwetter:
It has to be a unique string/strings. E.g.Tried that a bit already but did not get to the end.

Hints

```
[[ -z $(awk '/Master-Key: / { print $2 }' "$1") ]] 
[[ -n $(awk '/Session-ID: / { print $2 }' "$1") ]]
[[ 2 -eq $(grep -ac CERTIFICATE "$1" ) ]]

```

Iansus:
Well I think "Cipher is " should work, except the "NONE" case.
At least it does in my tests. The corrections were about logic flaws I
introduced in the pull requests.

Great tool anyhow !

drwetter:
Am 29. September 2015 19:12:38 MESZ, schrieb Iansus <notifications@github.com>:
>Well I think "Cipher is " should work, except the "NONE" case.
>At least it does in my tests. The corrections were about logic flaws I
>introduced in the pull requests.
>
>Great tool anyhow !
>Le 29 sept. 2015 6:57 PM, "Dirk Wetter" <notifications@github.com> a
>écrit :
>
>> It has to be a unique string/strings. E.g.Tried that a bit already
>but did
>> not get to the end.
>>
>> Hints
>>
>> [[ -z $(awk '/Master-Key: / { print $2 }' "$1") ]]
>> [[ -n $(awk '/Session-ID: / { print $2 }' "$1") ]]
>> [[ 2 -eq $(grep -ac CERTIFICATE "$1" ) ]]
>>
>>
>> —
>> Reply to this email directly or view it on GitHub
>>
><https://github.com/drwetter/testssl.sh/pull/206#issuecomment-144119933>.
>>
>
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/drwetter/testssl.sh/pull/206#issuecomment-144124410

As I tested it it was not 100% reliable. Keep in mind that people use strange openssl versions or have strange server side ssl stacks.

It's THE  core check  thus we need 100%.

Cheers, Dirk!




-- 
Sent from my mobile

drwetter:
Hi @Iansus ,

FYI: I have something in the works. As workaday life permits I'll finish that and commit it.

Cheers, Dirk

drwetter:
Every s_client connect call is now differently judged for success. A function is used:

```
sclient_connect_successful() {
     [[ $1 -eq 0 ]] && return 0
     [[ -z $(awk '/Master-Key: / { print $2 }' "$2") ]] && return 1
     [[ -z $(awk '/Session-ID: / { print $2 }' "$2") ]] && return 1
     # what's left now is: master key not empty and Session-ID not empty ==> probably client based auth with x509 certificate
     CLIENT_AUTH=true
     return 0
}
```
The function is called with ``sclient_connect_successful $? $TMPFILE``. That should be not much of a speed disadvantage, Maybe one of the check will vanish. But that needs to be proved by intensive tests.

I've done a lot of tests on regular hosts it and it worked (but needed some polishing the first two drafts).

I have not tested it much on hosts needed a certificate, @Iansus maybe you can give it a shot?

Cheers, Dirk

Iansus:
I'll check that ;)

drwetter:
cool!

PR #207, opened 2015-09-30
Fix CBC cipher selection for BEAST detection

typingArtist:
(redoing https://github.com/drwetter/testssl.sh/pull/205)

My server runs ECDHE-ECDSA ciphers only, and as documented by SSL Labs, it’s vulnerable to BEAST as it supports TLS 1.0 and Cipher 0xc00a (TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA). However, testssl.sh doesn’t detect this situation as it’s never choosing the affected CBC cipher. Not all CBC ciphers have CBC in their **openssl** name. Their RFC counterpart **does** have CBC in its name, but openssl doesn’t help us find the proper set of ciphers.
Since the list of CBC ciphers isn’t going to grow in the future, I think it’s safe to hard-code the list. Nevertheless, I supplied a small script to extract the CBC cipher list from the ciphers given in openssl-rfc.mapping.html, based on their RFC name.

This is from the Qualys SSLLabs report of my server:

```
Protocols
TLS 1.2 Yes
TLS 1.1 Yes
TLS 1.0 Yes
SSL 3 No
SSL 2 No

Cipher Suites (SSL 3+ suites in server-preferred order; deprecated and SSL 2 suites at the end)
TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256 (0xcc14) ECDH 384 bits (eq. 7680 bits RSA) FS 256
TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 (0xc02c) ECDH 384 bits (eq. 7680 bits RSA) FS 256
TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 (0xc02b) ECDH 384 bits (eq. 7680 bits RSA) FS 128
TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384 (0xc024) ECDH 384 bits (eq. 7680 bits RSA) FS 256
TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA (0xc00a) ECDH 384 bits (eq. 7680 bits RSA) FS 256
```

The last one is actually the one used in TLS 1.0 connections and is not tried by the old version of testssl.sh as its openssl name is ECDHE-ECDSA-AES256-SHA (no CBC in the name).

```
BEAST attack 	Not mitigated server-side (more info)   TLS 1.0: 0xc00a
```

The following is a cutout of the output of `testssl.sh --wide` on my server *without* the patch applied:

```
--> Testing for BEAST vulnerability

 BEAST (CVE-2011-3389)

 TLS1:
Hexcode  Cipher Suite Name (OpenSSL)    KeyExch.   Encryption Bits        Cipher Suite Name (RFC)
-----------------------------------------------------------------------------------------------------------------------
 no CBC ciphers for TLS1 (OK)
```

Now again *with* the patch:

```
--> Testing for BEAST vulnerability

 BEAST (CVE-2011-3389)

 TLS1:
Hexcode  Cipher Suite Name (OpenSSL)    KeyExch.   Encryption Bits        Cipher Suite Name (RFC)
-----------------------------------------------------------------------------------------------------------------------
 xc00a   ECDHE-ECDSA-AES256-SHA         ECDH       AES        256         TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA

VULNERABLE -- but also supports higher protocols (possible mitigation): TLSv1.1 TLSv1.2
```

Please pull.


drwetter:
Hello @typingArtist,

you have a very good point here, thx!  

I had the following comment already in a different tab but you were faster....

We should work to get this better. 

There's already a variable named ``cbc_cipher_list``. I would not like to have two lists. I guess my first list needed to be extended then. 
 A simple CBC cipher list of all current potential BEAST ciphers can be generated by ``testssl.sh -x CBC`` which also greps for the RFC names. The TLS 1.1/1.2 ciphers where the IV issue has been addressed need to be sorted out. The list MUST not exceed 128 ciphers as some servers bail out (in line 3407).  Maybe for the time being it's fine and I do a workaround afterwards.

Naming convention is: global vars in CAPITAL declared in the beginning globally, local vars in lowercase locally declared. 

Cheers, Dirk

typingArtist:
I don’t think we have to cherry-pick the list for only those supported in ≤TLS1.0 – higher CBCs shouldn’t work in SSLv3 and TLS1.0 anyway – it’s just a matter of performance. The length limitation, you mentioned, of 128 ciphers is now obsolete anyway as openssl is not used for cipher list generation anymore.

drwetter:
Thx a bunch! 

Few remarks. 
1) you brought the check a step fwd!
2) still there are some CBC ciphers missing, testssl.sh -x CBC with mapping file will tell (not your fault)
3) for speed improvements it would not hurt  to add 1-2 calls per protocol in the beginning. If those fail the remaining 228 (114*2) ciphers don't have to be tested
4) color should be changed to yellow if TLS 1.1 or TLS 1.2 is offered as this mitigates BEAST. In any case: It's a client issue. The test only lines out a potential problem on the client side

typingArtist:
Hi Dirk,

Thanks for the merge and the added comments.

1) you’re welcome!
2–4): see #208

PR #209, opened 2015-10-02
Readme: add link(s) to external contrib, such as docker images

jumanjiman:


PeterMosmans:
@jumanjiman , thanks for providing Docker images. However, as far as I could see your Docker image uses the default openssl version, which is "suboptimal" for testing purposes. I could be wrong though, as I haven't checked out the actual Docker image.

You could either add the binary openssl version that's distributed with testssl.sh (although this will probably change in the near future), or add buildscripts for the "recommended" 1.0.2-chacha fork.
(Yes, I'm biased :smile: )

drwetter:
@jumanjiman: Thx for the PR and the work on the docker image. Haven't looked into it yet but @PeterMosmans has a valid point here.

jumanjiman:
i've created an alpine linux package for openssl-chacha.
after i put it up on github, i'll update my docker images to use it, then post back here :-)

drwetter:
Hi @jumanjiman

don't know about the dependencies in docker images but better take "my" binaries as I spend some time to make it work under most Linux distros. A docker image which is only ok for the maybe 1% of alpine Linux users doesn't make sense to me.

If I include a hint in the Readme I would also need to be sure that the image is ok. I lack currently time. Is there anybody else who can look at it thoroughly? Requirements: Good understanding of Linux (the system) & docker, good idea what testssl.sh is about and people would use the image.

Cheers, Dirk

jumanjiman:
does your binary depend on glibc?

drwetter:
Am 3. Oktober 2015 00:28:37 MESZ, schrieb Paul Morgan <notifications@github.com>:
>does your binary depend on glibc?

Have a look at it.

As less as possible. The standard bins are statically linked.

Dirk 



-- 
Sent from my mobile

jumanjiman:
> The standard bins are statically linked.

indeed! :smile: 

as of https://github.com/jumanjihouse/docker-testssl/commit/d6ab5d36efc7957f7ed0b25e3795fc5e8aeb0332 my image uses the statically-compiled bin from https://testssl.sh/openssl-1.0.2e-chacha.pm.tar.gz as linked from https://testssl.sh/

the resultant image is 21 MiB uncompressed (7 MiB compressed https://quay.io/repository/jumanjiman/testssl?tab=tags).

____

> A docker image which is only ok for the maybe 1% of alpine Linux users doesn't make sense to me.

this doesn't make sense to me.
docker images are isolated user-space apps.
one can run the image as a container on *any* modern distro, such as coreos, ubuntu, fedora, rhel atomic, etc.

i choose alpine user-space because it's smaller than the 100's of MiB that fedora or ubuntu consume.

____

> If I include a hint in the Readme I would also need to be sure that the image is ok. I lack currently time. Is there anybody else who can look at it thoroughly?

the image now uses *your* statically-compiled openssl bin.

you can see the test results at https://circleci.com/gh/jumanjihouse/docker-testssl/27
(click on the dropdown marked "make test" to see output).

i don't know if any of your community is using docker yet, but
perhaps @progrium or @andyshinn or @kelseyhightower (sorry, gents) would be kind enough to
review the image source and offer feedback.

andyshinn:
I can't speak for testssl.sh as I've never used it. But the Docker image should be sound as long as the binaries being pulled in are static compiled. Basically, if they run, there shouldn't be any incompatibility with musl libc.

Other thoughts:

* The Docker community has a lot of "everything is just latest". It would be nice if I could see how the CI process was building the other tags. At first glance, it looks like the publish script is only automated for master and the other tags were built "somewhere else" and pushed manually.
* For me personally, I'd consider the `latest` tag to be a "stable" release and maybe a `master` tag to point at a `HEAD` build. This is at least a little less jarring for those who tend to run "tagless" (what do we call Docker users who don't specify a tag?).

drwetter:
Am 10/03/2015 um 05:03 AM schrieb Paul Morgan:
>     The standard bins are statically linked.
> 
> indeed! :smile:
> 
> as of jumanjihouse/docker-testssl@d6ab5d3
> <https://github.com/jumanjihouse/docker-testssl/commit/d6ab5d36efc7957f7ed0b25e3795fc5e8aeb0332> my
> image uses the statically-compiled bin from https://testssl.sh/openssl-1.0.2e-chacha.pm.tar.gz
> as linked from https://testssl.sh/
> 
> the resultant image is 21 MiB uncompressed (7 MiB compressed
> https://quay.io/repository/jumanjiman/testssl?tab=tags).

ok, thx Paul.

> -----------------------------------------------------------------------------------------------
> 
>     A docker image which is only ok for the maybe 1% of alpine Linux users doesn't make sense
>     to me.
> 
> this doesn't make sense to me.
> docker images are isolated user-space apps.
> one can run the image as a container on /any/ modern distro, such as coreos, ubuntu, fedora,
> rhel atomic, etc.

see.. that's the reason ;-) I was using docker images once in a while but have
little insight how it works from the inside.

Have you gotten feedback yet? The basic idea sounds really good to me as it takes
care of having a proper test environment.

Cheers, Dirk

drwetter:
Am 10/03/2015 um 07:17 AM schrieb Andy Shinn:
> I can't speak for testssl.sh as I've never used it. But the Docker image should be sound as
> long as the binaries being pulled in are static compiled. Basically, if they run, there
> shouldn't be any incompatibility with musl libc.

so, there's no/little chance for dynamically libs if they'll be included in the container? Just
wondering because I also distribute Kerberos binaries:

HOST:~/testssl/bin|% ldd openssl.Linux.x86_64-krb5
        linux-vdso.so.1 (0x00007ffde6550000)
        libkrb5.so.3 => /usr/lib64/libkrb5.so.3 (0x00007f778d1a8000)
        libk5crypto.so.3 => /usr/lib64/libk5crypto.so.3 (0x00007f778cf75000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f778cd71000)
        libz.so.1 => /lib64/libz.so.1 (0x00007f778cb5b000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f778c7ad000)
        libcom_err.so.2 => /lib64/libcom_err.so.2 (0x00007f778c5a9000)
        libkrb5support.so.0 => /usr/lib64/libkrb5support.so.0 (0x00007f778c39c000)
        libkeyutils.so.1 => /lib64/libkeyutils.so.1 (0x00007f778c198000)
        libresolv.so.2 => /lib64/libresolv.so.2 (0x00007f778bf81000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f778d477000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f778bd63000)
        libselinux.so.1 => /lib64/libselinux.so.1 (0x00007f778bb3f000)
        libpcre.so.1 => /usr/lib64/libpcre.so.1 (0x00007f778b8d2000)
HOST:~/testssl/bin|%

Probably that's more a somewhat esoteric question for most used cases. ;-)

Cheers, Dirk

MatthewVance:
I had the same idea about Dockerizing testssl.sh and started working independently on it a week or two ago. When I shared my version with @drwetter earlier today, he informed me of this related effort and thought we might be able to join forces. You all are welcome to make use of mine as you see fit. 

While I think we essentially accomplish the same end result, [my version](https://github.com/MatthewVance/testssl-docker) in a different manner.
Mine builds the [1.0.2-chacha fork](https://github.com/PeterMosmans/openssl/) by @PeterMosmans from source since it's the preferred OpenSSL library. Mine is based off of Debian Jessie rather than Alpine Linux. While it's bigger than Alpine, Docker's [Best practices for writing Dockerfiles](https://docs.docker.com/articles/dockerfile_best-practices/) recommends the use of Debian when possible since it's very tightly controlled, relatively small, and still full featured. Because of this, I think a lot of folks will already have a local copy of the Debian image on their machines so the extra storage space and download time required shouldn't be an issue.

Please let me know what you all think.

Regards, Matt

jumanjiman:
@andyshinn Thanks for the review!

> It would be nice if I could see how the CI process was building the other tags. At first glance, it looks like the publish script is only automated for master and the other tags were built "somewhere else" and pushed manually.

Those other tags were built on circleci and pushed from there to docker hub, but i stopped building those tags as of
https://github.com/jumanjiman/docker-testssl/commit/41186e6cb07dc46564876bcd1251233a84f864da



> For me personally, I'd consider the latest tag to be a "stable" release and maybe a master tag to point at a HEAD build. This is at least a little less jarring for those who tend to run "tagless" (what do we call Docker users who don't specify a tag?).

that's how i do my other repos.
in this case, testssl merged 2.7dev into 2.6 branch, so i punted and just build against master with a "date-time" as part of the tag. that way, each user can decide which tag to use.
it's not the best tag, but it seemed wrong to have a 2.6 tag after 2.7dev got merged into 2.6 branch.

@drwetter 



> so, there's no/little chance for dynamically libs if they'll be included in the container? Just
> wondering because I also distribute Kerberos binaries:

dynamic libs are fine in container as long as the libs in the container match the build.

my container is optimized for size, and alpine uses musl, not glibc.
therefore your static bins are great (and i now use them as described above), but dynamic libs don't match musl.

see also http://www.etalabs.net/compare_libcs.html

@MatthewVance 

> Mine builds the 1.0.2-chacha fork by @PeterMosmans from source since it's the preferred OpenSSL library.

mine too. i originally used stock alpine version of openssl, then i switched to alpine/musl build of the `-chacha` version, then i switched to @PeterMosmans statically-compiled version since it's independent of the underlying libc.





> Also, mine is based off of Debian Jessie rather than Alpine Linux. While it's bigger than Alpine, Docker's Best practices for writing Dockerfiles recommends the use of Debian when possible since it's very tightly controlled, relatively small, and still full featured.

whenever a vendor says "best practice", i silently append "for some segment of the user population". :grin: 

some users need RHEL and only RHEL for its gov certifications.

there's room for multiple approaches.

my PR lists available docker images as a bullet-list.
there's room for multiple links, and i for one like choice.
debian? check. rhel? check. alpine? check.

i would see yours as just another bullet in the list.

the idea is that @drwetter feels comfortable saying, "the community feels this external contrib is probably ok. i don't certify it; i just link to it as something to consider that won't embarrass me".

@drwetter 

> The basic idea sounds really good to me as it takes
> care of having a proper test environment.

exactly!

drwetter:
I'll comment later on that but didn't want to let this PR wait and wait and ..

drwetter:
> >    For me personally, I'd consider the latest tag to be a "stable" release and maybe a master
> >  tag  to point at a HEAD build. This is at least a little less jarring for those who tend to run 
> > "tagless"   (what do we call Docker users who don't specify a tag?).
>
> that's how i do my other repos.
> in this case, testssl merged 2.7dev into 2.6 branch, so i punted and just build against master
 > with a "date-time" as part of the tag. that way, each user can decide which tag to use.
> it's not the best tag, but it seemed wrong to have a 2.6 tag after 2.7dev got merged into 2.6 
> branch.

I didn't get the thing with the tags @ github completely yet. :-) I added the tags manually for the stable branches. However when I do a ``git show v2.6`` I see as latest the commit I did as a backport of an EV certificate OID. That seems correct to me. What's the best now to tag the 2.7dev tree (best from the first 2.7dev commit on)?

> my container is optimized for size, and alpine uses musl, not glibc.
> therefore your static bins are great (and i now use them as described above), but dynamic libs 
> don't match musl.

Didn't know until your comment bit about musl,  thx!

1-2 things from me:

On https://testssl.sh/openssl-1.0.2e-chacha.pm.ipv6.Linux.tar.gz you find binaries with Peter's branch with IPv6 support. They are not on github yet and superseed the ones without IPv6 in the name. The 2.7dev branch has IPv6 support and would need those binaries.

Please don't forget the copyright s notices for OpenSSL (OPENSSL-LICENSE.txt) and for testssl (GPLv2), this I haven't checked yet. 

The rest of the discussion I consider this more your business and I won't interfere ;-)  Hope though that there would be some overlap, in a positive sense. I got the impression though that you understood what you were doing so the right thing to do is to add "your links" (both) -- that stays so until people start cursing at me ;-) 

It should make the project more usable and helps get out the message. Thank you!

Cheers, Dirk

MatthewVance:
Sorry I’m slow getting back to everyone. I’m glad to see all the discussion and progress on this. I also appreciated seeing my link merged into the pull request. 

@jumanjiman I didn’t mean to imply mine was better than yours. When I talked about building the OpenSSL libraries from source in my version, I meant that I ran make to compile them vs pulling in a binary. I think both are valid approaches and both get us what we want--a version of testssl.sh that uses the OpenSSL fork by @PeterMosmans. 

That custom OpenSSL library requirement was one reason I decided to Dockerize testssl.sh. In actuality, I had trouble getting the tool to work with the distributed OpenSSL binaries (likely something I did wrong). As I thought about how to best solve this dependency requirement, I realized Docker might be a good work around and could also help others get up and running quickly with this neat ~~SSL~~ TLS testing tool. 

> there's room for multiple links, and i for one like choice.
> debian? check. rhel? check. alpine? check.

I completely agree. Choice is good. It’s great Linux gives us so many choices. It’s also really cool how a Docker container can use any distro and essentially run from any host distro. 

> whenever a vendor says "best practice", i silently append "for some segment of the user population". :grin:

Valid pint. In this case I agreed with Docker's recommendation for a variety of reasons...it also helps that I’m a fan of Debian and more familiar with it. With that being said, I still think the concept behind Alpine is neat. Like you said, it's easy to add bullets for different Dockerfiles so users have choice. 

@drwetter 
> Please don't forget the copyright s notices for OpenSSL (OPENSSL-LICENSE.txt) and for testssl (GPLv2), this I haven't checked yet. 

I already have those notices in place in my [`README.md`](https://github.com/MatthewVance/testssl-docker/blob/master/README.md). If further notices are needed, please let me know and I’ll be happy to make the changes. I released my code under the very permissive [MIT License](https://github.com/MatthewVance/testssl-docker/blob/master/LICENSE) which I believe is compatible with GPLv2.

jumanjiman:
> I didn’t mean to imply mine was better than yours.

good, i didn't think you did!



> it also helps that I’m a fan of Debian

agree. you mentioned you have numerous debian-based images.
it makes perfect sense to use a common base.




> On https://testssl.sh/openssl-1.0.2e-chacha.pm.ipv6.Linux.tar.gz you find binaries with Peter's branch with IPv6 support. They are not on github yet and superseed the ones without IPv6 in the name. The 2.7dev branch has IPv6 support and would need those binaries.

gtk. i'll look at that.



> Please don't forget the copyright s notices for OpenSSL (OPENSSL-LICENSE.txt) and for testssl (GPLv2)

i'll be sure to add them in.

> What's the best now to tag the 2.7dev tree (best from the first 2.7dev commit on)?

good question. every distro has its own preferences.
http://semver.org/ is a popular choice.

@MatthewVance @drwetter One cool thing (to me) would be to structure our docker images so that we could have sample tests for different (external) scenarios, where our docker containers are interchangeable in a test harness. IOW, "here's a couple ways to use testssl.sh to test arbitrary services. Choose any of `<these>` docker containers to run your tests because these containers support the same run-time options", where run-time optiions are consistent with https://benchmarks.cisecurity.org/tools2/docker/CIS_Docker_1.6_Benchmark_v1.0.0.pdf (such as `--read-only --cap-drop all`).

MatthewVance:
@jumanjiman, I'm glad I didn't offend you. It's not always easy to convey the right tone when writing messages and sometimes after re-reading/thinking about something I said, I realize it may not have come across as intended. 

> One cool thing (to me) would be to structure our docker images so that we could have sample tests for different (external) scenarios, where our docker containers are interchangeable in a test harness.

This sounds like a good idea and one I'm willing to collaborate on with you. I'll be upfront with you though. I'm not a developer so I'll need your help/guidance in actually knowing how to go about doing this.  

Would it also make sense to have some standardized tests to make sure our containers perform all the functions that testssl.sh supports? I know my own container currently has an [open issue](https://github.com/MatthewVance/testssl-docker/issues/1) where specifying URLs rather than IP addresses results in a `Fatal error: No IPv4 address for "example.com" available` error message. I'm not sure why either because when I open a bash shell within the container, I'm able to ping a domain name so it's not like the container can't resolve domains to IP addresses.

jumanjiman:
> This sounds like a good idea and one I'm willing to collaborate on with you.

awesome!

i suspect my colleague @sometheycallme (tim) would like to help, too.



> Would it also make sense to have some standardized tests to make sure our containers perform all the functions that testssl.sh supports?

that'd be great.

at `$dayjob`, we've been considering https://github.com/sstephenson/bats for security test harness.
testssl.sh is written in bash, so bats seems like a good fit here. many admins/devs are comfortable with shell, typically more so than python/ruby/go(lang). 

what do you think of bats?

in an ideal world, we converge on common tests and get the tests and docker images merged into testssl.sh repo. this reduces barrier to entry for others.


> I know my own container currently has an open issue where specifying URLs rather than IP addresses results in a Fatal error: No IPv4 address for "example.com" available error message. I'm not sure why either because when I open a bash shell within the container,

i'll take a look and comment in your repo. (but not tonight :tired_face: )

MatthewVance:
@jumanjiman I'm not familiar with bats. I'll have to take a look at it (another day :tired_face:). 

I just fixed a couple of wrong refs in my readme. I used by github user name in the docker run commands rather than my docker hub user name so all my example commands were wrong. Not sure how I missed that earlier. :disappointed:

Even though I didn't touch the Dockerfile, this triggered an automated build which failed. Turns out, the OpenSSL fork changed a few days ago so my SHA1SUM check didn't work. Since the Git URL I'm using to pull down the custom OpenSSL source isn't based on a release, this will happen again so I removed the SHA1SUM check for OpenSSL since trying to keep it up-to-date is infeasible for me. My local re-build worked so I just pushed the fix. Hopefully, the Docker Hub automated build will also work. I don't plan to wait around tonight to make sure...:sleeping:

> i'll take a look and comment in your repo. 

Thanks!

MatthewVance:
@jumanjiman FWIW, the resolution to the matthewvance/testssl-docker#1 issue was installing the dnsutils package. @promythyus identified the fix. He also resolved a `hexdump: command not found` issue I had missed. This is a prime example of why having a good set of tests for our images would be helpful.

drwetter:
Guys ( @jumanjiman /  @sometheycallme / @MatthewVance ): have you seen the web interface @ https://github.com/TKCERT/testssl.sh-webfrontend from @thomaspatzke ?

Just wondering whether you picked that up?

PR #221, opened 2015-10-16
Changed wording for easier readability.

wlovins:
The sentence did not read correctly with the number of available ciphers where it was, so I moved it to where I feel it should be.


drwetter:
thx!

PR #225, opened 2015-10-30
Fix: tput: No value for $TERM and no -T specified

Harinus:
Avoid "tput: No value for $TERM and no -T specified" when running from CGI or similar by checking for interactive shell

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -92,10 +92,12 @@ date --help >/dev/null 2>&1 && \
      readonly HAS_GNUDATE=false
 echo A | sed -E 's/A//' >/dev/null 2>&1 && \
      readonly HAS_SED_E=true || \
-     readonly HAS_SED_E=false 
-TERM_DWITH=${COLUMNS:-$(tput cols 2>/dev/null)}     # for future custom line wrapping
-TERM_CURRPOS=0                          # ^^^ we also need to find out the length or current pos in the line
+     readonly HAS_SED_E=false
 
+if [[ $- == *i*]]; then			#Prevent tput errors if running non interactive shell
+	TERM_DWITH=${COLUMNS:-$(tput cols 2>/dev/null)}     # for future custom line wrapping
+	TERM_CURRPOS=0                          # ^^^ we also need to find out the length or current pos in the line
+fi
 
 # following variables make use of $ENV, e.g. OPENSSL=<myprivate_path_to_openssl> ./testssl.sh <host>
 # 0 means (normally) true here. Some of the variables are also accessible with a command line switch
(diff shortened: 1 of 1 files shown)


drwetter:
Thx!

I'll revert the change in line 5200 though as it doesn't make sense to me.

Harinus:
Hi,

sorry for the little delay..

1) I assume that when any script is running non-interactive the console is not visible to the User so there are no colors necessary. Anyways the result, returned by the attached CGI example is still colorized by AHA.

2) No non-cgi example, sorry.. But I  attached my CGI Use-Case
tput_err

```
#!/bin/bash

target="https://a" #The error produced by the invalid URL does not affect the mentioned behaviour

echo "Content-type: text/html"
echo ""

result=$(eval "'/var/www/bin/testssl.sh' --openssl='/var/www/bin/openssl.Linux.x86_64-krb5' '"$target"' 2>&1 \
| aha --title 'Result for "$target"' ")

echo "$result"
```

drwetter:
Hi Martin,

Am 11/02/2015 um 11:27 AM schrieb Martin Hoffmann:
> Hi,
> 
> sorry for the little delay..

no prob.

> 1) I assume that when any script is running non-interactive the console is not visible to the
> User so there are no colors necessary. 

I would rather prefer the user to decide. There's an option --color. Have you looked into that?

> Anyways the result, returned by the attached CGI example is still colorized by AHA.

That's on purpose. If you don't want colors, you need to to tell testssl you don't want
that. Personally I would see it as a great disadvantage not to have colors when piping
through aha.

With your PR and my correction I did today the errors of tput due to the missing TERM variable
vanished.

HTH, Dirk

Harinus:
Hi Dirk,

piping through aha does not cause the colors to be removed.. Because the shell is still interactive in that case.. Just the CGI environment (and maybe some other automated tasks) use the non-interctive shell.
Anyways the latest Version fixed the issue for me! Thanks

One last point: Since the 2.6 you changed the Headlines from blue to underline+bold in commit
https://github.com/drwetter/testssl.sh/commit/eb4913268287843cd6df7af2a78a3389dc01ae19
In my opinion the old style was much easier to read, especially in the HTML output. You may consider reverting that change ;)

Best regards, Martin

drwetter:
Hey Martin,

> One last point: Since the 2.6 you changed the Headlines from blue to underline+bold in commit
eb49132

(phew, that was fortunately the master)

> In my opinion the old style was much easier to read, especially in the HTML output. You may 
> consider reverting that change ;)

Ok, I consider that. To tell the reason behind it: I wanted to reduce the colors not really necessary in order to have them available in the future for something else. For what is not clear in detail yet. ;-)

I must admit though that using blue for the headline had more eye candy than just underlining it.  Maybe I revery that change, maybe not. The best would be to have a config file for this. More or less it's a matter of taste and it's difficult to get this alright for everybody. The coloring defaults are anyway fairly limited as one person has a black background terminal, another a white one, one could have 256 colors, the other one only 8. </offtopic>


Cheers, Dirk

PR #229, opened 2015-11-06
mDNS Fixes

lainegholson:
Support rDNS with mDNS for the private network (192.168) and support for OS X Bonjour with `dig`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4515,7 +4515,7 @@ determine_rdns() {
      local saved_openssl_conf="$OPENSSL_CONF"
      OPENSSL_CONF=""                         # see https://github.com/drwetter/testssl.sh/issues/134
 
-     if [[ "$NODEIP" == 192.168.*.* ]]; then
+     if [[ "$NODE" == *.local ]]; then
           if which avahi-resolve &>/dev/null; then
                rDNS=$(avahi-resolve -a $NODEIP 2>/dev/null | awk '{ print $2 }')
           elif which dig &>/dev/null; then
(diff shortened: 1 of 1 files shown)


teward:
Private Networks are not just 192.168.x.x, there are actually three distinct private IPv4 ranges that are explicitly considered private IP space.  Unless mDNS / Avahi only operates on 192.168.x.x and not the others, should this not be updated to work with the other private IPv4 ranges?

drwetter:
Good idea in general.

Major points: For reverse DNS  AND 198.168.* you'll rely on bonjour/avahi/mDNS  That won't work for the majority of data centers.

@teward is right, see RFC 1918.  That check seems first unnecessary. I suggest the logic if regular lookups fail try avahi-resolve and maybe dig. The problem with those two are loong timeouts though. 

Thus I would rather check  whether $NODE is ".local" also for reverse DNS. That would penalise not all.

Minor points: if possible I would stick with awk here. Looks better if you don't change those too often.

lainegholson:
So just checking if $NODE=.local and doing avahi/mDNS rDNS and everything
will be fine, right?

And about awk: I just used the way you scraped dig, which uses sed

drwetter:
Am 7. November 2015 04:28:19 MEZ, schrieb Laine Gholson <notifications@github.com>:
>So just checking if $NODE=.local and doing avahi/mDNS rDNS and
>everything
>will be fine, right?

Should be, yes. If you could do one run of tests, that would help.


>And about awk: I just used the way you scraped dig, which uses sed

Touché ;)

I can change that later. 

Cheers, Dirk 


>
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/drwetter/testssl.sh/pull/229#issuecomment-154611367


-- 
Sent from my mobile

lainegholson:
Everything works fine, myhostname.local and google.com both resolve.
testssl sh

drwetter:
Thx

PR #241, opened 2015-11-26
Show SHA384 certificates as secure

lainegholson:
And also fix a possible typo


drwetter:
Thx for catching this!

Issue #236, opened 2015-11-14
color-blind


drwetter:
On my recent workshop the request popped up to help color-blind people, 4% of men have difficulties to tell red from green apart.  Probably it's the best to swap for those people the colors green by blue. That can be done with a switch.

Most flexible would be a config file though.

Issue #232, opened 2015-11-11
Option to only show "not OK" warnings


frdmn:
I would love to have a feature which will show warnings that you need to fix (red ones, `Not OK`'s), instead of showing the full output of the checks.

What do you think?

drwetter:
you mean like ``--show-only=severe`` or --show-only=high`` ?

frdmn:
@drwetter exactly! :)

drwetter:
could do that by I would rather push that back on the agenda as other points are more important --  unless somebody jumps in

frdmn:
@drwetter Sure thing, no hurry. :+1:

Dude4Linux:
I'm looking for a similar capability.  There are currently four levels of severity, "OK", "INFO", "WARN", and  "NOT OK".  I was thinking of something like --log-level="WARN" which would limit logged output to the severity levels equal or greater than the specified log level, in this case "WARN" and "NOT OK".

@frdmn would this meet your needs or do you also want to limit output to stdout?

frdmn:
@Dude4Linux Yes, totally. I don't really care if it's stderr or stdout. :)

Dude4Linux:
@frdmn What I was getting at was limiting the output to the log file seems much simpler than trying to limit the output to the console.  Only the fileout() routine would need to be modified and the severity level is one of the arguments.

drwetter:
Probably we need a thorough renaming first. I've discussed today with @thecky that e.g. pr_red  becomes "something with high", pr_litered ""something with critical" etc.

Then this has to be in line with fileout().

This both is not rocket science. The difficult thing is to print out the findings based by a filter. The thorough approach is .also to suppress what is currently being tested, i.e.  ``SSLv3      not offered (OK)`` shouldn't show anything.  Patching the colored output only will show ``SSLv3      `` which is ugly. 

That requires a lot of work.

I think this is a good idea. Personally I do not have the resources. My plan is a) working on fixing the bugs b) doing the 2.8 release.

Dude4Linux:
@drwetter sounds like we need to be patient until the code changes you speak of are complete.

BTW, Thanks for developing testssl.sh.  I've found it very useful in testing the family of TurnKey GNU/Linux appliances.  I'm wanting to develop an Ansible playbook to use it in QA testing.

drwetter:
Sure.

I do not mind if someone steps fwd and does this. That's why the project is here and why it's open source.

[...]

PR #247, opened 2015-12-12
added --colorblind and --mapping options

thecky:


drwetter:
Thanks Thomas! Also bear with me pls.

drwetter:
Thx a bunch, Thomas! Pls also bear w/ me.

thecky:
Hi Dirk, I renamed the (most) pr_* functions and tput variables (f.e. pr_red -> pr_svrty_critical) as first step related to #232.
In set_color_functions() is now the "color-blind option" (swap green/blue #236) implemented, for tput and bash colorcodes.
The --mapping=no-rfc option suppress the cipher RFC-names in the output and also the width adjustment for "KeyExch." is fixed for COLOR=1.

drwetter:
Hi Thomas,

thx first! And sorry I didn't notice the commit 5 days ago -- strange.

Another bigger merge.. hmm.. Same as for the other pull requests: In the future pls only one thing at a time, it makes a review and testing easier (here e.g.: 1: the color blind + rename thing 2: the rfc thingy).

Hope I find the time before Xmas to merge this manually (branch conflicts). Pls bear with me.

Cheers, Dirk

drwetter:
As ``--mapping=rfc`` is the default, is there any reason we need it?

thecky:
Well, due the fact that --mapping=rfc is the default, I can remove the "rfc" - is just one line. If there's no other need for a later additional parameter (like no-ossl), I can also change this to an boolean option (f.e. --no-rfc-mapping).

To the ANSI escape codes - I think you mean the ANSI block between the two tput blocks (~line 423 and 473). These ANSI blocks can also be moved to the end of the if/else block (so the order is tput, tput (old bsd), ansi - instead tput, ansi, tput (old bsd)). Or do you mean an separate set_color_function for the ANSI codes?

drwetter:
Hi Thomas,

Later on I would love t have just the RFC/IANA style cipher name instead of the ones from OpenSSL  -- per switch -- but I would be happy for now if there would be just this one option to remove the RFC ones. If not wanted otherwise people need to remove the file -- kind of ugly as next occasion it'll show up again.

ANSI is @ https://github.com/drwetter/testssl.sh/blob/master/testssl.sh#L293-L346 . Termcap/terminfo @ https://github.com/drwetter/testssl.sh/blob/master/testssl.sh#L293-L346 . Those are two different things. Maybe testssl would only need one but I do not see a reason to mix the two parts today. Reversing https://github.com/thecky/testssl.sh/commit/d5e7d14ea9e75f9337514a36afe6e53dbec40026 should do it.

Cheers, Dirk

PR #248, opened 2015-12-13
Add HTTP2/ALPN support

lainegholson:
OpenSSL doesn't show the ALPN advertised protocols, so I try to test the common ones (HTTP/2, recent HTTP/2 drafts, SPDY/3.1, and HTTP/1.1). I also changed some of the SPDY code because connecting to `$NODE` will cause `gethostbyname` errors with the static binaries, and the extra line that `run_spdy` prints seems redundant.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2485,6 +2485,7 @@ run_spdy() {
 run_http2() {
      local tmpstr
      local -i ret=0	
+     local had_alpn_proto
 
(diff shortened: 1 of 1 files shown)


drwetter:
awesome, @lainegholson !

Give me some time to do sanity checks before merging, I have a busy week aead, er... ahead ;-)

drwetter:
Thx a lot, @lainegholson

PR #242, opened 2015-12-10
Add JSON and CSV output to testssl.sh

seccubus:
@drwetter ,

This is how I intend to add JSON output to ssl test so it can be used by other tools. Does this seem like to correct way to do it?

Please let me know.

P.S. Don't merge this pull request yet, it is work in progress.
P.P.S. It's done now.

PR #259, opened 2015-12-21
Feature/redirect plain

seccubus:
This PR (which build on the csv/json file PR) adds a warning if somebody redirects a user off an https site.


drwetter:
Thx, @seccubus, for the pile of work and your PR!

I have two problems though now: 1st: I am lacking time. 2nd: It would haven been easier for me to get 1 PR per function, so that I can test each change before merging.

Let me scratch my head a bit to work out the least painful path for us.

seccubus:
If you do #242 first, this should merge without any issues.

Do you want me to handle the backticks?

drwetter:
Frank, the backticks are not a problem as they easily can be patched again.

I normally do manually unit testing, so if ONE patch with ONE feature comes in it's easy to test. A bundled patch doesn't work and it either causes lots of afterwork to test or a different approach.

Can't you just first reduce the PR to "just" the JSON feature first? That is an easier goer as it doesn't require FOR NOW unit testing -- it's a new feature and can mature over time in the master branch.

Cheers, Dirk

drwetter:
Hi Frank,

I have several PR sitting here and would start working on it soon. Thus the question to you: Is that possible for you to rebase and send single PRs or do you rather want me to do this?

Cheers, Dirk

seccubus:
I can make this work. If you can pull it then, that would be great! 
Means I can finally integrate it into our tool chain.

drwetter:
yes, would be great if you could give me the JSON patch first for integration and the rest step by step.

seccubus:
Close since this is already covered and in PR #268 and PR #260

PR #260, opened 2015-12-21
Redirect test that doesn't depend om JSON/CSV output

seccubus:
@drwetter you can add this one first if you do'nt want to go through #242 first. but #259 adds support for JSON/CSV as well.

And... No backticks in this one ;)


drwetter:
Thx!

PR #266, opened 2016-01-15
Detect multiple server certificates

dcooper16:
Currently, run_server_defaults() will only display information about a single server certificate, even if the host being tested has more than one certificate. This PR tries to find all of the certificates that a server uses and displays information about each one.

An example of a server that has multiple certificates is blog.cloudflare.com (198.41.215.66), which has both one certificate with a 2048-bit RSA key and another with a 256-bit EC key.


dcooper16:
No problem. There's no need to rush. I just hope you find this useful.

Sorry if I just caused a problem. I'm new to Github, and I can't seem to get it to do want I want. I made a few changes to my proposed code and wanted to replace the old commit with the new one, but the result was just an added commit that seemed to incorporate changes from other recent commits. I tried to fix it by cancelling the PR, but even that didn't work. I seems that looking at the "files changed" tab is okay, but the individual commits aren't.

The changes that I made are only in run_server_defaults(). I added some code up front to speed up the code (to reduce the number of cipher suites tried during the loop), and I also added some code to handle the case in which the server does not appear to have any certificates (maybe it only supports anonymous cipher suites or ones that use pre-shared keys).

drwetter:
Am 15. Januar 2016 22:19:33 MEZ, schrieb dcooper16 <notifications@github.com>:
>No problem. There's no need to rush. I just hope you find this useful.

Of course. 

>Sorry if I just caused a problem.

Not at all :)

> I'm new to Github, and I can't seem
>to get it to do want I want. I made a few changes to my proposed code
>and wanted to replace the old commit with the new one, but the result
>was just an added commit that seemed to incorporate changes from other
>recent commits. I tried to fix it by cancelling the PR, but even that
>didn't work. I seems that looking at the "files changed" tab is okay,
>but the individual commits aren't.
>
>The changes that I made are only in run_server_defaults(). I added some
>code up front to speed up the code (to reduce the number of cipher
>suites tried during the loop), and I also added some code to handle the
>case in which the server does not appear to have any certificates
>(maybe it only supports anonymous cipher suites or ones that use
>pre-shared keys).

Thx for your description. I'll look at it closer soon.

Only thing was thinking on was not to include it in the default run, as it costs for 97% of the usage more time for the benefit for ~3% of the web sites. In the end and during the preferences run one could determine whether there's another certificate and just raise the flag saying "you want to run the option AAA for checking for another certificate".

More to it next week.

Cheers, Dirk 

-- 
Sent from my mobile

dcooper16:
I made some changes to the code in order to speed it up significantly. Rather than test every cipher suite, the new code just tries seven blocks of cipher suites:
     1)  cipher suites using certificates with RSA signature public keys
     2) cipher suites using certificates with RSA key encipherment public keys
     3)  cipher suites using certificates with DSA signature public keys
     4)  cipher suites using certificates with DH key agreement public keys
     5) cipher suites using certificates with ECDH key agreement public keys
     6) cipher suites using certificates with ECDSA signature public keys
     7) cipher suites using certificates with GOST R 34.10 (either 2001 or 94) public keys

So, now run_server_defaults just calls determine_tls_extensions seven times instead of over 100 times.

I also modified some of the code in determine_tls_extensions and certificate_info in order to speed things up. The current code calls "$OPENSSL s_client ..." many times: in determine_tls_extensions to, in get_host_cert to get the server's certificate, in run_server_defaults to see what happens if SNI is not provided, in get_all_certs to determine how many intermediate certificates are provided, and in determine_trust to see if the server's certificate has been validated. I modified the code to eliminate all of these except the one in determine_tls_extensions and the one to see what happens if SNI is not provided. The revised code seems to provide the same functionality without sending as many messages to the server.

In addition to speeding up the code, another reason that I made this change is that when testing my code on a web server, I discovered that (probably due to load balancing) the server didn't always return the same certificate. In fact, I encountered three different certificates. So, that means with the current code, each call to "$OPENSSL s_client ..." to get the host certificate, determine trust, determine number of intermediate certificates provided, etc., may be working with a different host certificate.

Given my experience last week I (tried) not to update this pull request. Instead I just posted the new code on in a new branch on my fork. So, it can be seen at https://github.com/dcooper16/testssl.sh/blob/multiplecerts/testssl.sh.

drwetter:
Oh, in fact that is a dramatic increase in speed, thx!
For the changes we need to look at -- at least on https://testssl.sh your version doesn't succeed for the request w/o SNI. You might want to log the current master and yours and do a ``diff -uw``.

Probably either you or @seccubus JSON PR won't apply -- depending on whose PR I merge second. I doubt that both can be merged without conflicts.

One way or the other I'll figured out.

drwetter:
By the way: good work!
-- 
Sent from my mobile

drwetter:
Hi @dcooper16 

Ok, the JSON patch I applied yesterday. In addition to the changes of that it seemed to me that your branch was a little behind.

Bottomline: With git mergetool is that kind of difficult to me to get it right -- just tried.  I can offer to merge it completely manually or you rebase and send me a new PR. What would be the best for you? 

Cheers, Dirk

dcooper16:
I'll try to do it. First I'll work create a version of the script that incorporates my edits into the latest version, then I'll try to create the PR. I still haven't figured out Github and pull requests. I'm much more familiar with just using diff to create patch files.

drwetter:
maybe that helps? It's a diff to the old version. 

[mcerts.patch.txt](https://github.com/drwetter/testssl.sh/files/108690/mcerts.patch.txt)

PR #268, opened 2016-01-18
New JSON/CSV output pull request

seccubus:
Replaces pull requests #242 an #259 

Rebased on todays master and added JSON/CSV output to HTTP and Client simulations as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+.DS_Store
+*.json
+*.csv
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -73,6 +73,7 @@
 
 * Frank Breedijk
  - JSON and CSV output
+ - Client simulations
 
 * Bug reports:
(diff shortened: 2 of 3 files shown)


seccubus:
@drwetter this is a cleaned up version of my earlier work that should merge without any problems.

drwetter:
See merge. Thx a bunch, @seccubus .

There's still some work to do though (thus labeled it as experimental).I'll follow up with some thoughts.

Please mind for future PRs the coding guideline @ https://github.com/drwetter/testssl.sh/wiki/Coding-Style, especially:

* no backticks
* double square brackets
* pls use bash internal functions where possible ``echo $1|sed "s/\"/\\'/g   -->  echo "${var/\"/\'}"  `` [ doesn't work with multi line ]

Despite that, you brought testssl.sh a good step fwd again. Thx

PR #272, opened 2016-01-23
added protocol defaultports for --starttls option

thecky:
The --starttls=<protocol> failed, because the corresponding port isn't set, so the default 443 was used.
Error-Msg: Fatal error:  ./bin/openssl.Linux.x86_64 couldn't establish STARTTLS via smtp to XX.XX.XX.XX:443

This patch set the right PORT-variable for starttls specified protocol.

It's also possible to set another port for the starttls protocol with --starttls=<protocol>:<port>

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4562,6 +4562,7 @@ $PROG_NAME <options> URI    ("$PROG_NAME URI" does everything except -E)
   special invocations:
 
      -t, --starttls <protocol>     does a default run against a STARTTLS enabled <protocol>
+                                   a differed port may be specified with <protocol>:<port>
      --xmpphost <to_domain>        for STARTTLS enabled XMPP it supplies the XML stream to-'' domain -- sometimes needed
      --mx <domain/host>            tests MX records from high to low priority (STARTTLS, port 25)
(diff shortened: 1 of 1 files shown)


drwetter:
Hmm, that it fails if no port is specified was more or less on purpose -- maybe the error is not nice.

The ports are sometimes not that obvious, e.g. SMTP+STARTTLS could be also 587.

thecky:
>The ports are sometimes not that obvious, e.g. SMTP+STARTTLS could be also 587.

For this it's also possible to assign the port with --startssl=smtp:587
IMO without this port assignment --starttls will always fail, because the the used port in the -connect statement is always 443 (well, if you use --mx domain - testssl.sh sets PORT to 25 and starttls works)

drwetter:
mind the URI, one part of it is the port -- otherwise one would also for plain TLS enabled ports not be able to connect at anything different than 443.

PeterMosmans:
I'd rather not have this included in the current script. This will break compatibility (as the syntax changes), and removes the option for users to specify their own port.

thecky:
Well, my fault - I truely have overlook the URI part. I'll revert this one.

PR #275, opened 2016-01-28
Detect multiple server certificates

dcooper16:
Thanks for the diff file. I has already created such a file and used it to get started merging my changes into the latest version. It just took me a while to get this finished since I was doing a lot of testing, fixing a few minor bugs, and making other minor improvements to the code.


drwetter:
Very cool, thx!

Merged.

Patch with minor cleanups followed. I also changed in there the indentation a bit: If there's only one certificate, it'll look like before.

PR #277, opened 2016-01-31
fixed output in ciphertest with --color=1

thecky:
When --color=1 is used, the output in the ciphertest is missing some spaces between the
KeyExch. and Encryption columns. This is a result of the pr_off() function.
This commit add an additional check in neat_list() and insert the missing
spaces.


drwetter:
Hi Thomas 

Thx ! I am afraid though I have not committed a fix I did a while back which also fixes other issues :-/

cheers, Dirk

Issue #278, opened 2016-02-01
Chain of trust problem


drwetter:
Hi @dcooper16.

it seems that ``openssl verify`` is not as reliable as just doing a ``openssl s_client connect`` and then look out for the return code  / verify (ok).

I have a problem with a self signed certificate which is not marked as such in the "Chain of trust' (try localhost:631 if you're running CUPS) And e.g. for the host ``testssl.sh`` the Java keystore fails, thus marking the whole site as not trusted.

Unless you have a better idea I would revert that part back to the old check.

Cheers, Dirk

drwetter:
Partly SSL_CERT_DIR seems to be the "culprit", at least for testssl.sh and the Java store. 

Where do you have that variable from? In strace I see that the system store is being used the root StartCom certificate is being added if SSL_CERT_DIR is not set to /dev/null. That root CA doesn't seem to be in java.pem.

So there I guess should probably remove the java store.

For the second case it looks the ret code is not properly parsed:

```
mylaptop:/tmp/ssltester.J3VrVa|% ~/testssl/bin/openssl.Linux.x86_64 verify -purpose sslserver -CAfile ~/testssl/etc/java.pem  host_certificate.txt 
host_certificate.txt: CN = <myselfsignedhost-CN>
error 18 at 0 depth lookup:self signed certificate
OK
mylaptop:/tmp/ssltester.J3VrVa|% echo $?                                                     
mylaptop:/tmp/ssltester.J3VrVa|%           
```

dcooper16:
Okay, it seems that for the verify command, I need to parse the text that is output by the command rather than relying on the return code.

Most of the time when I've been running my tests I've only had a single file with my own locally-generated root certificate in $INSTALL_DIR/etc/*.pem, but I noticed that "Chain of trust" always said "Ok" for any public servers that I tried. So, I knew that OpenSSL had to be using a system store of trusted certificates in addition to those provided by the "-CAfile" option. After searching through the source code I found it was using SSL_CERT_DIR as the local of the system store and setting that to /dev/null forced OpenSSL to only use the trusted certificates included in "-CAfile"

drwetter:
OT: yes, after grepping -r through the source tree I spotted the somewhat hidden reference to X509_get_default_cert_dir_env() too.

Just wondering whether SSL_CERT_FILE shouldn't be dev-nulled too, just to be safe.

drwetter:
Good work!

PR #279, opened 2016-02-01
Fix chain of trust problem

dcooper16:
This should fix issue #278. I'm not sure whether openssl verify will ever print out more than one error, so to be safe, I wrote the code to handle the possibility that it might; if there is more than one error, it just takes the first and ignores the rest.


drwetter:
That fixed it. Thanks!

PR #280, opened 2016-02-01
Number of trusted certificate files in $INSTALL_DIR/etc/*.pem

dcooper16:
The number of .pem files in $INSTALL_DIR/etc is currently hard-coded into determine_trust. This modifies the code so that the number of files can be changed without having to change the code.


drwetter:
As I was kicking out the java store today I thought actually the same but didn't do it ;-) Thx! :+1: 

I did something else I cannot push anymore ... ;-(

PR #286, opened 2016-02-07
fix typo in b93fc824

tjconcept:
I think this is a typo.

Saw an error when I tried a test using the latest from `master` so I skimmed the source code and assumed the `debug` should've been `debugme` - changing it removed the warning.


drwetter:
Thx!

PR #292, opened 2016-02-09
Detect multiple certificates cleanup

dcooper16:
This corrects the indentation within determine_trust() when there are multiple certificates and the output for "Chain of trust (experim.)" takes up more than one lines.

In addition, it fixes the ID field of the JSON output for entries related to the certificate. At the moment, each ID string begins with a blank space. This changes it to remove the space if there is one certificate and to add "Server Certificate #X" at the beginning of each ID if there is more than one certificate.

Perhaps there's a better way than just using, for example, "Server Certificate #1 key_size" as a way to distinguish multiple "key_size" entries in the JSON file. This is just one idea, and it can certainly be changed if those who intend to use the JSON output prefer something else.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2658,7 +2658,7 @@ verify_retcode_helper() {
 }
 
 determine_trust() {
-	local heading=$1
+	local json_prefix=$1
 	local -i i=1
 	local -i num_ca_bundles=0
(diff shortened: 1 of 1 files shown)


drwetter:
Thx, @dcooper16 !

PR #300, opened 2016-02-24
Consistent case severity codes

Dude4Linux:
In using scripts with testssl.sh, I found that the case of severity codes was not always consistent, e.g. 'info' vs 'INFO'.  I tried to find all inconsistent use of case in the severity codes and informative comments.


drwetter:
Good, thank you!

Dude4Linux:
@drwetter, You're welcome.

PR #311, opened 2016-03-07
Format readme for better readability

noqqe:
Hi,

just noticed that the formatting of the readme is not really nice to read in terminal.
I did a quick reformat to make headlines valid (with a space) and rearranged textblocks to 
match 80 chars per line limit.


drwetter:
thx!

PR #326, opened 2016-03-25
run_allciphers() speedup

dcooper16:
The run_allciphers() function currently works by calling "$OPENSSL s_client" once for each cipher suite supported by $OPENSSL. In the case of "OpenSSL 1.0.2-chacha (1.0.2e-dev)" that means 195 calls to "$OPENSSL s_client" even though servers tend to only support a small fraction of these cipher suites.

This PR produces the same output as the current run_allciphers() with fewer calls to "$OPENSSL s_client", which results in the function running faster (usually much faster). The basic idea behind the revised function is to test cipher suites in blocks. If $OPENSSL supports 195 cipher suites, then it groups these cipher suites into 4 blocks of 64 (with the final block being smaller). It makes one call to "$OPENSSL s_client" with cipher suites 1-64, and if it fails, then it knows that none of these 64 cipher suites are supported by the server and it doesn't need to perform any more tests on these 64 cipher suites. If it succeeds, then it breaks the 64 cipher suites into 4 blocks of 16 and calls "$OPENSSL s_client" with each of those blocks. The blocks of 16 that are successful are broken into blocks of 4, and for each of the successful blocks of 4 the individual cipher suites are tested.

For testssl.sh and www.google.com the number of calls to "$OPENSSL s_client" is reduced from 195 to 88. For github.com the number of calls is reduced to 56!

I haven't made any changes to run_cipher_per_proto yet, but if this PR is accepted I can make the same changes in that function.

Thanks,

David

P.S. While writing this code, I corrected one error that appears in the current versions of run_allciphers() and run_cipher_per_proto(). Currently, there is a call to "$OPENSSL x509 -noout -text -in $TMPFILE" for each cipher suite if $SHOW_SIGALGO is true. This call fails, however, if $SHOW_EACH_C is also true and the cipher being processed isn't supported by the server. So, I changed the "if" line so that "$OPENSSL x509 -noout -text -in $TMPFILE" is only called if $SHOW_SIGALGO is true and the cipher is supported by the server.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1570,51 +1570,102 @@ test_just_one(){
 # test for all ciphers locally configured (w/o distinguishing whether they are good or bad
 run_allciphers(){
      local tmpfile
-     local nr_ciphers
-     local -i sclient_success=0
-     local hexcode n ciph sslvers kx auth enc mac export
+     local -i nr_ciphers
+     local n sslvers auth mac export
+     local -a hexcode ciph kx enc export2
+     local -i i j parent child end_of_bundle round_num bundle_size num_bundles mod_check
+     local -a ciphers_found
      local dhlen
      local available
+     local ciphers_to_test
 
-     nr_ciphers=$(count_ciphers "$($OPENSSL ciphers 'ALL:COMPLEMENTOFALL:@STRENGTH' 2>$ERRFILE)")
+     # get a list of all the cipher suites to test (only need the hexcode, ciph, kx, enc, and export values
+     nr_ciphers=0
+     $OPENSSL ciphers -V 'ALL:COMPLEMENTOFALL:@STRENGTH' 2>>$ERRFILE | (while read hexcode[nr_ciphers] n ciph[nr_ciphers] sslvers kx[nr_ciphers] auth enc[nr_ciphers] mac export2[nr_ciphers]; do
+         nr_ciphers=$nr_ciphers+1
+     done
+
      outln
      pr_headlineln " Testing all $nr_ciphers locally available ciphers against the server, ordered by encryption strength "
      "$HAS_DH_BITS" || pr_warningln "    (Your $OPENSSL cannot show DH/ECDH bits)"
      outln
      neat_header
 
-     $OPENSSL ciphers -V 'ALL:COMPLEMENTOFALL:@STRENGTH' 2>>$ERRFILE | while read hexcode n ciph sslvers kx auth enc mac export; do
-     # FIXME: e.g. OpenSSL < 1.0 doesn't understand "-V" --> we can't do anything about it!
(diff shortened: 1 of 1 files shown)


drwetter:
Hi Dave,

jolly good idea! First tests look also fine and give for the WAN a considerable speed improvement. Whohoo!

In the LAN however it looks like the old version is still 10-20 faster.

NEW  | OLD
---------|-------
8.857 |  6.573
8.885 | 6.557
6.194  |  5.402
4.810 | 4.184
5.812 5.| 5.397

Measurement with ``testssl -e <hostname>``. The first two are the same hosts.

Any idea why this is so?

Thx again, Dirk

drwetter:
The most dramatic decrease I spotted so far is e.g. rc4.badssl.com. 

Time old: 51.6 seconds, new: 9.1 seconds.

dcooper16:
Hi Dirk,

I don't know why you're seeing slower times on the LAN. The revised version would be slower if the server supported a lot of cipher suites that were spread across the list that is being tested. Not only could the result in more calls to "$OPENSSL s_client" than the current version, but the revised version does more internal computations.

I've been using a version of the code that outputs the number of calls made to "$OPENSSL s_client" and so far the revised version has always made fewer calls and run faster. Even when I set my test server with "SSLCipherSuite ALL:COMPLEMENTOFALL", the revised version only called  "$OPENSSL s_client" 172 times, and it ran a few seconds faster than the current version.

In practice, since the cipher suites are ordered by strength, I would hope that the revised code would just quickly determine that none of the cipher suites near the end of the list are supported.

I've been running testssl.sh within a virtual machine on my desktop and have been testing it against an instance of Apache running on the same virtual machine (so that the URI I specify is "localhost") and yet the current version of testssl.sh with the "-e" option takes 40 to 45 seconds to run. So, I can't even begin to replicate the results you're seeing where old version is running in well under 10 seconds. Maybe in your LAN tests, each of the "$OPENSSL s_client" tests is so fast that the cost of some extra tests is less than the cost of the extra processing time in the revised code (e.g., to construct each of the $ciphers_to_test strings.

David

drwetter:
Hi Dave,

I'll certainly integrate your code as for the majority of users it's a significant improvement. Thank you for this, much appreciated!

For the LAN issue: please let me do before some further testing or maybe have others test this (wink!) 

FYI: The first speed tests I did were on a client which is like a dog, an old Core2Duo. LAN is Gigabit and servers are some fast VMs on beefy 2 year old system and bare metal systems.. Despite the old client hardware: Out of my guts I can't imaging either think the processing time is the important factor. So either we are wrong with that assumption or there's something else.

What I could think of and if it's worth it: having additionally my legacy code in there which is not the default and can be called by a switch.

Thx, Dirk

drwetter:
thx again, @dcooper16 !

I'll modify it later to have the old code there for testing.

PR #328, opened 2016-03-28
Modified pr_functions to take a second argument / commentet out unused pr_functions

thecky:
This PR modify the pr_functions to take an optional second argument. This will be used later to print the "checkname" and the result with one function/in one line.

I also commentet out the pr_blue/grey functions - they're currently unused.


thecky:
There's still a bug...I make an new PR ;)

drwetter:
Thomas,

I don't get it, you gotta help me.

thecky:
Hi Dirk,

till now you do sth like this for the output:
pr_bold " TLS 1.2    ";
...codelines...
pr_done_bestln "offered (OK)"

With the second arg for (in this example pr_done_bestln) we could write:
...codelines...
pr_done_bestln "offered (OK)" " TLS 1.2   "

This is imo neccessary for the severity output.
The third commit (6dcf024) just fix the bugs which I did in the first commit.

drwetter:
@thecky: Sorry, I have another solution in mind -- let's dicuss this after next release.

PR #330, opened 2016-03-29
Fix typo in mapping-rfc.txt

dcooper16:
There is a tab on the line for SSL_CK_RC2_128_CBC_WITH_MD5. When testssl.sh is called with "-E" and "--show-each," this causes the string "not a/v" to be printed two characters to the right of the same string on every other line (at least on Linux systems). This PR just deletes the tab character.


drwetter:
:+1:   Thx for spotting and your PR

PR #335, opened 2016-03-31
run_cipher_per_proto speedup

dcooper16:
*** Updated to align with yesterday's changes to run_allciphers() ***

This PR makes basically the same changes to run_cipher_per_proto() as I previously made to run_allciphers(). The main difference is that in this function, round 0 consists of a single call to "$OPENSSL s_client" with "-cipher" including all of the locally supported ciphers. The reason for the difference is that in run_allciphers() its saves time to assume the server supports at least one cipher suite. In the case of run_cipher_per_proto(), however, it is likely that the server will not support some protocols at all, so its usually faster to start with a single call to "$OPENSSL s_client" that tests whether the server supports the protocol at all.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1670,11 +1670,14 @@ run_allciphers() {
      return 0
 }
 
-# test for all ciphers per protocol locally configured (w/o distinguishing whether they are good or bad
-run_cipher_per_proto(){
+# test for all ciphers per protocol locally configured (w/o distinguishing whether they are good or bad)
+run_cipher_per_proto() {
      local proto proto_text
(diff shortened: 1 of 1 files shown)


drwetter:
Thx David. That test was anyway a dog and you brought the time needed down (old/new):

rc4.badssl.com:  181 / 54 seconds
(internal host 1):     17./ 16 seconds
(internal host 2):     21-24 / 24+24 seconds
dev.testssl.sh           57/ 24,5 seconds
facebook.com          185 / 87 seconds
cloudflare.com         45 / 20 seconds

Thx again for the good work!

PR #338, opened 2016-04-08
Fix typo in neat_list()

dcooper16:
The last line of neat_list currently uses $HEXC as the parameter to show_rfc_style(), but it should use $hexcode. At the moment using $HEXC instead of $hexcode makes no difference, since hexcode="$1" and in all calls to neat_list() the first parameter is $HEXC. However, this bug could create problems in the future since neat_list() will misbehave if the value of the first parameter (hexcode) isn't the same as $HEXC.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1502,7 +1502,7 @@ neat_list(){
           done
      fi
      #echo "${#kx}"                            # should be always 20 / 13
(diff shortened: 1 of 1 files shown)


drwetter:
thx for spotting and fixing!

PR #341, opened 2016-04-11
run_allciphers(),run_cipher_per_proto(), and SSLv2

dcooper16:
This PR addresses two problems related to SSLv2 in run_allciphers() and run_cipher_per_proto().

In run_cipher_per_proto(), the call to "$OPENSSL s_client" is changed to that $SNI is not included if $proto is -sslv2 or -sslv3. As noted in a comment within run_prototest_openssl(), "newer openssl throw an error if SNI is supplied with SSLv2" and "SSLv3 doesn't have SNI (openssl doesn't complain though -- yet)."

run_allciphers() will sometimes incorrectly report that a server supports an SSLv2 cipher, even if the server does not support SSLv2 at all. The problem occurs if there is a supported SSLv3 cipher suite that has the same OpenSSL name as an SSLv2 cipher suite (e.g., RC4-MD5). Since the call to "$OPENSSL s_client" only uses the OpenSSL name, but the results report both the name, hexcode, and RFC cipher suite name, both the SSLv2 and SSLv3 cipher suites are reported as being supported (e.g., 0x04=RC4-MD5=TLS_RSA_WITH_RC4_128_MD5 and x010080=RC4-MD5=SSL_CK_RC4_128_WITH_MD5). This PR fixes the problem by testing SSLv2 cipher suites separately from non-SSLv2 cipher suites.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1571,16 +1571,18 @@ test_just_one(){
 run_allciphers() {
      local tmpfile
      local -i nr_ciphers=0
-     local n sslvers auth mac export
-     local -a hexcode ciph kx enc export2
+     local n auth mac export
+     local -a hexcode ciph sslvers kx enc export2
      local -i i j parent child end_of_bundle round_num bundle_size num_bundles mod_check
      local -a ciphers_found
      local dhlen
      local available
      local ciphers_to_test
+     local sslv2_locally_supported=false sslv2_supported=false
 
-     # get a list of all the cipher suites to test (only need the hexcode, ciph, kx, enc, and export values)
-     while read hexcode[nr_ciphers] n ciph[nr_ciphers] sslvers kx[nr_ciphers] auth enc[nr_ciphers] mac export2[nr_ciphers]; do
+     # get a list of all the cipher suites to test (only need the hexcode, ciph, sslvers, kx, enc, and export values)
+     while read hexcode[nr_ciphers] n ciph[nr_ciphers] sslvers[nr_ciphers] kx[nr_ciphers] auth enc[nr_ciphers] mac export2[nr_ciphers]; do
+          [[ "${sslvers[nr_ciphers]}" == "SSLv2" ]] && sslv2_locally_supported=true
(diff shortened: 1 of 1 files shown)


drwetter:
Hi David,

there's a global variable named HAS_SSL2 or so. Can't this be used instead of ssl2_locally_supported?

There's also HAS_SSL3 as I can't tell whether all distrubutors still have it enabled.
This needs to be checked before a openssl call with SSLv3 to play safe.


Thx, Dirk


-- 
Set from my mobile. Excuse my brevity&typos

dcooper16:
I changed the code to use the global $HAS_SSL2 rather than $sslv2_locally_supported.

I don't think there's a need to use $HAS_SSL3 in run_allciphers(), since the call to "$OPENSSL s_client" for non-SSLv2 ciphers does not specify a protocol. It's also not needed in run_cipher_per_proto(), since there is already a call to locally_supported() before anything further is done with a protocol.

dcooper16:
In further testing I discovered another change that needed to be made.

run_all_ciphers_per_proto() calls "$OPENSSL ciphers $proto -V 'ALL:COMPLEMENTOFALL:@STRENGTH'" for each protocol. The problem is that in versions of OpenSSL prior to 1.1.0, the ciphers function did not understand the "-tls1_1" or "-tls1_2" options. So, for $proto equal to "-tls1_1" or "-tls1_2" the protocol was just ignored and all ciphers for all protocols were listed, including SSLv2 ciphers. Since "$OPENSSL s_client" only uses cipher names, this could lead to false positives. For example, if the server supports RC4-MD5 (0x04), then the output for TLS1.1 and TLS1.2 would incorrectly list both RC4-MD5 (0x04) and RC4-MD5 (0x010080).

The change I just made fixes that by using "-tls1" in the $OPENSSL ciphers" call if $proto is  "-tls1_1" or "-tls1_2" and OpenSSL 1.1.0 isn't being used.

drwetter:
Ok, thx. For further comments see #345.

PR #342, opened 2016-04-13
More extensions in socksend_tls_clienthello()

dcooper16:
This PR adds the signature algorithms, heartbeat, session ticket, and next protocol extensions to the client hello message created by socksend_tls_clienthello() for TLS 1.0 and above. It also adds the supported elliptic curves and ec points format extensions if the client hello message includes any ECC cipher suites.

I tested this version against several servers with $EXPERIMENTAL set to true and get the same results as with the current code with $EXPERIMENTAL set to false.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4272,10 +4272,16 @@ socksend_tls_clienthello() {
           ,00                      # server_name type (hostname)
           ,00, $len_servername_hex # server_name length. We assume len(hostname) < FF - 9
           ,$servername_hexstr      # server_name target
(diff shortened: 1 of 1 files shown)


drwetter:
Thx, David!

One request: On ssllabs there's recently often the EC 1d (curve25519) mentioned, see https://github.com/ssllabs/ssllabs-scan/issues/348#issuecomment-224507590. Also Windows 2016 seems to come with that. Could you just add it to the ClientHello?

Cheers, Dirk

PR #345, opened 2016-04-29
More SSLv2 (and SSLv3) related fixes

dcooper16:
In doing some work on cipher_pref_check() I noticed that it was failing on SSLv2 since the call to "$OPENSSL s_client" includes SNI. I've also noticed in my testing that "$OPENSSL s_client" will not connect to an SSLv2-only server unless the "-ssl2" flag is included. So, I carefully checked each call to "$OPENSSL s_client" in the program (other than in run_allciphers and run_cipher_per_proto, since those functions are already addresses in PR #341) to see whether they would inappropriate fail with an SSLv2-only (or SSLv3-only) server.

As a general rule, if the call doesn't currently include the protocol, then I added "-ssl2" if $OPTIMAL_PROTO is "-ssl2", indicating that the server only supports SSLv2, and I removed any $SNI if a protocol is specified if a protocol is specified and it is either SSLv2 or SSLv3.

I tested it on an SSLv2-only server, and the results are much better. I also tested it on a collection of other servers, none of which support SSLv2, and the results are the same as with the current code.

The only thing I haven't been able to test is how the revised code works when the "--starttls" option is used. I don't believe the changes I made would cause anything to break in that case, but I also don't think code will work any better in that case, if the server only supports SSLv2. Of course, since no server should support SSLv2 (let alone only SSLv2), it shouldn't really be an issue.

One thing that I did not change, but that I do not understand; why does determine_optimal_proto() try the protocols in the order "-tls1_2 -tls1 -ssl3 -tls1_1 -ssl2" rather than "-tls1_2 -tls1_1 -tls1 -ssl3 -ssl2"? Doesn't the current ordering imply that TLS v1.0 and SSLv3 are better than TLS v1.1?


drwetter:
The order is just a probability.

How likely is a SSLv2 server?

dcooper16:
It does seem very unlikely that there would still be any SSLv2 only servers out there.

There are, however, some changes in this PR that affect servers that implement SSLv2 in addition to other protocols, or even servers that don't implement SSLv2 at all. For example, on one server I tested, the results for the RC4 test are:
```
 Checking for vulnerable RC4 Ciphers 

 RC4 (CVE-2013-2566, CVE-2015-2808)        VULNERABLE (NOT ok): ECDHE-RSA-RC4-SHA RC4-SHA RC4-MD5 RC4-MD5
```
`RC4-MD5` appears twice since both the SSLv2 (x010080) and SSLv3 (x04) cipher suites are listed, even though the server does not support SSLv2.

If you'd like, we can close this PR and I can submit a few smaller PRs that address individual functions.

Note that PR #341 is related to this one, and also applies to servers that do not implement SSLv2. For example, on a server that only supports RC4-MD5 (and that does not support SSLv2) testssl.sh returns the following:
```
 Testing all 195 locally available ciphers against the server, ordered by encryption strength 

Hexcode  Cipher Suite Name (OpenSSL)       KeyExch.  Encryption Bits     Cipher Suite Name (RFC)
---------------------------------------------------------------------------------------------------------------------------
 x04     RC4-MD5                           RSA        RC4       128      TLS_RSA_WITH_RC4_128_MD5                           
 x010080 RC4-MD5                           RSA        RC4       128      SSL_CK_RC4_128_WITH_MD5  
```
```
 Testing all locally available ciphers per protocol against the server, ordered by encryption strength 

Hexcode  Cipher Suite Name (OpenSSL)       KeyExch.  Encryption Bits     Cipher Suite Name (RFC)
---------------------------------------------------------------------------------------------------------------------------
SSLv2 
SSLv3 
 x04     RC4-MD5                           RSA        RC4       128      TLS_RSA_WITH_RC4_128_MD5                           
TLS 1 
 x04     RC4-MD5                           RSA        RC4       128      TLS_RSA_WITH_RC4_128_MD5                           
TLS 1.1 
 x04     RC4-MD5                           RSA        RC4       128      TLS_RSA_WITH_RC4_128_MD5                           
 x010080 RC4-MD5                           RSA        RC4       128      SSL_CK_RC4_128_WITH_MD5                            
TLS 1.2 
 x04     RC4-MD5                           RSA        RC4       128      TLS_RSA_WITH_RC4_128_MD5                           
 x010080 RC4-MD5                           RSA        RC4       128      SSL_CK_RC4_128_WITH_MD5
```
PR #341 fixes it so that the SSLv2 ciphers are not incorrectly listed.

drwetter:
Took some time to look myself at the difference as I didn't have a server for comparison. #341 is merged now. But it didn't address the double RC4-MD5 in the RC4 test.

Most easiest fix would be to just remove one RC4. But if a server in fact offers x04 and x010080 the output is still wrong.

Otherwise for now I would rather prefer not to merge this if it addresses SSLv2-only servers.

dcooper16:
Okay, let's abandon this PR and I'll submit ones smaller ones that address specific issues.

testssl.sh
```diff
@@ -6951,7 +6951,7 @@ run_rc4() {
      local hexcode dash rc4_cipher sslvers kx auth enc mac export
      local rc4_ciphers_list="ECDHE-RSA-RC4-SHA:ECDHE-ECDSA-RC4-SHA:DHE-DSS-RC4-SHA:AECDH-RC4-SHA:ADH-RC4-MD5:ECDH-RSA-RC4-SHA:ECDH-ECDSA-RC4-SHA:RC4-SHA:RC4-MD5:RC4-MD5:RSA-PSK-RC4-SHA:PSK-RC4-SHA:KRB5-RC4-SHA:KRB5-RC4-MD5:RC4-64-MD5:EXP1024-DHE-DSS-RC4-SHA:EXP1024-RC4-SHA:EXP-ADH-RC4-MD5:EXP-RC4-MD5:EXP-RC4-MD5:EXP-KRB5-RC4-SHA:EXP-KRB5-RC4-MD5"
```
drwetter: That still looks the same (2x RC4-MD5).

drwetter:
Most important to me is the fix for run_rc4(). Did you remove the fix in d2f1e15  by accident?

drwetter:
Sorry, I was too fast , didn't see #424.

drwetter:
Thank you  for cleaning this up and sorry for letting it sit here for some time

PR #346, opened 2016-05-06
Additional checks in run_protocols()

dcooper16:
One server I am testing responds to an SSLv3 ClientHello with TLSv1.2. If tls_sockets is being used, then testssl.sh responds with "#FIXME: downgraded. still missing a test case here." This PR fixes that, and in general checks the responses in run_protocols() more closely.

If tls_sockets is being used and the connection fails even though the server supports an earlier version of SSL/TLS, then it flags an error. If tls_sockets returns 2, then it verifies that $DETECTED_TLS_VERSION is equal to the highest version number supported by the server (that is also less than the version number in the ClientHello).

In addition, in order to test servers' support for version negotiation, it adds a new test that sends a TLSv1.4 ClientHello and verifies that the server responds with the highest version number that it supports. (This test only runs if both $using_sockets and $EXPERIMENTAL are true and server actually supports some version of SSL/TLS other than SSLv2.)


drwetter:
Hi David,

good that you're bringing things forward. "Unfortunately" your PRs are no no-brainers and require testing. For the latter I need a time slot. Gimme a bit...

Thx for the work, Dirk

dcooper16:
No problem. I realise that the PR related to SSLv2 may be of very limited use. I still haven't found any servers that still support SSLv2, let alone only SSLv2. I had to compile an old version of Apache along with an old version of OpenSSL before I could do any actual testing with SSLv2.

The main motivation for this PR is that the CUPS server on the virtual machine in which I am doing the tests returns an TLSv1.2 ServerHello in response to an SSLv3 ClientHello and testssl.sh currently responds with "#FIXME: downgraded. still missing a test case here."

drwetter:
David, that looks very good. Sorry for letting this sit for a while.

I found the TLS 1.4 version tolerance particular important for the current release -- apple.com and ibm.com fail. 

Just wondering whether it makes sense to have a test for TLS 1.3. Is there any server out there yet?

 Cheer, Dirk

drwetter:
Thx!

drwetter:
I enabled the tolerance check by default as I believe it's reliable enough and it has a huge benefit.

dcooper16:
> Just wondering whether it makes sense to have a test for TLS 1.3. Is there any server out there yet?

None that I've been able to test against. https://github.com/tlswg/tls13-spec/wiki/Implementations mentions two servers that supposedly implement some draft version of TLS 1.3 (not the latest draft), but neither of them work for me. When I try tls13.cloudflare.com, it downgrades to TLS 1.2, and tls13.crypto.mozilla.org seem to support SSL/TLS at all.

drwetter:
Ok. thx.

Just wondering whether it makes sense to add an (internal) check for TLS 1.3 and **if** there's a difference to display it.

At the moment I don't think this has much benefit or makes a difference but looking into the future it might matter -- if TLS 1.3 ever will be deployed. Looking at totally unimportant servers ;-) like apple.com, ibm.com I think it'll take some time..

What's your stance?

dcooper16:
I think it's too early to try to add a specific check for TLS 1.3. https://tls13.cloudflare.com claims to support browsers that implement TLS 1.3 draft 13, but any time I try to connect I get TLS 1.2 in the response. I even have a local copy of testssl.sh in which I have tried to implement the ClientHello according to draft-ietf-tls-tls13, but https://tls13.cloudflare.com responds to the ClientHello with TLS 1.2.

I'm a bit surprised that the "Version tolerance" code turned out to be useful. I thought this would only be a problem with older servers, and that they would be caught when they failed with TLS 1.2 ClientHello's rather than downgrading.

I turns out, however, that there is a discussion going on right now on the TLS WG mail list, and according to Hubert Kario's testing "there are about 3-4% of servers which are outright incompatible to TLSv1.3 version in Client Hello." (See https://mailarchive.ietf.org/arch/msg/tls/1hRdN-FRAVAAIVSw-gwCFEE5oSA)

drwetter:
Ok!

Hubert Kario still sees a difference in TLS 1.3 and TLS 1.4  compatibility, that's also what Ivan Ristic told me today (though SSLlabs tests TLS 1.3 and TLS 1.152 but I guess 1.152 is as odd as 1.4). The difference reported from Hubert is minor 170 /4300 vs. 183/4300. 

13/4300 is 0.3% but at least it seems that testing TLS 1.4 only is by this statistics better than 1.3

Issue #348, opened 2016-05-11
../testssl.sh: line 4343: [[: 2F99: value too great for base (error token is "2F99")


matteocorti:
When testing uc13.ethz.ch I get

`../testssl.sh: line 4343: [[: 2F99: value too great for base (error token is "2F99")
`
Debugging output:

```
  SSLv3      sending client hello...
"\x16\x03\x00\x00\xcb\x01\x00\x00\xc7\x03\x00\x54\x51\x1e\x7a\xde\xad\xbe\xef\x31\x33\x07\x00\x00\x00\x00\x00\xcf\xbd\x39\x04\xcc\x16\x0a\x85\x03\x90\x9f\x77\x04\x33\xd4\xde\x00\x00\xa0\xc0\x14\xc0\x0a\xc0\x22\xc0\x21\xc0\x20\x00\x39\x00\x38\x00\x37\x00\x36\x00\x88\x00\x87\x00\x86\x00\x85\xc0\x0f\xc0\x05\x00\x35\x00\x84\xc0\x13\xc0\x09\xc0\x1f\xc0\x1e\xc0\x1d\x00\x33\x00\x32\x00\x80\x00\x81\x00\x82\x00\x83\x00\x31\x00\x30\x00\x9a\x00\x99\x00\x98\x00\x97\x00\x45\x00\x44\x00\x43\x00\x42\xc0\x0e\xc0\x04\x00\x2f\x00\x96\x00\x41\x00\x07\xc0\x11\xc0\x07\x00\x66\xc0\x0c\xc0\x02\x00\x05\x00\x04\xc0\x12\xc0\x08\xc0\x1c\xc0\x1b\xc0\x1a\x00\x16\x00\x13\x00\x10\x00\x0d\xc0\x0d\xc0\x03\x00\x0a\x00\x63\x00\x15\x00\x12\x00\x0f\x00\x0c\x00\x62\x00\x09\x00\x65\x00\x64\x00\x14\x00\x11\x00\x0e\x00\x0b\x00\x08\x00\x06\x00\x03\x00\xff\x01\x00"
reading server hello...
00000000  30 2f 99 16 00 00 00 00  50 a0 42 0b 00 00 00 00  |0/......P.B.....|
00000010  21 16 a0 d4 d9 b9 b1 8b  e2 d3 ee 4f 0d c3 d8 0d  |!..........O....|
[...]

Issue #340, opened 2016-04-11
False positive from parse_tls_serverhello()


dcooper16:
I was just testing some code I wrote against an internal web server that I hadn't tried before, and it reported false-positive results. The problem seems to be the error response from the server. When tls_sockets() is called for SSLv3 (tls_sockets "00" "$TLS_CIPHER"), which the server does not support, the response is 15030000020228. However, when tls_sockets() is called with a TLS version that the server supports, but with a list of cipher suites that it doesn't support (e.g., tls_sockets 03 c0,3c), the response is: 1503030002017015030300020228.

The current version of parse_tls_serverhello() will always return 0 if the response begins with an alert that is an unrecognized name warning; it just assumes that this warning is followed by a server hello handshake message. In tls_sockets(), $lines is set to 2, so the check in tls_sockets() that the number of lines in the response isn't 1 doesn't catch the error either.

I don't know much about the server, other than "wget -S" reports that it is an Apache server.

I don't know whether this is a bug worth fixing. At the moment, it seems unlikely that one of the calls to tls_sockets() would fail as a result of the client hello not including any cipher suites supported by the server. The only way I can induce the bug using the current testssl.sh code is using the "--devel" option (e.g. bash testssl.sh --devel 03 "c0,3c" URI). The code that I've been developing at https://github.com/dcooper16/testssl.sh/blob/extended_tls_sockets/testssl.sh doesn't have this problem, since the parse_tls_serverhello() in that branch explicitly verifies that the response includes a server hello message.

Anyways, here is a dump of running  bash testssl.sh --devel 03 "c0,3c" URI: 
[debug3.txt](https://github.com/drwetter/testssl.sh/files/213378/debug3.txt)

Thanks,

Dave

drwetter:
closing via #353.

Thx!

PR #353, opened 2016-05-17
parse_tls_serverhello() with added checks

dcooper16:
This PR provides a revised version of parse_tls_serverhello() that should be more robust in terms in variations in the response provided by the server and that includes several checks to verify that the response is valid. This should address issues #348 and #340.

I tested this code against quite a few servers (more than 50) and verified that the results are the same as with the current code.

This version of parse_tls_serverhello() is a simplified version of the one that I am working on at https://github.com/dcooper16/testssl.sh/blob/extended_tls_sockets/testssl.sh, which parses a lot more of the server's response, and so also includes many more validity checks.


drwetter:
Thx very much, David! :+1: 

I have a request: Could pls add your contributions so far as PR to CREDITS.md? That'll make my life easier. :-)

PR #356, opened 2016-05-19
Fix typo in Server key size check

dcooper16:
When certificate_info() is trying to determine what type of public key the server has so that it can determine whether the key size is acceptable, it sometimes looks at $cert_sig_algo rather than $cert_key_algo. This PR fixes that and also adds support for DSA public keys.


drwetter:
thx again

PR #357, opened 2016-05-19
Recognize more signature algorithms

dcooper16:
This PR adds to the list of signature algorithms recognized in certificate_info().


drwetter:
Thx!

Minor thing: "grep -A, head, tail" can probably done better with  ``awk '/pattern1/,/pattern2/ { print $0 }'``

Issue #332, opened 2016-03-30
Output for PFS is confusing


dcooper16:
There seems to be four different modes for printing the results of run_pfs when PFS is offered, depending on whether the --show-each and --wide flags are set.

The output seems okay is the --wide flag is set or if the --show-each flag is not set, but if --show-each is set and --wide is not, the output is confusing.

If neither flag is set, the output is "PFS is offered (OK)" in green, followed by a list of **supported** PFS cipher suites (in black).

If --show-each is set and --wide is not, the output is "PFS is offered (OK)" in green, followed by a list of **all** PFS cipher suites (in black), which I think is confusing.

I believe that for the case in which --show-each is set and --wide is not, either the --show-each flag should be ignored for this output line or something should be done to visually distinguish between the cipher suites that are supported and those that are not.

Thanks,

David

drwetter:
Am 30. März 2016 16:43:07 MESZ, schrieb David Cooper <notifications@github.com>:
>There seems to be four different modes for printing the results of
>run_pfs when PFS is offered, depending on whether the --show-each and
>--wide flags are set.
>
>The output seems okay is the --wide flag is set or if the --show-each
>flag is not set, but if --show-each is set and --wide is not, the
>output is confusing.
>
>If neither flag is set, the output is "PFS is offered (OK)" in green,
>followed by a list of **supported** PFS cipher suites (in black).
>
>If --show-each is set and --wide is not, the output is "PFS is offered
>(OK)" in green, followed by a list of **all** PFS cipher suites (in
>black), which I think is confusing.
>
>I believe that for the case in which --show-each is set and --wide is
>not, either the --show-each flag should be ignored for this output line
>or something should be done to visually distinguish between the cipher
>suites that are supported and those that are not.

You're right. --show-each alone is BS in the PFS section. It could make sense if other ciphers would be greyed out. I'll fix that later if I am home.

Thx, Dirk




-- 
Set from my mobile. Excuse my brevity&typos

PR #361, opened 2016-05-24
run_pfs() and run_rc4() show each fixes

dcooper16:
When run_rc4() is run with the "--show-each" option, but without the "--wide" option, a list of all RC4 ciphers is printed, without any distinction between those that are supported by the server and those that are not. This is the same issue I noted in #332 for run_pfs().

In run_pfs(), the displayed output was corrected, but all ciphers were still being added to $pfs_ciphers, so the list of supported PFS ciphers sent to fileout() was incorrect.

This PR fixes both issues.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3679,7 +3679,7 @@ run_pfs() {
                          out "$pfs_cipher "
                     fi
                fi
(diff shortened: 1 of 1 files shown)


drwetter:
Oh, yes! Thank you @dcooper16 :-)

PR #364, opened 2016-05-27
removed brackets from plain IPv6 addresses.

typingArtist:
• the supplied openssl.Linux.x86_64 does not support plain IPv6 addresses at all.
• Gentoo’s openssl-1.0.2h does support IPv6 addresses, but without the brackets only.
• lookups using dig, host, etc. don’t need the brackets either.
So, it’s better to use the IPv6 address without brackets.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7101,7 +7101,6 @@ nodeip_to_proper_ip6() {
      local len_nodeip=0
 
      if is_ipv6addr $NODEIP; then
-          NODEIP="[$NODEIP]"
           len_nodeip=${#NODEIP}
           CORRECT_SPACES="$(draw_line " " "$((len_nodeip - 16))" )"
           # IPv6 addresses are longer, this varaible takes care that "further IP" and "Service" is properly aligned


drwetter:
Thx. 

testssl.sh must have brackets. A) because it confuses the last digits otherwise as a port
B) because the supplied binaries won't work. I am not sure why Gentoo is handling this differently as Fedora (where the IPv6 patch here comes from)

```
eagle:~/testssl/git.testssl.sh/bin|130% ./openssl.Linux.x86_64 s_client -connect '2a00:1450:4005:803::200e:443'                                     15:12:31
getaddrinfo: Servname not supported for ai_socktype
connect:errno=0
eagle:~/testssl/git.testssl.sh/bin|1% . ~/.zshrc.demo                                                                                               15:12:36
dirks@laptop:~/testssl/git.testssl.sh/bin|0% ./openssl.Linux.x86_64 s_client -connect '2a00:1450:4005:803::200e:443'
getaddrinfo: Servname not supported for ai_socktype
connect:errno=0
dirks@laptop:~/testssl/git.testssl.sh/bin|1% ./openssl.Linux.x86_64 s_client -connect '[2a00:1450:4005:803::200e]:443' </dev/null | egrep 'Protocol|Cipher'
depth=3 C = US, O = Equifax, OU = Equifax Secure Certificate Authority
verify return:1
depth=2 C = US, O = GeoTrust Inc., CN = GeoTrust Global CA
verify return:1
depth=1 C = US, O = Google Inc, CN = Google Internet Authority G2
verify return:1
depth=0 C = US, ST = California, L = Mountain View, O = Google Inc, CN = *.google.com
verify return:1
DONE
New, TLSv1/SSLv3, Cipher is ECDHE-RSA-CHACHA20-POLY1305
    Protocol  : TLSv1.2
    Cipher    : ECDHE-RSA-CHACHA20-POLY1305
dirks@laptop:~/testssl/git.testssl.sh/bin|0%
```

typingArtist:
Hi Dirk,

I understand. Just double-checked with a CentOS 7 box which is behaving as you describe for Fedora.

However, none of the other tools (dig, host, nslookup) support the brackets either, so I expect those to fail on Fedora as well. We should definitely find a solution that works for all the used tools, shouldn’t we?

I’ll dig into the Gentoo issue of its special, incompatible IPv6 patch. Perhaps this patch can be made compatible with bracketed addresses.

drwetter:
Hello @typingArtist,

believe me it all works perfect as is with IPv6. :-) 

Square brackets are common also for URis (browsers, mail, etc.), that's why after resolving I also don't do anything different and use square brackets.

Don't know WTH Gentoo is doing and I am not really tempted to use that, sorry.

If you or others have some cycles to spend though: what's missing to get IPv6 completed is an IPv6 proxy. The FC patch doesn't have that. IIRC LibreSSL has that, but maybe Gentoo is easier to adapt as it's OpenSSL ..

Cheers! Dirk

typingArtist:
Hi Dirk,

thanks for your feedback. I’m quite aware of IPv6 and it being represented with brackets in browsers/URLs. However, I’d like that to fixed, in one way or the other, and I want to help solve that, so let’s discuss things further in https://github.com/drwetter/testssl.sh/issues/365.

drwetter:
Hi @typingArtist,

from https://tools.ietf.org/html/rfc3986
```
A host identified by an Internet Protocol literal address, version 6
   [RFC3513] or later, is distinguished by enclosing the IP literal
   within square brackets ("[" and "]").  This is the only place where
   square bracket characters are allowed in the URI syntax.  In
   anticipation of future, as-yet-undefined IP literal address formats,
   an implementation may use an optional version flag to indicate such a
   format explicitly rather than rely on heuristic determination.

      IP-literal = "[" ( IPv6address / IPvFuture  ) "]"
```

I'll look later into #365  . 

Cheers! Dirk

Issue #276, opened 2016-01-31
Multiple certificate check does not show SHA-1 certificate


prdonahue:
Here's how to get our SHA-1:

    $ openssl s_client -connect www.cloudflare.com:443 -tls1_1 </dev/null 2>/dev/null | openssl x509 -noout -text | grep "Signature Algorithm"
    Signature Algorithm: sha1WithRSAEncryption
    Signature Algorithm: sha1WithRSAEncryption

And here's what the script shows:

    Server Certifcate #1
    Server key size              2048 bit
    Signature Algorithm          SHA256 with RSA
    Fingerprint / Serial         SHA1 A27166492178CC73EDF2D171D87D5BA1484DFFF8 / E3BB289893780DEAB01913B0A1400D77
                                SHA256 996F1E3DB7354AE92CC98CE12E473A8D01FFFD3E52F39AC514FEAFA6156AA499
    Common Name (CN)             (no CN field in subject) (CN in response to request w/o SNI: "no CN field in subject")
    subjectAltName (SAN)         "cloudflare.com" "www.cloudflare.com" 
    Issuer                       "COMODO RSA Extended Validation Secure Server CA" ("COMODO CA Limited" from "GB")
    EV cert (experimental)       yes 
    Certificate Expiration       304 >= 60 days (2015-11-30 16:00 --> 2016-11-30 15:59 -0800)
    # of certificates provided   3
    Chain of trust (experim.)    "/etc/*.pem" cannot be found / not readable
    Certificate Revocation List  http://crl.comodoca.com/COMODORSAExtendedValidationSecureServerCA.crl
    OCSP URI                     http://ocsp.comodoca.com
    OCSP stapling                offered

    Server Certifcate #2
    Server key size              EC 256 bit
    Signature Algorithm          ECDSA with SHA256
    Fingerprint / Serial         SHA1 F6F6A4ABEDE20AFC147BEAFD1FAFC9304902EA49 / 066CB625C622BE29CA69425208A1AD28
                                SHA256 E5554E6A21828B6F17546C2B77227F37003C7A092693BEB394BEEFF80A735880
    Common Name (CN)             (no CN field in subject) (CN in response to request w/o SNI: "no CN field in subject")
    subjectAltName (SAN)         "cloudflare.com" "www.cloudflare.com" 
    Issuer                       "COMODO ECC Extended Validation Secure Server CA" ("COMODO CA Limited" from "GB")
    EV cert (experimental)       yes 
[...]

Issue #94, opened 2015-05-05
CN <--> hostname match


drwetter:

drwetter:
A PoC is in https://github.com/drwetter/testssl.sh/blob/master/utils/checkcert.sh but fails sometimes and needs more work

dcooper16:
Hi @drwetter ,

I'm trying to do something for #276. Doing the additional test (TLSv1.1 w/o SNI) is easy, but the check of whether the same host is being hit is a bit more difficult. I want to say that if the hosts name ($NODE) appears in the certificate's CN or SAN, then then new certificate should be displayed, even if the CN and/or SAN aren't the same as in certificates that were previously seen. So, I looked at the existing code in certificate_info(), but I can't understand what it is trying to do:

```
if echo -n "$cn" | grep -q '^*.' ; then
     out " (wildcard certificate"
     cnfinding+="(wildcard certificate "
     if [[ "$cn" == "*.$(echo -n "$cn" | sed 's/^\*.//')" ]]; then
          out " match)"
          cnfinding+=" match)"
          wildcard=true
     else
          cnfinding+=" NO match)"
          cnok="INFO"
          #FIXME: we need to test also the SANs as they can contain a wild card (google.de .e.g) ==> 2.7dev
    fi
fi
```
What is the purpose of "if [[ "$cn" == "*.$(echo -n "$cn" | sed 's/^\*.//')" ]]; then"? The result of this is to either display "wildcard certificate match" or "wildcard certificate NO match," but the $cn is just being compared to a modified version of itself, so what does "match" or "NO match" mean? Was the first $cn supposed to be $NODE?

Here are the results I get for tools.ietf.org:

```
 TLS extensions (standard)    "renegotiation info/#65281" "EC point formats/#11" "session ticket/#35" "heartbeat/#15"
 Session Tickets RFC 5077     300 seconds (PFS requires session ticket keys to be rotated <= daily)
 SSL Session ID support       yes
 TLS clock skew               random values, no fingerprinting possible 
 Signature Algorithm          SHA256 with RSA
 Server key size              2048 bits
 Fingerprint / Serial         SHA1 B1114E17451F8626DA8C506DEA5A7EFFC9110EA8 / 9983BFB1CE0BE894
                              SHA256 4832567962EC34C2F22B8892CAB8F87065A9375B4A7F7B939D4E3D1B15467DD1
 Common Name (CN)             "*.tools.ietf.org" (wildcard certificate match) (works w/o SNI)
[...]

PR #371, opened 2016-06-01
Fix issue #276

dcooper16:
Here is my proposed change to fix issue #276.

run_server_defaults() currently calls determine_tls_extensions() seven times, once for each public-key type (RSA encryption, RSA signature, DSA, DH, ECDH, ECDSA, and GOST). For each call to determine_tls_extensions() that is successful, the PR calls determine_tls_extensions() a second time for that public-key type, but this time with SNI="" and telling determine_tls_extensions() to only try TLSv1.1.

If the second call to determine_tls_extensions() is successful and the certificate returned by the server has not been encountered before, then the new certificate is displayed if either:

- The server name ($NODE) matches the CN or one of the SANs in the certificate; or
- The CN in the new certificate is the same as the CN in the first certificate returned by the server **and** at least one of the SANs in the new certificate also appears in the first certificate returned by the server.

The second option is most likely only needed for the case in which the URL provided to testssl.sh is an IP address rather than a DNS name.


dcooper16:
In response to the comment in #94, I changed the code in `compare_server_name_to_cert()` to compare the `$servername` against both the DNS names and the IP addresses in the subjectAltName extension of `$cert`. I had previously just copied the code from `certificate_info()` for extracting the subjectAltName:
```
     sans=$($OPENSSL x509 -in $HOSTCERT -noout -text 2>>$ERRFILE | grep -A3 "Subject Alternative Name" | grep "DNS:" | \
          sed -e 's/DNS://g' -e 's/ //g' -e 's/,/ /g' -e 's/othername:<unsupported>//g')
```
but I guess that's not what's needed here.

This code is still not complete, however. I believe it should work if `$servername` is a DNS name or an IPv4 address, but I don't think it will work if `$servername` is an IPv6 address. I just compare `$servername` to the IP address as it is displayed by OpenSSL, but I have no reason to believe that OpenSSL and testssl.sh format IPv6 addresses in the same way.

drwetter:
Thx, @dcooper16 !

Three certificates on cloudflare.com (anticipated) -- four certificates on FB.

On simple hosts like testssl.sh or dev.testssl.sh there's a slight performance penalty. I guess there's not so much one can do right?

There are two minor points which need to be addressed before release. I'll file an issue for this.

Thx again, Dirk

PR #373, opened 2016-06-07
openssl2rfc and rfc2openssl

dcooper16:
This PR provides implementations of openssl2rfc and rfc2openssl. It also uses openssl2rfc() in run_server_preference() to help determine how to display the "negotiated cipher." I believe that using the RFC names addresses the current FIXME:

`     FIXME BEAST: We miss some CBC ciphers here, need to work w/ a list"`

I don't understand why ciphers with ECDHE and AES were marked as pr_svrty_minor, whereas all other CBC ciphers (that were caught) were marked as pr_svrty_medium, but I left that unchanged.


drwetter:
Hi @dcooper16,

* idea is good, thx!
* ``$OPENSSL ciphers -V "$1" 2>>$ERRFILE | head -1 | awk '{ print $1 }')``  doesn't look like it's picking the one it is supposed to do. Why is there head -1??
* If "$MAPPING_FILE_RFC" cannot be found -- which is something which is still ok for now -- L1455 throws an error
* FYI: My plan was to work in the future with a fixed table and to extend / replace ``mapping-rfc.txt`` as it is inefficent to look s.th. repeatedly with the openssl binary. Maybe arrays are even more efficient
* your PR unveils an error in the existing code: CBC ciphers shouldn't be a medium finding if TLS 1.2 or 1.1 is available

FYI: While my commit yesterday I stumbled 1) numerous times over ``cipher=$(grep -aw "Cipher" $TMPFILE | egrep -avw "New|is" | sed -e 's/^.*Cipher.*://' -e 's/ //g')`` which today seems to me waayyy too difficult (1x sed, 1x grep, 1x egrep). I'am about to change that into ``awk '/Cipher.*:/ { print $3 }'``. 


1) that one goes on me ;-)

dcooper16:
Hi @drwetter,

I tested openssl2rfc() on every cipher supported by OpenSSL andrfc2openssl on every cipher in etc/mapping-rfc.txt, so I am fairly certain that it function correctly (at least $MAPPING_FILE_RFC can be found). If there is a better implementation, however, that's fine. I just needed openssl2rfc() for some code I was writing, so thought I'd go ahead and implement both of them and then submit the results.

In openssl2rfc(), the "head -1" is there since the call to OpenSSL doesn't always return one line. If I use @PeterMosmans OpenSSL and and the cipher is RC4-MD5, the output is
```
          0x00,0x04 - RC4-MD5                 SSLv3 Kx=RSA      Au=RSA  Enc=RC4(128)  Mac=MD5 
     0x01,0x00,0x80 - RC4-MD5                 SSLv2 Kx=RSA      Au=RSA  Enc=RC4(128)  Mac=MD5
```
So, I use "head -1" so that the "awk" grabs the 2-byte hexcode.

I use "head -1" for a similar reason in rfc2openssl(), since the grep of $MAPPING_FILE_RFC will return 2 lines for SSL_RSA_FIPS_WITH_DES_CBC_SHA or SSL_RSA_FIPS_WITH_3DES_EDE_CBC_SHA, and I want the hexcode from the first matching line (xFEFE or xFEFF).
```
xFEFE  SSL_RSA_FIPS_WITH_DES_CBC_SHA
xFEFF  SSL_RSA_FIPS_WITH_3DES_EDE_CBC_SHA
xFFE0  SSL_RSA_FIPS_WITH_3DES_EDE_CBC_SHA
xFFE1  SSL_RSA_FIPS_WITH_DES_CBC_SHA
```
I fixed the code so that the code doesn't throw an error if $MAPPING_FILE_RFC is missing.

I like the idea of extending mapping-rfc.txt. In https://github.com/dcooper16/testssl.sh/blob/extended_tls_sockets/testssl.sh, I have versions of run_allciphers() and run_cipher_per_proto() that use sslv2_sockets() and tls_sockets() rather than "$OPENSSL s_client." However, at the moment they rely on "$OPENSSL ciphers -V" to get a list of cipher suite hexcodes to text. I have a local version that doesn't rely on $OPENSSL, but it needs a file similar to  mapping-rfc.txt, but where the ciphers are ordered by strength, and where "Kx" and "Enc" (as provided by "$OPENSSL ciphers -V" are also included.

drwetter:
Hi @dcooper16,

thx!

On the long run I thought about using s.th. like this: https://github.com/drwetter/testssl.sh/blob/master/etc/mapping.txt . 

Atm however I am personally spending my spare time for getting the bugs squashed so that 2.8 can be released eventually.

Cheers, Dirk

dcooper16:
This PR can certainly wait under after 2.8 is released.

However, I decided to try the idea of using an array. I load the array by first creating a temporary file and then loading the contents of the file into an array. The array seems to be a bit faster, but both are so fast that I not convinced the speed difference matters. I would think that in most cases, the OS will keep a copy of the mapping file in RAM (disk cache), so uses of the file wouldn't involve disk operations.

The mapping file I created is similar to https://github.com/drwetter/testssl.sh/blob/master/etc/mapping.txt, but I copied the formatting of "$OPENSSL ciphers -V". I also sorted the ciphers by encryption strength, so that the file could easily be used in all of the functions (e.g., run_allciphers() and run_cipher_per_proto()) that needs the ciphers to be ordered in that way. In addition, it includes all of the ciphers from mapping-rfc.txt, even if I couldn't find an OpenSSL name for it, since such ciphers can still be tested using tls_sockets().

drwetter:
> This PR can certainly wait under after 2.8 is released.

yes.

> However, I decided to try the idea of using an array. I load the array by first creating a temporary > file and then loading the contents of the file into an array. The array seems to be a bit faster, 
> but both are so fast that I not convinced the speed difference matters. I would think that in 
> most cases, the OS will keep a copy of the mapping file in RAM (disk cache), so uses of the file > wouldn't involve disk operations.
>
> The mapping file I created is similar to https://github.com/drwetter/testssl.sh/blob/master
> /etc/mapping.txt, but I copied the formatting of "$OPENSSL ciphers -V". \

"Raw data" (i.e. from openssl) could be better, if one doesn't do the same operations again and again on it. My mapping file above was generated with ``testssl.sh -V`` and several openssl versions I have (1.1.0, 1.0.2i-pm and others).

Actually as I released 2.6 I also thought to replace all ``while-do-done << $($OPENSSL ciphers -V)``.

> In addition, it includes all of the ciphers from mapping-rfc.txt, even if I couldn't find an OpenSSL > name for it, since such ciphers can still be tested using tls_sockets().

Did you have a look @ https://testssl.sh/openssl-rfc.mappping.html / https://github.com/drwetter/testssl.sh/blob/master/openssl-rfc.mappping.html ?

Cheers, Dirk

PeterMosmans:
Hi @dcooper16 @drwetter - See my comment in #379 wrt a slight renaming proposal

drwetter:
Thx, David. Merged into 2.9dev via https://github.com/drwetter/testssl.sh/commit/aab0487a9662c574c8b114a20522d3b670fb5a8f.

Sill not convinced for the current master ;-)

PR #377, opened 2016-06-08
Adding x25519 and x448 to ClientHello

dcooper16:
Hi @drwetter,

This PR adds curves x25519 and x448 to the ClientHello created by socksend_tls_clienthello(), as you requested in #342.

I added both EC 1d and EC 1e since both are listed in the IANA registry: http://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8.


drwetter:
great, thank you!

Issue #378, opened 2016-06-09
New TLS 1.2 overflow (regression)


drwetter:
While #189 was fixed I figured that ``-E`` in those case with TLS 1.2 doesn't show a single cipher :-( 

@dcooper16 : would you be able to look into it if @feld allows me to pass the hostname to you? I suppose ``OPENSSL s_client -cipher "${ciphers_to_test:1}" `` in ``run_cipher_per_proto()`` is the culprit.

Current status in this particular case is
```
 Testing all locally available ciphers per protocol against the server, ordered by encryption strength 

Hexcode  Cipher Suite Name (OpenSSL)    KeyExch.   Encryption Bits        Cipher Suite Name (RFC)
--------------------------------------------------------------------------------------------------------------------------
SSLv2 
SSLv3 
TLS 1 
 x35     AES256-SHA                     RSA        AES        256         TLS_RSA_WITH_AES_256_CBC_SHA                      
 x2f     AES128-SHA                     RSA        AES        128         TLS_RSA_WITH_AES_128_CBC_SHA                      
 x0a     DES-CBC3-SHA                   RSA        3DES       168         TLS_RSA_WITH_3DES_EDE_CBC_SHA                     
TLS 1.1 
 x35     AES256-SHA                     RSA        AES        256         TLS_RSA_WITH_AES_256_CBC_SHA                      
 x2f     AES128-SHA                     RSA        AES        128         TLS_RSA_WITH_AES_128_CBC_SHA                      
 x0a     DES-CBC3-SHA                   RSA        3DES       168         TLS_RSA_WITH_3DES_EDE_CBC_SHA                     
TLS 1.2 

```

Whereas before it was 
```
Testing all locally available ciphers per protocol against the server, ordered by encryption strength 

Hexcode  Cipher Suite Name (OpenSSL)    KeyExch.   Encryption Bits
-------------------------------------------------------------------------
SSLv2 
SSLv3 
TLS 1 
 x35     AES256-SHA                     RSA        AES        256                                                                                     
 x2f     AES128-SHA                     RSA        AES        128                                                                                     
 x0a     DES-CBC3-SHA                   RSA        3DES       168                                                                                     
TLS 1.1 
 x35     AES256-SHA                     RSA        AES        256                                                                                     
 x2f     AES128-SHA                     RSA        AES        128                                                                                     
 x0a     DES-CBC3-SHA                   RSA        3DES       168                                                                                     
TLS 1.2 
[...]

PR #380, opened 2016-06-10
run_cipher_per_proto() 128-cipher limit

dcooper16:
Here is the PR to address #378.

I went ahead and added in a call to `has_server_protocol()`, and I fixed a couple of minor bugs related to that function:
- the `grep` in `has_server_protocol()` was missing the `-q` option.
- run_protocols() was adding tls1 to $PROTOS_OFFERED if the server downgraded to SSLv3.


drwetter:
Thx, and thx also for muting the grep.

Issue #379, opened 2016-06-10
Reconsider naming scheme for ciphers 0xCC13-0xCC15


drwetter:
There are new ciphers in 1.1.0 which carry the same OpenSSL name (0xCCAA-0xCCAE). 

@PeterMosmans : any suggestion?

See https://github.com/drwetter/testssl.sh/blob/master/etc/mapping.txt (I already renamed to them there *OLD but I don't know whether this is common or there's a common naming)

PeterMosmans:
Hmmm, that's a good one. The implementation differs, as the 1.0.2-chacha has the "old" Adam Langley implementation. Officially they can be considered equal ciphers. Let me see if I can come up with some relevant documentation, so that we're doing the right, future-proof thing here.

drwetter:
Ok, now that we have the RFC/IANA names changed  would like to propose a change for your branch open openssl. Those names still have a conflict (1st current 1.1.0 version)

```
dirk@laptop:/data/tmp/openssl.git|1%  LD_LIBRARY_PATH=$PWD apps/openssl ciphers -V | grep CHACHA       
          0xCC,0xA9 - ECDHE-ECDSA-CHACHA20-POLY1305 TLSv1.2 Kx=ECDH     Au=ECDSA Enc=CHACHA20/POLY1305(256) Mac=AEAD
          0xCC,0xA8 - ECDHE-RSA-CHACHA20-POLY1305 TLSv1.2 Kx=ECDH     Au=RSA  Enc=CHACHA20/POLY1305(256) Mac=AEAD
          0xCC,0xAA - DHE-RSA-CHACHA20-POLY1305 TLSv1.2 Kx=DH       Au=RSA  Enc=CHACHA20/POLY1305(256) Mac=AEAD
          0xCC,0xAE - RSA-PSK-CHACHA20-POLY1305 TLSv1.2 Kx=RSAPSK   Au=RSA  Enc=CHACHA20/POLY1305(256) Mac=AEAD
          0xCC,0xAD - DHE-PSK-CHACHA20-POLY1305 TLSv1.2 Kx=DHEPSK   Au=PSK  Enc=CHACHA20/POLY1305(256) Mac=AEAD
          0xCC,0xAC - ECDHE-PSK-CHACHA20-POLY1305 TLSv1.2 Kx=ECDHEPSK Au=PSK  Enc=CHACHA20/POLY1305(256) Mac=AEAD
[...]

PR #381, opened 2016-06-13
Updated ChaCha20 cipher names

PeterMosmans:
See https://tools.ietf.org/html/draft-ietf-tls-chacha20-poly1305-04 (the latest version as of this writing is 04).
The previous version received the suffix _2013. See https://tools.ietf.org/html/draft-agl-tls-chacha20poly1305-04

See #379 for the background information/request.


drwetter:
Thx!

The lines in the output are getting longer and longer ;-( Soon everybody will be required to purchase a 21:9 monitor. ;-)

The "2013": That was you, right, and not a suggestion by/for IETF?

PeterMosmans:
Yeah, that was me :cry: / :smiley_cat:

PR #382, opened 2016-06-13
Allow the file output feature and mass_test feature to work together

seccubus:
Had to kill the run_mass_scan_parallell routine due to concurrency problems.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -455,7 +455,11 @@ strip_quote() {
 
 fileout_header() {
      if [[ $APPEND ]]; then
-          "$do_json" && [[ ! -f "$JSONFILE" ]] && printf "[\n" > "$JSONFILE"
+          if [[ -f "$JSONFILE" ]]; then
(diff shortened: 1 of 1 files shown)


drwetter:
@seccubus :  run_mass_testing_parallel() is currently a dead function

seccubus:
The error was in the other function. But cannot be fixed in the parallel version.

drwetter:
Hmm, ok.

But unless you specify one file on the command line there's no chance a file being overwritten. Is this (user error) that what you try to address?

seccubus:
I think its is not a user error that one would expect all assessments to end up in a single file. Otherwise for a db inport one would have to iterate over several files.

drwetter:
Ok, was checking your perspective.

Idea is fine. I need to do some tweaking on it. Bear with me.

Thx, Dirk

seccubus:
We also need this for Seccubus integration.

drwetter:
Thx @seccubus !

Issue #304, opened 2016-02-26
More unit testing: openssl s_server


drwetter:
Full coverage will be difficult to achieve. We need more unit testing though.

The first step IMO should be to run some tests, e.g. as `make report` does in openssl. We can use `openssl s_server` on one side and `openssl s_client -connect` on the other side.

Help would be much appreciated.

drwetter:
Idea, see e.g. https://github.com/PeterMosmans/openssl/blob/1.0.2-chacha/test/testssl

drwetter:
see also https://github.com/drwetter/testssl.sh/blob/2.9dev/utils/00_unittest_baseline.sh

drwetter:
This issue is a duplicate. closing in favor for #756

PR #392, opened 2016-06-28
Lets add some unit tests to testssl.sh - Using abdsll.com work

seccubus:
I would like to see some unit tests for tetstssl.sh so we can check the quality. 
badsll.com provides a good testbed for this.

Ideally an integration between GitHub and Travis CI is set up: https://docs.travis-ci.com/user/languages/perl

@drwetter can you let me know if you like this idea and if you need help setting up Travis CI?

I plan to add more tests to this later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store
--- a/t/01_badssl.com.t
+++ b/t/01_badssl.com.t
@@ -1,0 +1,117 @@
+#!/usr/bin/env perl
+
+use strict;
+use Test::More;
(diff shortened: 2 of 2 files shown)


drwetter:
Hi Frank,

Sure I like the idea -- there's also an issue, see #304. 

Are you by any chance in Rome ? maybe you can help setting it up, never used travis-ci .If it's paiinless for me let's do it before 2.8 as it helps maturing testssl.sh. If not I would rather suggest to wait.

Cheer, Dirk

PR #398, opened 2016-07-03
This Fixes #258 - Checks if /dev/fd is mounted on FreeBSD

### LINKED ISSUE #258: Prints /dev/fd/62: No such file or directory
On FreeBSD I am getting these "No such file or directory":

```
 LOGJAM (CVE-2015-4000), experimental      VULNERABLE (NOT ok), uses DHE EXPORT ciphers, common primes not checked.
 BEAST (CVE-2011-3389)                     ./testssl.sh: line 3732: /dev/fd/62: No such file or directory
./testssl.sh: line 3732: /dev/fd/62: No such file or directory
./testssl.sh: line 3732: /dev/fd/62: No such file or directory
```

This line is:
```
read hexcode dash cbc_cipher sslvers kx auth enc mac < <($OPENSSL ciphers -V "$ciph" 2>>$ERRFILE)        # -V doesn't work with openssl < 1.0
```

bash-4.3.42

seccubus:


drwetter:
thx, @seccubus .

Better then ``ls /dev/fd`` would be using echo. However logically like ``mount | grep '/dev/fd' | grep -q fdescfs`` is the right test.

You want me to do this or do you update your PR?

seccubus:
I'll fix it.

drwetter:
Hi @seccubus,

thx. Your branch was not up to date yet, so I rather chose to copy a few lines after testing.

Cheers, Dirk

seccubus:
No worries. 
More reasons to look at splitting up into smaller files I guess.

welwood08:
Hi, I just noticed the chosen solution fails inside a FreeBSD jail.

Previously I did encounter #258 and successfully mounted fdescfs to the appropriate path from the host system (since it's not permitted to do so within the jail). The workaround was successful, but now after updating to head this change simply throws a fatal error even though fdescfs is correctly mounted at /dev/fd inside the jail.

I'm still able to use the script from the host system (although I had not tried before), so perhaps it's not worth fixing for jails if the check would need to be less robust.

drwetter:
feel free to make a suggestion or provide a PR. Or let me know what command exactly fails...

seccubus:
Any way to detect if you are inside a jail?

drwetter:
guess the mount command fails...

welwood08:
Yes, the mount command only returns the root mount on my system even when I have mounted fdescfs into the jail from the host. I was researching to see if there was a better way to detect presence of fdescfs, but no luck.

You can detect being inside a jail with `sysctl -n security.jail.jailed`, returns 1 in jail or 0 in host.

drwetter:
``mount`` is for my taste a command I rather wanted to avoid. sysctl tops that :-)

Any way to distinguish between the output of ``mount`` jailed or not? Is devfs visible in jail?

welwood08:
```
will@gem ~> mount
ocean/jail/gem on / (zfs, local, nfsv4acls)
will@gem ~> mount -p
ocean/jail/gem          /                       zfs     rw,nfsv4acls    0 0
```
Although the word `jail` does appear, that is merely part of the path I chose on the host system.

Are your objections to `sysctl` based on the Linux implementation?
"A sysctl utility first appeared in	4.4BSD."
"In BSD, the system call is implemented directly in the kernel, as described in the sysctl(3) manual page. In Linux, the sysctl is implemented as a wrapper around file system routines that access contents of files in the /proc directory. The result is that it is much more expensive to use the sysctl interface in Linux."

drwetter:
> Are your objections to sysctl based on the Linux implementation?

It's too much low level IMO. Other people were requesting ``rpm`` or similar. Also if there's a check before whether it exists it blows up the code similar to c stylish #ifdef with long system stuff in between. So I rather would like to keep that to a very minimum.

Any reason to not to check for the existence of devfs and then decide whether testssl is jailed?

On my ufs system it looks like 
```
/dev/ada0p2 on / (ufs, local, journaled soft-updates)
devfs on /dev (devfs, local, multilabel)
/dev/ada0p4 on /usr/home (ufs, local)
fdescfs on /dev/fd (fdescfs)
```

welwood08:
Again, devfs is mounted by the host and not visible in the jail's `mount`. With devfs, `/dev/fd` is created with only files for 0-2, presumably this is why this PR has ` -le 3` involved in the fdescfs check.

To thwart another idea I had, the jail's `/etc/fstab` is completely empty.

How about directly testing if bash emits a `/dev/fd/...: No such file or directory` error after checking the system is BSD?

drwetter:
Ok, could try that. I was more referring to this line of output in ``mount``

```
devfs on /dev (devfs, local, multilabel)
```

That seems standard on my FreeBSD 10+9 without using the fdescfs hook. 

Can you confirm you don't see that line in your jail?

welwood08:
The output I provided above showing a single line of output from each mount command is not truncated or censored. Having done a bit more reading since then, I believe it is due to the jail parameter `enforce_statfs` having a default value of 2 - see the manual page for `jail(8)`.

drwetter:
0c22ea9a0e92156d160ef7a9a02a366a79f19196 should have fixed the jail detection. Pls let me know if not

welwood08:
Confirmed I can now run the script from within a jail again, thanks.

If I may suggest improvements to the message:
```
You seem to be running testssl.sh in a jail. If you see errors like
"/dev/fd/62: No such file or directory", bash needs fdescfs mounted
at the jail's /dev/fd - there are several ways to do this, such as
setting the "mount.fdescfs" jail parameter as described in jail(8).
```

To maybe throw another spanner in the works: as I understand it, the fdescfs requirement of FreeBSD's bash is a side-effect of the build systems having it enabled and bash doing build-time feature detection. Consequently, people who have built bash from source (perhaps in a jail) without fdescfs mounted won't even need fdescfs at /dev/fd - I don't know what bash uses instead under these conditions but it could be another reason the check should instead simply see if bash throws the error in the first place rather than heuristics on the output of `mount`.

Issue #399, opened 2016-07-04
POODLE test not detecting all CBC ciphers properly


aelth:
First of all, thank you for the great tool Dirk!

I found a potential issue with POODLE test today - testssl reported that the site was not vulnerable to Poodle, but SSLv3 is enabled and the negotiated cipher is indeed using CBC mode:

```
 Testing protocols (via sockets except TLS 1.2, SPDY+HTTP2) 

 SSLv2      not offered (OK)
 SSLv3      offered (NOT ok)
...
Testing server preferences 

 Has server cipher order?     yes (OK)
 Negotiated protocol          TLSv1
 Negotiated cipher            AES128-SHA
 Cipher order
     SSLv3:     AES128-SHA 
...
 POODLE, SSL (CVE-2014-3566)               not vulnerable (OK)
 TLS_FALLBACK_SCSV (RFC 7507), experim.    Downgrade attack prevention NOT supported
...
Hexcode  Cipher Suite Name (OpenSSL)    KeyExch.   Encryption Bits        Cipher Suite Name (RFC)
--------------------------------------------------------------------------------------------------------------------------
 x2f     AES128-SHA                     RSA        AES        128         TLS_RSA_WITH_AES_128_CBC_SHA   
```

If you look at the openssl->RFC mapping, AES128-SHA is using CBC mode. I believe the problematic line is https://github.com/drwetter/testssl.sh/blob/master/testssl.sh#L5986 where openssl is trusted to contain 'CBC' substring in all ciphers actually using CBC, which unfortunately openssl fails to do:(

drwetter:
I am assuming it works now, thus I am closing this.

Tested it on a similar system

PR #402, opened 2016-07-06
Check for all CBC ciphers in Poodle test

dcooper16:
This PR should address issue #399.

I created the list of ciphers using the `CIPHERS_BY_STRENGTH` file from PR #373, making a list of all ciphers that had "CBC" in the RFC name and for which I had been able to find a corresponding OpenSSL name. Then, since that list contained more than 128 ciphers, I removed any ciphers from the list where the name ended in "-SHA256" or "-SHA384", as it is my understanding that those ciphers can only be used with TLS 1.2.


drwetter:
Thx @dcooper16! Also about removing the TLS 1.2 ciphers.

openssl 1.1.x didn't show more ciphers (!= TLS1.2). Tested it on ~10 systems and the results are what I expected.

dcooper16:
The OpenSSL ciphers command only seems to distinguish between SSLv2 and non-SSLv2. However, I looked at the results of the "-E" option on several servers, and in each case, the cipher suites that ended in "-SHA256" or "-SHA384" only appeared under TLSv1.2, whereas all of the other cipher suites appeared under all supported protocol versions.

I also just found this sentence in SP 800-52 Rev. 1:

> Cipher suites that specify secure hash algorithms other than SHA-1 are not supported prior to TLS 1.2.

There shouldn't be any harm in including the "-SHA256" and "-SHA-384" ciphers, even if the ClientHello specifies a protocol earlier than TLSv1.2. The only reason I bothered to remove them is that there are more than 128 CBC ciphers, there is at least one server that cannot handle ClientHello's with 128 or more ciphers, and removing the "-SHA256" and "-SHA-384" ciphers brought the list down to 101 ciphers.

drwetter:
Am 8. Juli 2016 15:02:37 MESZ, schrieb David Cooper <notifications@github.com>:
>
>I also just found this sentence in SP 800-52 Rev. 1:
>
>> Cipher suites that specify secure hash algorithms other than SHA-1
>are not supported prior to TLS 1.2.
>
>There shouldn't be any harm in including the "-SHA256" and "-SHA-384"
>ciphers, even if the ClientHello specifies a protocol earlier than
>TLSv1.2. The only reason I bothered to remove them is that there are
>more than 128 CBC ciphers, there is at least one server that cannot
>handle ClientHello's with 128 or more ciphers, and removing the
>"-SHA256" and "-SHA-384" ciphers brought the list down to 101 ciphers.

Good point!


Cheers, Dirk 


-- 
Set from my mobile. Excuse my brevity&typos

PR #411, opened 2016-07-11
Server cipher order NPN tests should use SNI

welwood08:
I noticed the NPN parts of the cipher order tests were not returning any ECDSA ciphers where I expected them to match the results of the immediately preceding TLS 1.2 test. Found it wasn't using SNI so my test server was using the default domain (snakeoil RSA certificate) instead of the tested domain (dual ECDSA/RSA certificates).

I have used other s_client calls as a template, so `-host`/`-port` are replaced with `-connect` (the only remaining use of `-host` elsewhere is commented out). `$PROXY` has also been removed from the inner 2 calls due to finding a comment asserting that proxies don't support NPN. If any of these were deliberately so, please correct me.


drwetter:
sigh, yes!

Thank you.

Issue #122, opened 2015-06-22
Complete client hello (socket)


drwetter:
all extensions are missing

drwetter:
For TLS 1.2 Windows SChannel seems to need this.

drwetter:
@dcooper16 : for TLS 1.2 we still have the EXPERIMENTAL flag. 

Q: technically is that still needed with your improvements?

Q: have you tested it? I remember with my old implementation IIS 7.5 server hiccuped.

dcooper16:
My guess is that the EXPERIMENTAL flag is no longer needed. I've tested it on quite a few servers, including an IIS/8.0 server that didn't work for TLS 1.2 with sockets before, but does now. However, I haven't tested it against an IIS/7.5 server that has been configured to support TLS 1.2.

There is one minor issue that I've discovered with `tls_sockets()`, but it's not an issue that will affect v2.8.

Currently, `socksend_tls_clienthello()` sets the supported elliptic curves extension as follows:
```
          00, 0a,                    # Type: Supported Elliptic Curves , see RFC 4492
          00, 3e, 00, 3c,            # lengths
          00, 01, 00, 02, 00, 03, 00, 04, 00, 05, 00, 06, 00, 07, 00, 08,
          00, 09, 00, 0a, 00, 0b, 00, 0c, 00, 0d, 00, 0e, 00, 0f, 00, 10,
          00, 11, 00, 12, 00, 13, 00, 14, 00, 15, 00, 16, 00, 17, 00, 18,
          00, 19, 00, 1a, 00, 1b, 00, 1c, 00, 1d, 00, 1e,
```
It just lists all of the named curves specified in Section 5.1.1 of RFC 4492 (plus a few new ones) in numerical order. The "problem" is that it turns out that the order has meaning:
```
   Items in elliptic_curve_list are ordered according to the client's
   preferences (favorite choice first).
```

The result is that if `tls_sockets()` is used to implement `run_allciphers()` (see https://github.com/dcooper16/testssl.sh/tree/extended_tls_sockets), then the ephemeral key size displayed for ECDHE ciphers may be different than what would be shown if OpenSSL were used. In the case of fpki-graph.fpki-lab.gov, using OpenSSL shows 256 bits for ECDHE ciphers, whereas the sockets version shows 163 bits for ECDHE ciphers (and SSL labs shows 571 bits).

[...]

PR #412, opened 2016-07-11
Reorder supported curves

dcooper16:
This PR reorders the list of supported elliptic curves sent by `socksend_tls_clienthello()` when ciphers that use ephemeral elliptic curves are offered. The reordering is based on the conversation in #122, where it was agreed to order the curves from strongest to weakest.


drwetter:
thx @dcooper16 !

PR #413, opened 2016-07-11
Determine support elliptic curves for ECDHE- ciphers

dcooper16:
As was discussed in #122, the PR displays the set of elliptic curves supported by the server for use with ECDHE-* ciphers. Since the ECDHE-* ciphers are PFS ciphers, I added the new code to run_pfs().

Not all versions of OpenSSL support the "-curves" option for "s_client". At the moment, if no curves are found, then the code just doesn't display anything.

It was my original intention to implement the check for supported curves in sockets as well, but I ran into a problem. I originally wrote the sockets version to try to connect to the server with only ECDHE-* ciphers specified and with the Supported Elliptic Curves extension only specifying one curve. If the connection was successful, then I assumed the specified curve was supported. This worked in most cases, however, it appears that some servers (e.g., www.healthcare.gov) ignore the Supported Elliptic Curves extension and so the connection will succeed even if the curve the server uses isn't included in the Supported Elliptic Curves extension.

The solution is to compare the curve used by the server to the one specified in the Supported Elliptic Curves extension. However, doing this involves parsing more of the server's response. In some cases, this part of the response may not appear in the first packet returned by the server, and so tls_sockets() would need to request more data. This is already done in https://github.com/dcooper16/testssl.sh/tree/extended_tls_sockets, and so I added a sockets version of the test for supported elliptic curves to that version of testssl.sh. However, I don't think we want to make such significant changes to tls_sockets() and parse_tls_serverhello() in v2.8, so I think a sockets version of the test for supported elliptic curves should wait until after v2.8 is done.

P.S. The reason that I don't use "$OPENSSL ecparam -list_curves" is that $OPENSSL supports a lot of different curves (more than 80), whereas only 30 can be used for the ephemeral EC key in the ECDHE-* ciphers. I could check that a curve is locally supported before trying to use it with "$OPENSSL s_client", but there doesn't seem to be any benefit in doing that as opposed to get getting a error response from "$OPENSSL s_client".

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4489,13 +4489,15 @@ run_server_defaults() {
 
 run_pfs() {
      local -i sclient_success
-     local pfs_offered=false
+     local pfs_offered=false ecdhe_offered=false
      local tmpfile
      local dhlen
-     local hexcode dash pfs_cipher sslvers kx auth enc mac
+     local hexcode dash pfs_cipher sslvers kx auth enc mac curve
      local pfs_cipher_list="$ROBUST_PFS_CIPHERS"
+     local ecdhe_cipher_list=""
+     local -a curves=("sect163k1" "sect163r1" "sect163r2" "sect193r1" "sect193r2" "sect233k1" "sect233r1" "sect239k1" "sect283k1" "sect283r1" "sect409k1" "sect409r1" "sect571k1" "sect571r1" "secp160k1" "secp160r1" "secp160r2" "secp192k1" "prime192v1" "secp224k1" "secp224r1" "secp256k1" "prime256v1" "secp384r1" "secp521r1" "brainpoolP256r1" "brainpoolP384r1" "brainpoolP512r1" "X25519" "X448")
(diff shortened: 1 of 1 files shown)


drwetter:
Beautiful, thank you!

drwetter:
your socket implementation detects curve25519 e.g. @ google.com :+1: 

But as you said lets leave this for the next dev version. One comment: While looking at the results and the future sockets are the right thing to do, I've noticed in other tests a speed impact. This is probably due to the polling in ``sockread_serverhello()``

```
   SOCK_REPLY_FILE=$(mktemp $TEMPDIR/ddreply.XXXXXX) || return 7
     dd bs=$1 of=$SOCK_REPLY_FILE count=1 <&5 2>/dev/null &
     wait_kill $! $maxsleep
```
At the time where testssl.sh used only for some tests sockets this was more or less negligible. The more we make use of sockets the more important it becomes to put some thoughts into it.  

Cheers, Dirk

PR #415, opened 2016-07-14
Speed up finding supported curves

dcooper16:
Rather than try each curve one at a time, this PR follows model in `cipher_pref_check()`.  First include all curves in ClientHello, then successively remove from the ClientHello those curves that have been offered by the server until the connection fails. This makes the number of calls to `$OPENSSL s_client` one more than the number of supported curves rather than the number of curves in NamedCurve supported by $OPENSSL.

Note, however, that OpenSSL defines MAX_CURVELIST as 28 and fails if the `-curves` option includes more than 28 curves. Since OpenSSL 1.1.0 offers 29 curves from NamedCurve, this PR breaks the list of supported curves in 2. At the cost of one additional calls to `$OPENSSL s_client` it ensures that the number of curves provides to the `-curves` option is below the limit.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4495,9 +4495,12 @@ run_pfs() {
      local hexcode dash pfs_cipher sslvers kx auth enc mac curve
      local pfs_cipher_list="$ROBUST_PFS_CIPHERS"
      local ecdhe_cipher_list=""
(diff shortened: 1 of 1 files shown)


drwetter:
Thx, David!

PR #417, opened 2016-07-20
Fix JSON output of Issuer name

dcooper16:
`certificate_info()` currently outputs `$issuer` to the JSON file, where is should be outputting `$issuer_CN` in order for the information in the JSON file to match the information that is displayed.

This PR also fixes the problem that if an Issuer name contains a domain component attribute (DC=) then it will be mistakenly treated as a country attribute (C=).


drwetter:
thx for catching this!

PR #418, opened 2016-07-20
CA names with domain component attributes

dcooper16:
`certificate_info()` does not correctly display the Issuer name for CAs that use domain component attributes.

There is a server on the NIST intra-net that I test against that has a certificate issued by a NIST CA, and the issuer name in the certificate is of the form: `/DC=net/DC=example/DC=internal/CN=CAname`

Since there is no organizational name, testssl.sh displays the name as:
```
 Issuer                       "CAname" ("")
```
In this PR, if the Issuer name has 'DC=' attributes, but does not have an 'O=' attribute, the "DC=" attributes are combined into a DNS name that is used as if it were the organizational name:
```
 Issuer                       "CAname" ("internal.example.net")
```
I should note, however, that I have not been able to find any other examples of TLS server certificates that have been issued by CAs that have domain components ("DC=") in their names. So, it may not be worthwhile to change the code to try to accommodate such CAs.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3839,7 +3839,7 @@ certificate_info() {
      local ocsp_response=$5
      local ocsp_response_status=$6
      local cert_sig_algo cert_sig_hash_algo cert_key_algo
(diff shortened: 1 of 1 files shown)


drwetter:
Rare species but why not :-)  Thx.

PR #419, opened 2016-07-22
CN <--> hostname match

dcooper16:
This PR is intended to address issue #94 (CN <--> hostname match).


drwetter:
very good work, thank you!

Tested it on my side with a couple of hosts and w/o probs. (IP based SAN works too)!

PR #420, opened 2016-07-22
Fix check for self-signed certificate

dcooper16:
The check for whether a certificate is self-signed was using the undefined variable `$CN` rather than `$cn`.

Note: I didn't change the rest of the line, but I can't understand why `issuer_O` being `issuer=` or `issuer= ' would indicate that a certificate is self-signed certificate.

It seems that a better check would be to compare the entire issuer name to the entire subject name, since the names could be different even if they have the same `CN` attribute:

```
   issuer=$($OPENSSL x509 -in "$HOSTNAME" -noout -issuer | sed '/^issuer= //')
   subject=$($OPENSSL x509 -in "$HOSTNAME" -noout -subject | sed 's/^subject= //')
   if [[ "$issuer" == "$subject ]]; then
     # certificate is self-signed
   else
    # certificate is not self-signed
   fi
```

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4151,7 +4151,7 @@ certificate_info() {
      issuer_C="$(awk -F'=' '/ C=/ { print $2 }' <<< "$issuer")"
      issuer_DC="$(awk -F'=' '/DC=/ { print $2 }' <<< "$issuer")"
 
-     if [[ "$issuer_O" == "issuer=" ]] || [[ "$issuer_O" == "issuer= " ]] || [[ "$issuer_CN" == "$CN" ]]; then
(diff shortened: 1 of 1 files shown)


dcooper16:
I think I found an even more direct way to check whether a certificate is self-signed:
```
(export SSL_CERT_DIR="/dev/null; export SSL_CERT_FILE=/dev/null"
$OPENSSL verify -check_ss_sig -CAfile "$HOSTCERT" "$HOSTCERT" > "$HOSTCERT.ssout" 2>> $ERRFILE)
if grep -q "error" "$HOSTCERT.ssout"; then
     # certificate is not self-signed
else
     # certificate is self-signed
fi
```

drwetter:
Oh, yes. Eagle eye ;-)

Thx

drwetter:
The construct with the issuers is a legacy workaround for CAcert certificates. Not sure whether it is still needed at all.

PR #400, opened 2016-07-04
See issue #195 - CA pinning checks for intermediates and Root CAs

seccubus:
TODO:
* Unit tests - Done
* Better output - Done

All done, I feel this can be merged @drwetter

PR #422, opened 2016-07-25
Updated CA pinning PR

seccubus:
This PR builds on PR #400 

I've created a script for creating ca_hashes.txt

I've also addressed your comments.

Github.com does indeed pin a root CA that they don't have a certificate with as a counter measure against CA compromise.

I suggest we close #400 and work on from here.


drwetter:
Thx a lot @seccubus !

PR #424, opened 2016-07-25
SSLv2 fixes for run_rc4()

dcooper16:
This PR changes run_rc4() to correctly handle SSLv2 ciphers.

It addresses the problem in which servers that do not implement SSLv2, but that implement SSLv3 ciphers that share an OpenSSL name with an SSLv2 cipher (RC4-MD5 and EXP-RC4-MD5), are not incorrectly shown as having implemented the SSLv2 cipher.

It also addresses the problem that if a server does implement SSLv2 with an RC4 SSLv2-cipher suite, then that cipher suite it not shown as being implemented.


drwetter:
Thx again!

drwetter:
(tested & works as expected)

PR #425, opened 2016-07-25
SSLv2 fixes for test_just_one()

dcooper16:
This PR changes test_just_one() to correctly handle SSLv2 ciphers.

As with PR #424, this PR addresses the problem in which servers that do not implement SSLv2, but that implement RC4-MD5, EXP-RC2-CBC-MD5, EXP-RC4-MD5, or NULL-MD5 are shown as implementing both the SSLv2 and SSLv3 versions of the ciphers, and that any SSLv2 ciphers that a server does implement are not shown as being implemented.


drwetter:
Thx!

PR #427, opened 2016-07-26
SSLv2 fixes for server preference

dcooper16:
This PR addresses two issues related to SSLv2 for "--server-preference" checks.

First, some versions of OpenSSL that support SSLv2 will fail if `s_client` is provided both the `-ssl2` and `-servername` options.

Second, the line for extracting the chosen cipher,`cipher=$(awk '/Cipher.*:/ { print $3 }' $TMPFILE)`, fails for SSLv2. For SSLv2, the output from `$OPENSSL s_client` is as shown below, and the `cipher=` line extracts the word `between` from `Ciphers common between both SSL endpoints:` rather than `IDEA-CBC-MD5` from `   Cipher    : IDEA-CBC-MD5`.

```
...
Ciphers common between both SSL endpoints:                                                                                                                                                                   
RC4-MD5         RC2-CBC-MD5     IDEA-CBC-MD5                                                                                                                                                                 
DES-CBC-MD5     DES-CBC3-MD5                                                                                                                                                                                 
---                                                                                                                                                                                                          
SSL handshake has read 1191 bytes and written 373 bytes                                                                                                                                                      
---                                                                                                                                                                                                          
New, SSLv2, Cipher is IDEA-CBC-MD5                                                                                                                                                                           
Server public key is 2048 bit                                                                                                                                                                                
Compression: NONE                                                                                                                                                                                            
Expansion: NONE                                                                                                                                                                                              
No ALPN negotiated                                                                                                                                                                                           
SSL-Session:                                                                                                                                                                                                 
    Protocol  : SSLv2                                                                                                                                                                                        
    Cipher    : IDEA-CBC-MD5
...
```


drwetter:
Thx, David. 

For some reason the ``Cipher.*`` works for me -- admittedly your fix is more robust ;-)

PR #428, opened 2016-07-26
 SSLv2 fixes for determine_optimal_proto()

dcooper16:
This PR makes three changes to `determine_optimal_proto()`:
* It no longer tries an empty string for `$OPTIMAL_PROTO` twice.
* It does not include `-servername` for `-ssl2` or `-ssl3`, since some versions of OpenSSL that support SSLv2 will fail if `s_client` is provided both the `-ssl2` and `-servername` options.
* It displays a warning if `$OPTIMAL_PROTO` is `-ssl2`, since some tests in testssl.sh will not work correctly for SSLv2-only servers.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7839,9 +7839,11 @@ determine_optimal_proto() {
                all_failed=0
           done
           debugme echo "OPTIMAL_PROTO: $OPTIMAL_PROTO"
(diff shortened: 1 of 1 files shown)


drwetter:
thx!

PR #429, opened 2016-07-26
Don't ignore response to old OpenSSL warning

dcooper16:
In the check for old versions of OpenSSL, the results of the call to `ignore_no_or_lame()` are ignored, and so the program continues even if the user enters `no`.


drwetter:
oh, yes. Thx

Issue #230, opened 2015-11-06
Better support for internal CAs


Harinus:
A feature for exporting all trusted certificates from the local Browser (IE, Firefox, Chrome and others) would be nice (Manual or even automatic)
Imagine following Use-Case:
Using testssl.sh to check hundreds of internal webservers with certificates issued from a internal PKI (root Certificate self-signed) and certificates issued from public Root CAs.
To verify the Chain of trust effectively it is required to have all company issued certificates and the built-in modules in one PEM file.
I did this manually but its a lot of work and enforces maintenance when certificates will be renewed.

-> separated from Issue #199

drwetter:
The framework is basically there but the output is not that nice yet (see pic for one of my private CAs). You just need to put a the public key as *.pem of the Root-CA in the etc directory.

![plugin1](https://cloud.githubusercontent.com/assets/8036727/10999586/2e3f6d44-849d-11e5-95ae-e1ce4737560b.png)


Note the output of the checks against the other stores -- but I translate here just what verify(1) returns, i.e. X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN.

What is open maybe is to have a switch or so that **this** is the CA is the one to ultimately trust.

Makes sense?

Cheers, Dirk

Harinus:
Yep, i exported the Root-CA as *.pem. But: We have 15 of them in organized in 3 Trees of 3 Layers (Root->Policy->Device / User or External) so generating the .pem is not a task of 30 seconds.

I have a similar output to yours:
![store](https://cloud.githubusercontent.com/assets/5645416/11001175/3a9e466a-84a6-11e5-8159-f91aeb9f2bb1.png)


One point to avoid miscomuication: only my own PEM is green. Which is technically correct, but do i really want to have the information of all the others? which may not even be applicable to my company? therefore a feature for building one PEM for "simulating" a cooperate browser to check against would be what i see as an improvement. 

Your suggested swich sounds like a good solution: Check foreach (linux,mozilla,microsoft) if NOT OK then check(own_ca) if ok then "OK" else "notOK"

drwetter:
Am 11/06/2015 um 04:51 PM schrieb Martin Hoffmann:
> Yep, i exported the Root-CA as *.pem. But: We have 15 of them in organized in 3 Trees of 3
> Layers (Root->Policy->Device / User or External) 

(OMF?)

> so generating the .pem is not a task of 30 seconds.

You probably need to do it once in a while.

[...]

PR #440, opened 2016-08-07
Support for custom CA bundle files via new command line option --ca

okoeroo:
Added support for custom CA bundle files provided via the command line to be part of the test. Dropping the files in the $installdir/etc/ is not always convenient when you need to switch between (internal) CAs for detailed testing.

For robustness reasons, all CA bundle files can now also contain spaces in their path. This is part of the pull request as user-input could cause inconvenient errors and break the script when using files with a space.

Example: The internal CA is provided via the $installdir/etc/ directory and via the --ca flag together with the actual root CA file also provided via the command line option.

Command line: 
./testssl.sh  -S --ca etc/KPN\ internal.pem   --ca 111e6273.0   www.kpn.com

Output:
[...snip...]
Chain of trust               NOT ok: KPN internal (chain incomplete) KPN_internal (chain incomplete) KPN_misconfiguration_test (chain incomplete) KPN internal (chain incomplete)
                              OK: 111e6273.0 Mozilla Microsoft Linux Apple
[...snip...]

Example: When file does not exist

Command line:
bash-3.2$ ./testssl.sh  -S --ca foo  --ca 111e6273.0   www.kpn.com

Output:
bash-3.2$ ./testssl.sh  -S --ca foo  www.kpn.com

Provided CA bundle file does not exist or is not readable


drwetter:
Cool, thank you @okoeroo. 

Two obstacles: 1st: It's feature freeze. 2nd: the solution I had in mind was one like described in #230 i.e. throwing all private CAs into the etc dir with a different extension (here: 'pem' in caps) and **automagically** test them. It has two advantages: 1) not yet another cmdline option (minor) 2) mass testing is easier.

For 2) it's important how to display the result. Looking at what you did: It appears the right thing to me. So the code is quite useful.

Thus my question: Would you mind to change your PR to the solution described?

Issue #352, opened 2016-05-14
Shell warnings during Heartbleed and CCS tests  @ OSX?


xeron:
`testssl.sh`: latest master.
`OS X`: 10.11.4.
`bash`: 4.4.0(1)-rc1

Tried with default bash from OS X (3.2.57(1)-release) – same result.

```
$ ./testssl.sh --wide my.example.com
…
 Heartbleed (CVE-2014-0160)                ./testssl.sh: line 1445: warning: command substitution: ignored null byte in input
./testssl.sh: line 1445: warning: command substitution: ignored null byte in input
./testssl.sh: line 1445: warning: command substitution: ignored null byte in input
[REPEATED MULTIPLE TIMES]
…
not vulnerable (OK) (timed out)
 CCS (CVE-2014-0224)                       ./testssl.sh: line 1445: warning: command substitution: ignored null byte in input
./testssl.sh: line 1445: warning: command substitution: ignored null byte in input
./testssl.sh: line 1445: warning: command substitution: ignored null byte in input
[REPEATED MULTIPLE TIMES]
…
```

drwetter:
Could you pls give it a swing?

xeron:
Nope, the same issue for both Heartbleed and CCS:
```
 Heartbleed (CVE-2014-0160)                ./testssl.sh: line 1442: warning: command substitution: ignored null byte in input
./testssl.sh: line 1442: warning: command substitution: ignored null byte in input
```

drwetter:
Am 28. Mai 2016 22:08:48 MESZ, schrieb Ivan Larionov <notifications@github.com>:
>Nope, the same issue for both Heartbleed and CCS:
>```
>Heartbleed (CVE-2014-0160)                ./testssl.sh: line 1442:
>warning: command substitution: ignored null byte in input
>./testssl.sh: line 1442: warning: command substitution: ignored null
>byte in input
>```

Thx. 

I doubt this is a typical OS X behavior. But as a baseline @ all: can anybody please double check?

@ Ivan: you need to provide proper input if this should be a bug report, see wiki.

Cheers, Dirk 


-- 
Set from my mobile. Excuse my brevity&typos

drwetter:
That's a bash 4.4 "bug" which hiccups at NULL bytes : https://github.com/dracutdevs/dracut/issues/118#issuecomment-190857829

Better CCS and HB need to be changed to the socket functions used elsewhere (``sockread_serverhello()``)

[...]

PR #443, opened 2016-08-10
Replace sockread() with sockread_serverhello()

dcooper16:
This PR is in response to issue #352, where it was noted that Bash does not support binary data in strings.

I replaced all calls to `sockread()` with calls to `sockread_serverhello()`, and then, since it now used everywhere and not just to read ServerHello messages, I renamed `sockread_serverhello()` to `sockread()`.

I tested the revised code against several servers, including one that is vulnerable to CCS and Heartbleed, and got the same results as with the current code (although the hexdumps displayed in debug mode differ).

One concern I have is the code in `run_ccs_injection()`. The current code is:
```
     byte6=$(echo "$SOCKREPLY" | "${HEXDUMPPLAIN[@]}" | sed 's/^..........//')
     lines=$(echo "$SOCKREPLY" | "${HEXDUMP[@]}" | count_lines )
     debugme echo "lines: $lines, byte6: $byte6"

     if [[ "$byte6" == "0a" ]] || [[ "$lines" -gt 1 ]]; then
          pr_done_best "not vulnerable (OK)"
...
```
I revised this to:
```
     if [[ -s "$SOCK_REPLY_FILE" ]]; then
          byte6=$(hexdump -ve '1/1 "%.2x"' "$SOCK_REPLY_FILE" | sed 's/^..........//')
          lines=$(hexdump -ve '16/1 "%02x " " \n"' "$SOCK_REPLY_FILE" | count_lines )
          debugme echo "lines: $lines, byte6: $byte6"
     fi
     rm "$SOCK_REPLY_FILE"
     if [[ "$byte6" == "0a" ]] || [[ "$lines" -gt 1 ]]; then
...
```
In the revised code `byte6` is initialized to `0a` so that the response is `not vulnerable (OK)` if `$SOCK_REPLY_FILE` is empty. This has worked okay since for all of the servers that I tested that weren't vulnerable `$SOCK_REPLY_FILE` was empty. Since I haven't seen any other examples, I don't understand why check for vulnerability was written the way it was. So, I'm a bit concerned that the test in the revised code may produce incorrect results now that `hexdump -ve '1/1 "%.2x"' "$SOCK_REPLY_FILE"` is an accurate hexdump of the reply.


drwetter:
Cool! That was something I wanted to do already some time ago.

Would you mind to rebase? 

Q #1: Regarding testing, did you test against  vulnerable servers, too?
Q #2: Any reason for renaming ``sockread_serverhello`` --> ``sockread()``?

drwetter:
A vulnerable server have here for testing answers:

```
[..]
Testing for CCS injection vulnerability 

 CCS (CVE-2014-0224)                       
sending client hello, 
reading server hello

1st reply: 

payload #2 with TLS version x03, x03:

2nd reply: 
00000000  15 03 03 02 02 15                                 |......|
00000006

lines: 1, byte6: 150a
VULNERABLE (NOT ok)
```

Your version returns:
```
1st reply: 

payload #2 with TLS version x03, x03:

2nd reply: 
00000000  15 03 03 00 02 02 15                              |.......|
00000007
lines: 1, byte6: 0215
VULNERABLE (NOT ok)
```

While the finding is the same I am currently wondering about the different length of the second reply and as a result the content of "byte6" (which is actually rather a word6)

dcooper16:
As I noted above, I'm a bit concerned about this function as well.

In the case of the vulnerable server that you tested (above), I got the same results as you. I believe that the payload shown by the PR version (`15 03 03 00 02 02 15`) is the correct one, as it is a correctly formatted message indicating a fatal alert of `decryption_failed_RESERVED`. In the current code, the first byte of the fragment length (`00`) is being removed by `sockread()` due to the problems of trying to store binary data in a bash string.

The other reason for the difference in `byte6` is that the `echo` in the current code is adding a newline (`0a`) to the end of the string being sent to `sed 's/^..........//'`, whereas the PR code doesn't.

I only found one vulnerable server, and it returned a fatal alert of `decryption_failed_RESERVED` (same as yours), and for all of the other servers `$SOCK_REPLY_FILE` was empty.

I've been trying to understand what the test for the vulnerability is trying to do, but I can't figure it out:
```
     byte6=$(echo "$SOCKREPLY" | "${HEXDUMPPLAIN[@]}" | sed 's/^..........//')
     lines=$(echo "$SOCKREPLY" | "${HEXDUMP[@]}" | count_lines )
     debugme echo "lines: $lines, byte6: $byte6"

     if [[ "$byte6" == "0a" ]] || [[ "$lines" -gt 1 ]]; then
```
From my testing of the first two lines, `$lines` seems to be greater than 1 if `$SOCKREPLY` is 16 or more bytes in length. When I test the first line, I seem to get `$byte6` equal to `0a` if `$SOCKREPLY` is either empty or exactly 5 bytes long (with the PR code, `$byte6` is empty  if `$SOCKREPLY` is either empty or exactly 5 bytes long).

I understand that a server isn't vulnerable if `$SOCKREPLY` is empty, but why would a `$SOCKREPLY` that is 5 bytes long or one that is longer than 15 bytes also indicate that the server isn't vulnerable?

I found a script that is supposed to test for the OpenSSL CCS injection vulnerability at https://github.com/Tripwire/OpenSSL-CCS-Inject-Test. I'm not familiar with Python, but if it sends a second message (after the CCS injection), it seems to only declare the server not vulnerable, if no response it sent.

drwetter:
Needed to dig deeper and recall what I did and why 1,5 years ago ;-) From what I currently recall I borrowed the idea from https://gist.github.com/typista/f59a4a55c250300e6b7c and there's this line ``buf[i + 5] == '\x02') { // fatal`` which reflects the old code but not your with the additional '00'. 

The reply with 0a only is clear to me. 

BTW: testssl.sh sends also two messages, it just needs to otherwise the test won't work

drwetter:
``15 03 03 00 02 02 15`` is what I see in wireshark

dcooper16:
I tried running the program from https://gist.github.com/typista/f59a4a55c250300e6b7c, and it seems to create a lot of false positives. It doesn't work in the same way as the code in testssl.sh.

After receiving the response to the ClientHello, the code in https://gist.github.com/typista/f59a4a55c250300e6b7c sends two CCS messages in a row, **without performing a read in between**, and then declares the server to be vulnerable if it does not receive a fatal unexpected message alert (`0a`). The problem is that some servers respond with other fatal alerts (e.g., handshake failure), and neither testssl.sh nor https://github.com/Tripwire/OpenSSL-CCS-Inject-Test consider such servers to be vulnerable.

I am still concerned about the code that is currently in testssl.sh, since the behavior of `sed 's/^..........//'` seems to be so unusual. While it seems that the intention is to get the sixth byte of the response, `$byte6` will have a value of `0a` if `$SOCKREPLY` is empty, and in my testing it is the empty `$SOCKREPLY` that leads to a "not vulnerable (OK)" response, not a `$SOCKREPLY` in which the sixth byte is `0a`. (It is the response to the first CCS message that would have the fatal unexpected message alert, but testssl.sh doesn't check that response.)

Since https://github.com/Tripwire/OpenSSL-CCS-Inject-Test seems to produce correct results, I tried changing `run_ccs_injection()` to mimic that code (although not perfectly).

In this version, if the server responds to the (first) CCS message with a fatal alert, then it declares the server to be not vulnerable. Otherwise, it sends a close_notify warning message, and declares the server to be vulnerable if it gets a non-empty response.

The one vulnerable server that I know about behaves in the same way as yours does (no response to the first CCS message, and decryption failed alert in response to the second one), and this code detects the vulnerability. It also declares all of the other servers that I tested to be not vulnerable, some of which respond to the CCS message with a fatal alert and some of which provide no response to either message.

drwetter:
Thx.

The actual URL was a different one: https://gist.github.com/rcvalle/71f4b027d61a78c42607 but that gist doesn't exist anymore. Same author though and the code on my disk is pretty similar. Probably the code I implemented was not only from him. </current alzheimer> ;-)

Actually as I never myself experiences FP myself -- and I did thorough testing, other complaints I never got either -- at this moment (2.8rcX) I am hestitant to merge this. Need to do further research and testing. 

I believe though you did a good job here.

dcooper16:
Okay. I guess any attempt to modify `run_ccs_injection()` should be done in a separate PR that could be considered for after 2.8 is released.

So, I modified this PR to minimize the changes that it makes. I undid the renaming of `sockread_serverhello()` (even though it is now used for more than reading ServerHello messages) and I changed `run_ccs_injection()` so that it now it works exactly the same as in the current code. I did this by adding a line:
```
     sockreply=$(cat "$SOCK_REPLY_FILE" 2>/dev/null)
```
and then using `$sockreply` to obtain `$byte6` and `$lines`.

drwetter:
Thanks! It needed some testing before merge, excuse the delay

PR #445, opened 2016-08-11
Fix "--devel" with SSLv2

dcooper16:
If testssl.sh is called with `--devel 22` and the response from `sslv2_sockets()` is not 0, then `tls_sockets()` will be called, and the result of the `tls_sockets()` command will be output rather than the result of the `sslv2_sockets()` command.


drwetter:
Thx for your careful review and fix, David!

PR #449, opened 2016-08-23
Warning on Poodle test when no local SSLv3 support

dcooper16:
If the version of OpenSSL being used doesn't support `s_client -ssl3` (e.g., OpenSSL 1.1.0), `run_ssl_poodle()` displays "not vulnerable (OK)" even though it can't test whether the server is vulnerable.

This PR fixes it so that a "Local problem" warning is displayed if `s_client -ssl3` isn't supported.

The PR also removes the `$SNI` from the call to `$OPENSSL s_client`, since SSLv3 doesn't have SNI.


drwetter:
Thx, David

PR #450, opened 2016-08-23
Fix SSLv2-only test in determine_optimal_proto()

dcooper16:
The test for whether a server only supports SSLv2 was broken, since `$OPTIMAL_PROTO` will be `-ssl2` whether SSLv2 is the only protocol that succeeds or no protocol succeeds.

This PR sets `$OPTIMAL_PROTO` (or `$STARTTLS_OPTIMAL_PROTO`) to "" if no protocol succeeds.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7828,6 +7828,7 @@ determine_optimal_proto() {
                fi
                all_failed=0
           done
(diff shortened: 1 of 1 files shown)


drwetter:
Thx!

PR #451, opened 2016-08-23
Warning on BEAST when no local SSLv3 support

dcooper16:
If the version of OpenSSL being used doesn't support `s_client -ssl3` (e.g., OpenSSL 1.1.0), `run_beast()` doesn't display a warning that testing for CBC in SSLv3 isn't locally supported.

This PR adds a "Local problem" warning if the OpenSSL being used doesn't support `s_client -ssl3`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6742,11 +6742,21 @@ run_beast(){
      done
 
      for proto in ssl3 tls1; do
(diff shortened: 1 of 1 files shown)


drwetter:
Thx!

Issue #454, opened 2016-08-25
Certification validation problem - Accept any purpose 


nep7un:
Hi,

the code only verify if the sslserver purpose is enabled in the certificate. If not openssl gives error 26 X509_V_ERR_INVALID_PURPOSE. This is cool, however if I dump my certificate and see what purposes are enabled the following two are set:
Any Purpose : Yes
Any Purpose CA : Yes

The code should also check if any purpose is checked, not only sslserver.

regards,
Zsombor

dcooper16:
I don't think there is anything that testssl.sh can do about this. If OpenSSL reports that verification of the certificate failed with reason 26, then testssl.sh could check the value of the extended key usage extension and indicate that the certificate is okay, even though OpenSSL says otherwise, but that might be dangerous. Could we be certain that the certificate is okay otherwise?

It seems that if this is to be fixed, then it would need to be fixed in OpenSSL. This does seem to be an error in OpenSSL. [RFC 5280](https://www.ietf.org/rfc/rfc5280.txt) says:

> Certificate using applications MAY require that the extended key usage extension be present and that a particular purpose be indicated in order for the certificate to be acceptable to that application.
> 
> Applications that require the presence of a particular purpose MAY reject certificates that include the anyExtendedKeyUsage OID but not the particular OID expected for the application.

So, OpenSSL could require that TLS server certificates have an extended key usage extension that includes the id-kp-serverAuth OID. But, since OpenSSL accepts TLS server certificates that don't have an extended key usage extension, the second sentence above suggests that it should also accept a certificate that has an extended key usage extension with the anyExtendedKeyUsage OID (it seems that an application that doesn't require the extended key usage extension to be present doesn't "require the presence of a particular purpose"). This is open to interpretation, however, so others could interpret the text from [RFC 5280](https://www.ietf.org/rfc/rfc5280.txt) differently.

I am also sympathetic to the idea of flagging such certificates as problematic. I don't think a server certificate should be a general-purpose certificate (as the presence of anyExtendedKeyUsage would seem to imply). I think a server certificate should be issued for a specific purpose, or at least a limited set of purposes, and so if the extended key usage extension is included it should identity the specific purpose(s) for which the certificate was issued, rather than indicating that the certificate can be used for any purpose.

drwetter:
Thx for reporting @nep7un .

It should be addressed by @dcooper16 in  #455 . Pls let me know if not.

PR #455, opened 2016-08-25
Output correct error for unsupported certificate purpose

dcooper16:
This PR is in response to issue #454. I tried repeating the reported problem by creating a certificate in which the extendedKeyUsage extension was present and only included the anyExtendedKeyUsage OID. In running the test, I discovered two problems.

First, when `determine_trust()` is calling `verify_retcode_helper()` to display the reason that path validation failed, it assumes that there are at least two certificate bundles provided. (I was running the test using just one certificate bundle, containing my local root.) So, I changed `determine_trust()` to use `${verify_retcode[1]}` rather than `${verify_retcode[2]}` in the case that all bundles failed (it seems that 2 vs. 1 was an arbitrary choice).

Once that was fixed, testssl.sh output "NOT ok  (unknown, pls report) 26". So, the second thing this PR fixes is to output "NOT ok (unsupported certificate purpose)" if OpenSSL responds with an unsupported certificate purpose error.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3697,6 +3697,7 @@ verify_retcode_helper() {
 
 	case $retcode in
 		# codes from ./doc/apps/verify.pod | verify(1ssl)
+		26) out "(unsupported certificate purpose)" ;; 	# X509_V_ERR_INVALID_PURPOSE
 		24) out "(certificate unreadable)" ;; 	# X509_V_ERR_INVALID_CA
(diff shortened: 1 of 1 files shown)


drwetter:
Cool!

dcooper16:
Hi @drwetter,

Did you mean to close this PR without merging it?

Thanks,

David

drwetter:
Sorry, of course not @dcooper16 .

Thx for catching, Dirk

PR #456, opened 2016-08-25
Support DH server keys

dcooper16:
When `certificate_info()` is given a certificate with a DH public key it displays something like:
```
   Server key size              fixme:  dhKeyAgreement 3072 bits (FIXME: can't tell whether this is good or not)
```

This PR fixes that so that the output is:
```
   Server key size              DH 3072 bits
```


drwetter:
thx!

Issue #447, opened 2016-08-15
No TLS SNI support for HTTPS certificate checks


ogmueller:
Versions:
```
# ./testssl.sh -b 2>/dev/null | head -4 | tail -1
    (424cf23 2016-08-09 10:35:58 -- 1.531)
```

```
# ./testssl.sh -b 2>/dev/null | head -16 | tail -3
 Using "OpenSSL 1.0.2-chacha (1.0.2i-dev)" [~183 ciphers]
 on snell:./bin/openssl.Linux.x86_64
 (built: "Jun 22 19:32:29 2016", platform: "linux-x86_64")
```

```
# uname -a
Linux test 4.4.0-34-generic #53-Ubuntu SMP Wed Jul 27 16:06:39 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```

---

Target IP: 78.47.205.213
Target Website: https://ugi.teqneers.de/
 
---

The bug we found with testssl is, that it doesn't seem to support SNI (Server Name Indication). So if you connect to ugi.teqneers.de with HTTPS, testssl tells you, that it receives a self-signed certificate. The problem here is, that if you connect to the server using an IP address, it will just give you a self-signed cert, but the website, which i am trying to test does not. so if you use e.g. google chrome and visit https://ugi.teqneers.de the certificate is perfectly fine, because the server delivers different certs depending on the domain name and virtual server settings of the web server.

```
# ./testssl.sh -S ugi.teqneers.de

...

  Server Certificate #2
   Signature Algorithm          SHA256 with RSA
   Server key size              RSA 2048 bits
   Fingerprint / Serial         SHA1 FF5727BB40414B567C0BAE68F752DFBA68697D4F / F90197C8EBFD4004
                                SHA256 23824550CFC6E71A4294909E840271A958B5A1B0E4E489AFE2A68E34930C7B4C
   Common Name (CN)             "ugi.teqneers.de"
   subjectAltName (SAN)         --
   Issuer                       self-signed (NOT ok)
   Trust (hostname)             Ok via CN (works w/o SNI)
   Chain of trust               NOT ok (self signed)
   EV cert (experimental)       no
   Certificate Expiration       3593 >= 60 days (2016-06-20 11:24 --> 2026-06-18 11:24 +0000)
   # of certificates provided   1
   Certificate Revocation List  --
   OCSP URI                     --
   OCSP stapling                --

...

```

ogmueller:
[...]

PR #460, opened 2016-08-30
Display SNI information in "Server Certificate" line

dcooper16:
This PR is an attempt to address issue #447. If more than one certificate is being displayed, then a parenthetical saying "(in response to request w/o SNI)" is added for any certificate that was obtained using `$SNI=""`.

In addition, if the certificate was obtained without SNI, then `certificate_info()` doesn't call `$OPENSSL s_client` in order to obtain the non-SNI host certificate and it does not display a separate "Trust (hostname)" finding for the non-SNI certificate.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4113,6 +4113,7 @@ certificate_info() {
      local cert_keysize=$4
      local ocsp_response=$5
      local ocsp_response_status=$6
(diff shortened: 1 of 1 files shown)


drwetter:
looks good, thx!

PR #461, opened 2016-09-01
Fix tls_sockets() when SNI empty

dcooper16:
`socksend_tls_clienthello()` always includes a server name extension in the ClientHello (for TLS 1.0 and above), even if `$SNI` is empty. If `$NODE` is an IP address, then the IP address is placed in the extension, even though RFC 6066 says that only DNS names are supported in the extension.

This PR changes `socksend_tls_clienthello()` so that the server name extension is only included in the ClientHello if `$SNI` is not empty.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -5747,20 +5747,22 @@ socksend_tls_clienthello() {
                fi
           done
 
-          #formatted example for SNI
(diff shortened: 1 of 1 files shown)


drwetter:
thx

PR #468, opened 2016-09-12
compare_server_name_to_cert(): Fix unassigned vars.

knweiss:
Two instances of referenced but not assigned variables (`req` instead of `ret`).
```
In testssl.sh line 4130:
     if [[ $req -eq 0 ]]; then
           ^-- SC2154: req is referenced but not assigned.
```
Found by ShellCheck.


drwetter:
thx again

PR #472, opened 2016-09-14
run_rp_banner(), run_application_banner(): Three issues

knweiss:
- Fix two instances of referenced but not assigned variables.
- Fix a modified in subshell bug in `run_application_banner()`
- Fix indentation in `run_rp_banner()`


drwetter:
thx

PR #473, opened 2016-09-15
Fix handling of empty argument to "-nextprotoneg" parameter

### LINKED ISSUE #467: OCSP stapling status shown as unknown in contradiction to manual s_client check
testssl.sh current master reports an unknown OCSP stapling status even for known-to-be-good apache2 virtualhosts as reported by openssl s_client (Debian/Jessie).

**s_client:**
```
$ openssl version
OpenSSL 1.0.1t  3 May 2016

$ openssl s_client -connect 78.46.72.130:443 -servername nachtgeist.eu.org -tls1_2 -tlsextdebug -status | head -n20
CONNECTED(00000003)
TLS server extension "server name" (id=0), len=0
TLS server extension "renegotiation info" (id=65281), len=1
0001 - <SPACES/NULS>
TLS server extension "EC point formats" (id=11), len=4
0000 - 03 00 01 02                                       ....
TLS server extension "session ticket" (id=35), len=0
TLS server extension "status request" (id=5), len=0
TLS server extension "heartbeat" (id=15), len=1
0000 - 01                                                .
depth=2 O = Digital Signature Trust Co., CN = DST Root CA X3
verify return:1
depth=1 C = US, O = Let's Encrypt, CN = Let's Encrypt Authority X3
verify return:1
depth=0 CN = nachtgeist.eu.org
verify return:1
OCSP response: 
======================================
OCSP Response Data:
    OCSP Response Status: successful (0x0)
    Response Type: Basic OCSP Response
    Version: 1 (0x0)
    Responder Id: C = US, O = Let's Encrypt, CN = Let's Encrypt Authority X3
    Produced At: Sep  8 10:17:00 2016 GMT
    Responses:
    Certificate ID:
```

**testssl.sh:**
```
$ testssl -S https://nachtgeist.eu.org

No mapping file found

    testssl       2.8rc2 from https://testssl.sh/dev/
    (1.541 2016/09/07 19:34:26)

      This program is free software. Distribution and

 Using "OpenSSL 1.0.1t  3 May 2016" [~104 ciphers]
 on enterprise:/usr/bin/openssl
 (built: "May 15 19:23:55 2016", platform: "debian-amd64")


 Start 2016-09-12 03:58:01    -->> 78.46.72.130:443 (nachtgeist.eu.org) <<--

 further IP addresses:   2a01:4f8:120:43c7:0:5::
 rDNS (78.46.72.130):    srv05.itamservices.de.
 Service detected:       HTTP


 TLS extensions (standard)    "server name/#0" "renegotiation info/#65281" "EC point formats/#11" "session ticket/#35" "heartbeat/#15"
 Session Tickets RFC 5077     300 seconds (PFS requires session ticket keys to be rotated <= daily)
 SSL Session ID support       yes
 TLS clock skew               random values, no fingerprinting possible 
 Signature Algorithm          SHA256 with RSA
 Server key size              RSA 4096 bits
 Fingerprint / Serial         SHA1 9B81D80C00679604B8C3C179C80B61EFBF5696CA / 03677DD93BFEDB5C22527C778CC77B8FCADE
                              SHA256 C64BCB853EB8191ADF6D9BD632C3F66817BD969B2CDCE8D65B0BF2F264F7BCEE
 Common Name (CN)             "nachtgeist.eu.org" (CN in response to request w/o SNI: "itamservices.de")
 subjectAltName (SAN)         "nachtgeist.eu.org" "www.nachtgeist.eu.org" 
 Issuer                       "Let's Encrypt Authority X3" ("Let's Encrypt" from "US")
 Trust (hostname)             Ok via SAN and CN (SNI mandatory)
 Chain of trust               "/usr/bin/etc/*.pem" cannot be found / not readable
 EV cert (experimental)       no 
 Certificate Expiration       30 >= 30 days (2016-07-14 10:35 --> 2016-10-12 10:35 +0200)
 # of certificates provided   2
 Certificate Revocation List  --
 OCSP URI                     http://ocsp.int-x3.letsencrypt.org/
 OCSP stapling                (response status unknown)


 Done 2016-09-12 03:58:07    -->> 78.46.72.130:443 (nachtgeist.eu.org) <<--
```

In the traces above the testing was done against a vhost which has a LetsEncrypt-certificate configured via SNI.

Currently, the default vhost listening on the IP has a certificate configured which is signed by a not-well-known CA. However, testssl reported the same unknown OCSP response status, when I temporarily set the LetsEncrypt cert mentioned above also for the default vhost.

nachtgeist:
s_client's manpage states for -nextprotoneg:

"Empty list of protocols is treated specially and will cause the client to
advertise support for the TLS extension but disconnect just after reciving
ServerHello with a list of server supported protocols."

Consequently, the previous workaround of just quoting an empty variable is
insufficient and the "-nextprotoneg" parameter has to be removed entirely from
the command-line in case of an empty argument.

In other locations where "-nextprotoneg" is used
- its argument cannot be empty ($NPN_PROTOs is initialized to a non-empty value
  and set read-only) or
- its argument is intended to be empty (line 3724) or
- the command will not be invoked at all (for-loop parameter, line 3725)

This fixes #467 - again.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3933,11 +3933,15 @@ sclient_connect_successful() {
 determine_tls_extensions() {
      local proto addcmd
      local success
-     local alpn=""
+     local alpnOrNpnParam=""
      local savedir
      local nrsaved
 
-     "$HAS_ALPN" && alpn="h2-14,h2-15,h2"
+     if "$HAS_ALPN"; then
+          alpnOrNpnParam="-alpn \"http/1.1,spdy/1,spdy/2,spdy/3,stun.turn,stun.nat-discovery,h2,h2c,webrtc,c-webrtc,ftp\""
+     elif "$HAS_SPDY"; then
+          alpnOrNpnParam="-nextprotoneg \"h2-14,h2-15,h2\""
+     fi
 
      if [[ -n "$2" ]]; then
          protocols_to_try="$2"
@@ -3983,7 +3987,7 @@ determine_tls_extensions() {
 # alpn: echo | openssl s_client -connect google.com:443 -tlsextdebug -alpn h2-14 -servername google.com  <-- suport needs to be checked b4 -- see also: ssl/t1_trce.c
           addcmd=""
           [[ ! "$proto" =~ ssl ]] && addcmd="$SNI"
-          $OPENSSL s_client $STARTTLS $BUGS $1 -showcerts -connect $NODEIP:$PORT $PROXY $addcmd -$proto -tlsextdebug -nextprotoneg "$alpn" -status </dev/null 2>$ERRFILE >$TMPFILE
+          $OPENSSL s_client $STARTTLS $BUGS $1 -showcerts -connect $NODEIP:$PORT $PROXY $addcmd -$proto -tlsextdebug $alpnOrNpnParam -status </dev/null 2>$ERRFILE >$TMPFILE
           sclient_connect_successful $? $TMPFILE && success=0 && break
      done                          # this loop is needed for IIS6 and others which have a handshake size limitations
      if [[ $success -eq 7 ]]; then


testssl.sh
```diff
@@ -3933,11 +3933,11 @@ sclient_connect_successful() {
 determine_tls_extensions() {
      local proto addcmd
      local success
-     local alpn=""
+     local alpnParam=""
      local savedir
      local nrsaved
 
-     "$HAS_ALPN" && alpn="h2-14,h2-15,h2"
+     "$HAS_ALPN" && alpnParam="-nextprotoneg h2-14,h2-15,h2"
```
drwetter: This is anyway nonsense. ``ALPN=HTTP2``, ``NPN/nextprotoneg=SPDY``

But honestly I need to understand why this code ``openssl s_client  -showcerts -connect HOST:PORT -tlsextdebug -nextprotoneg "<whatever>" -status`` made it into the for loop.

nachtgeist: > This is anyway nonsense. ALPN=HTTP2, NPN/nextprotoneg=SPDY

Can't follow you. What exactly is nonsense?

drwetter: $HAS_ALPN" && alpnParam="-nextprotoneg h2-14,h2-15,h2".

It should either read  ~ ``$HAS_NPN" && npnParam="-nextprotoneg h2-14,h2-15,h2"`` or ~ ``$HAS_ALPN" && alpnParam="-alpn h2-14,h2-15,h2".``

nachtgeist: Ah, thx.

Since `man s_client` doesn't mention `-alpn` I guess `npnParam` is the one to go with.

dcooper16:
I don't know about the man page, but at least some versions of OpenSSL support both `-alpn` and `nextprotoneg`:
```
 -alpn val                  Enable ALPN extension, considering named protocols supported (comma-separated list)
 -nextprotoneg val          Enable NPN extension, considering named protocols supported (comma-separated list)
```
The Application-Layer Protocol Negotiation (ALPN) extension is defined in [RFC 7301](https://www.rfc-editor.org/rfc/rfc7301.txt), and the list of protocol names that may appear in the extension is available at http://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids.

According to [Wikipedia](https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation), "NPN was replaced with a reworked version, ALPN." It seems that there were a few [Internet-Drafts](https://www.ietf.org/archive/id/draft-agl-tls-nextprotoneg-04.txt) for NPN, but the most recent draft was published in May 2012. So, it seems that in terms of standards, work was started on NPN, but it was abandoned in favor of ALPN.

For NPN, the final Internet-Draft says that:
>    The "extension_data" field of a "next_protocol_negotiation" extension
>    in a "ClientHello" MUST be empty.

So, that presumably means that anything included in the list provided to `-nextprotoneg` is only used by OpensSSL and isn't sent to the server, so it would be up to OpenSSL to decide what protocol names to accept.

Presumably this was included in the `$OPENSSL s_client` line since the function is `determine_tls_extensions()` and including the flag allows one to determine whether the server supports the extension. So, I would suggest using `-alpn`.

nachtgeist:
>  don't know about the man page

see the testssl header output in #467 (Debian/Jessie)


> but at least some versions of OpenSSL support both -alpn and nextprotoneg:

Interesting, thanks


> So, I would suggest using -alpn

You said it yourself: "at least some". So essentially, your suggestion would break testssl.sh on...well...at least "the other some"...systems.

I'm not familiar with openssl's exact history and what parameter is supported since what version and which distro uses which `./configure` parameters, so I can't comment based on knowledge. But Dirk added `-nextprotoneg` in b9bfd488 almost a year ago and obviously nobody complained about it, otherwise it would have been changed by now. So that should remain the primary choice. If you want to use `-alpn`, the called openssl binary really needs to be probed for that feature prior to its usage.

drwetter:
>I'm not familiar with openssl's exact history and what parameter is
>supported since what version and which distro uses which `./configure`
>parameters, so I can't comment based on knowledge. But Dirk added
>`-nextprotoneg` in b9bfd488 almost a year ago and obviously nobody
>complained about it, otherwise it would have been changed by now. So
>that should remain the primary choice. If you want to use `-alpn`, the
>called openssl binary really needs to be probed for that feature prior
>to its usage.

NPN is on openssl for quite some time. Also on the server side a lot of Unix systems support it. As said there it's called SPDY. It was started by Google, later abandoned  in favor of HTTP2  = ALPN. The latter appears in recent openssl versions. Off the top of my head: late 1.0.1 or 1.0.2. Unfortunately on the sevrr side this is not in widespread use yet. Google claimed to withdraw support on their clients for SPDY in favor of HTTP2. Last weeks I looked it was still there.

Long story short: If I remember correctly that line made it into the code as otherwise the according  tls extention won't show up sometimes. Since then Laine added ALPN in other parts of the code. Should be worth a thought whether it is of use here too.

Cheers, Dirk



-- 

nachtgeist:
> Should be worth a thought whether it is of use here too.

Sure is, but well outside the scope of this ticket and the updated pull request to fix the check for OCSP stapling.

dcooper16:
> 
> > So, I would suggest using -alpn
> 
> You said it yourself: "at least some". So essentially, your suggestion would break testssl.sh on...well...at least "the other some"...systems.
...
> If you want to use -alpn, the called openssl binary really needs to be probed for that feature prior to its usage.

As Dirk has been pointing out, the code is already probing the OpenSSL binary for the `-alpn` feature. It's just that at the moment, it is adding `-nextprotoneg` if it has been determined that `-alpn` is supported. So, changing `-nextprotoneg` to `-alpn` would definitely **not** break testssl.sh on any systems.

Presumably the reason the current code isn't breaking any systems is that support for `-nextprotoneg` was added to OpenSSL before support for `-alpn` was added, so any OpenSSL that supports `-alpn` also supports `-nextprotoneg`. Still, as Dirk noted, if the command line is going to have `-nextprotoneg`, then the test needs to be changed from `$HAS_ALPN` to `$HAS_NPN`. But, one could always change the current `"$HAS_ALPN" && alpn="h2-14,h2-15,h2"` to something like:
```
     if "$HAS_ALPN"; then
          alpnParam="-alpn "http/1.1,spdy/1,spdy/2,spdy/3,stun.turn,stun.nat-discovery,h2,h2c,webrtc,c-webrtc,ftp"
     elif "$HAS_NPN"; then
          alpnParam="-nextprotoneg h2-14,h2-15,h2"
     fi
```
Both `-alpn` and `-nextprotoneg` could be included if OpenSSL supported both. However, in the testing that I've done, if both extensions are included in the ClientHello, then only ALPN appears in the ServerHello. There may be servers that support NPN, but not ALPN, and that would return NPN if both were included in the ClientHello, but I haven't seen that yet.

If we want the "TLS extensions" line to list both ALPN and NPN if the server supports both, then two separate calls would be needed.

> 
But Dirk added -nextprotoneg in b9bfd48 almost a year ago and obviously nobody complained about it, otherwise it would have been changed by now. So that should remain the primary choice.

If servers that initially adopted NPN are moving to ALPN, since NPN was experimental and ALPN was the standard that was finally adopted, then it is perfectly reasonable to consider changing testssl.sh to consider whether testssl.sh should change as well.

nachtgeist:
Thanks, David, for this clarification! I didn't make the (obvious) connection between $HAS_* and the actual probing...sorry for that.

I'll update the PR with your suggestet `if` branches.

nachtgeist:
instead of hard-coding those arguments in the `if` branch, shouldn't we re-use `$NPN_PROTOs` and `$ALPN_PROTOs` respectively?

dcooper16:
> instead of hard-coding those arguments in the if branch, shouldn't we re-use `$NPN_PROTOs` and `$ALPN_PROTOs` respectively?

Not necessarily. At the moment `$ALPN_PROTOs` is only used in `run_http2()`, and it has a value of
```
readonly ALPN_PROTOs="h2 h2-17 h2-16 h2-15 h2-14 spdy/3.1 http/1.1"
```
According to http://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids, the allowable values are:
```
http/1.1, spdy/1, spdy/2, spdy/3, stun.turn, stun.nat-discovery, h2, h2c, webrtc, c-webrtc, ftp
```
So, there are a few issues here:

- `run_http2()` needs a space separated list, whereas `determine_tls_extensions()` needs a comma separated list (easily dealt with).

- `run_http2()` is trying some values that are not in the IANA registry (I don't know if that means the value of `$ALPN_PROTOs` should be changed to match the registry or the values that are not in the registry should be included in the call by `determine_tls_extensions()`

- It may be that `determine_tls_extensions()` is supposed to include all values in the IANA registry, but `run_http2()` is only supposed to include those related to HTTP/2, while excluding others (e.g., ftp).

For NPN, `$NPN_PROTOs` is used in related to SPDY and has as value of:
```
readonly NPN_PROTOs="spdy/4a2,spdy/3,spdy/3.1,spdy/2,spdy/1,http/1.1"
```
This is very different from the value "h2-14,h2-15,h2" used in `determine_tls_extensions()`, but I don't know the reason for the difference.

Perhaps @drwetter and/or @lainegholson knows the reasons for the different values that are being used.

I just noticed that testssl.sh already has a flag that indicates whether the OpenSSL binary supports the `-nextprotoneg` flag. It is `$HAS_SPDY`.

drwetter:
Am 09/22/2016 um 04:39 PM schrieb David Cooper:

> 
> For NPN, |$NPN_PROTOs| is used in related to SPDY and has as value of:
> 
> |readonly NPN_PROTOs="spdy/4a2,spdy/3,spdy/3.1,spdy/2,spdy/1,http/1.1" |
> 
> This is very different from the value "h2-14,h2-15,h2" used in |determine_tls_extensions()|, but I don't know the reason for the difference.
> 
> Perhaps @drwetter <https://github.com/drwetter> and/or @lainegholson <https://github.com/lainegholson> knows the reasons for the different values that are being used.

IIRC I did some tests for ALPN and that was what worked best. For NPN this looks strange to me.


> I just noticed that testssl.sh already has a flag that indicates whether the OpenSSL binary supports the |-nextprotoneg| flag. It is |$HAS_SPDY|.

yes, I did a central check for this.

drwetter:
Am 09/22/2016 um 04:04 PM schrieb nachtgeist:
> Thanks, David, for this clarification! I didn't make the (obvious) connection between $HAS_* and the actual probing...sorry for that.
> 
> I'll update the PR with your suggestet |if| branches.

I can confirm at least on https://testssl.sh the ALPN extenstion will only show up if e.g.

openssl s_client -cipher $whatever -showcerts -connect 81.169.199.25:443 -servername testssl.sh -tls1_2 -nextprotoneg h2-14,h2-15,h2 -status

is being used. If I omit ``-nextprotoneg "<someprotocol>"`` it won't show up.

If I use ``-nextprotoneg ""` it shows up -- however ``sclient_connect_successful()`` won't
return 0.

(For ALPN the latter doesn't work at all -- as said/expected.)

I may add that later then....

Cheers, Dirk

drwetter:
Thx but the test only works with only either NPN or APLN, not both, i.e. if the server supports both, you will  only get ALPN returned. 

I'll fix that...

Issue #435, opened 2016-07-29
Add environment variable for certificate location


madsolar8582:
It is possible to overwrite the OpenSSL location using `OPENSSL=`, would it be possible to add another environment variable to set the location of the certificates used during the tests (e.g. `/usr/local/etc/openssl/cert.pem`)?

drwetter:
Hi @madsolar8582,

almost everything is possible .. Often the question is whether it makes sense. ;-) I am not sure what you're suggesting but maybe this #230 is what you want? 

Cheers, Dirk

madsolar8582:
Kinda. In the `determine_trust` function, the ca bundle is set to the installation directory's `etc` [subdirectory](https://github.com/drwetter/testssl.sh/blob/b0923a1833a71c157222318f6124f227a6a2fefc/testssl.sh#L3726). However, if you're just running the script wholesale, then there is no `etc` directory. If the `ca_bundles` variable could be set via an `ENV` setting, that I would think the script would be a bit more portable.

drwetter:
Am 6. August 2016 18:06:46 MESZ, schrieb Madison Solarana <notifications@github.com>:
>Kinda. In the `determine_trust` function, the ca bundle is set to the
>installation directory's `etc`
>[subdirectory](https://github.com/drwetter/testssl.sh/blob/b0923a1833a71c157222318f6124f227a6a2fefc/testssl.sh#L3726).
>However, if you're just running the script wholesale, then there is no
>`etc` directory. If the `ca_bundles` variable could be set via an `ENV`
>setting, that I would think the script would be a bit more portable.

Ok, got it. 

Plan for the near future is to have some code external. So at some time you won't be able to run testssl.sh wholesale.

Also my humble guess is you don't have the certificate stores supplied by testssl.sh  just by coincidence lying around elsewhere... ;-)


Cheers, Dirk 


-- 
Sent from my mobile. Excuse my brevity&typos+the phone's autocorrection

drwetter:
The certificates in "$INSTALL_DIR/etc/*.pem" are needed for this particular test. ``/usr/local/etc/openssl/cert.pem`` is not sufficient. In ""$INSTALL_DIR/etc/" are certificate stores from Linux, Apple, Mozilla, MS.

What do you mean by "wholesale", from a distribution?

madsolar8582:
As in, being able to set the env variables, then pipe the script into bash via curl without having to install the script in its entirety.

drwetter:
[...]

PR #475, opened 2016-09-16
allow ca_bundles path to be configurable from command line

mailsvb:
added a command line option to provide a path to ca certificates to allow for building the chain of trust


drwetter:
This would be for https://github.com/mailsvb/node-red-contrib-testssl, right?

Q: It doesn't make much sense to specify an OWN ca_bundle path as testssl.sh comes with several stores where the certificate chain. One other case I was hestitantly replying: #435 .

drwetter:
I would like to merge this in for 2.8. As it is supposed to have limited use and yhings will change the next release I would not like to have the command line option. 

Also: the patch unfortunately didn't survive travis CI, see details.

Would you mind addressing both points?

PR #478, opened 2016-09-24
Remove duplicated do_rc4 in debug_globals()

wdhongtw:
do_rc4 is duplicated in debug_globals(), but this duplication is unnecessary.

Remove this duplication and make the list in debug_globals() consistent with the one in query_globals().


drwetter:
Thank you

PR #485, opened 2016-09-28
Allow cipher list to be passed to sslv2_sockets()

dcooper16:
This PR changes `sslv2_sockets()` so that a list of ciphers may optionally be passed as an argument. This will support the use of `sslv2_sockets()` in some places where `$OPENSSL s_client` is currently used.


drwetter:
Thank you

PR #487, opened 2016-09-28
Add option for extract data from SSLv2 ServerHello

dcooper16:
This PR adds the option for `parse_sslv2_serverhello()` to extract information from the ServerHello (server key size and cipher suites supported) and write the information to `$TMPFILE` as well as to write the server's certificate to `$HOSTCERT`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -637,6 +637,38 @@ else
 	}
 fi
 
+# arg1: An ASCII-HEX string
+# arg2: file name
+# Append $arg1 in binary format to $arg2
+asciihex_to_binary_file(){
+     local string="$1"
+     local file="$2"
+     local -i len
+     local -i i ip2 ip4 ip6 ip8 ip10 ip12 ip14
+     local -i remainder
+
+     len=${#string}
+     [[ $len%2 -ne 0 ]] && return 1
+
+     for (( i=0; i <= len-16 ; i=i+16 )); do
+          ip2=$i+2; ip4=$i+4; ip6=$i+6; ip8=$i+8; ip10=$i+10; ip12=$i+12; ip14=$i+14
+          echo -e -n "\x${string:i:2}\x${string:ip2:2}\x${string:ip4:2}\x${string:ip6:2}\x${string:ip8:2}\x${string:ip10:2}\x${string:ip12:2}\x${string:ip14:2}" >> "$file"
+     done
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -641,6 +641,38 @@ else
...
+
+     for (( i=0; i <= len-16 ; i=i+16 )); do
+          ip2=$i+2; ip4=$i+4; ip6=$i+6; ip8=$i+8; ip10=$i+10; ip12=$i+12; ip14=$i+14
+          echo -e -n "\x${string:i:2}\x${string:ip2:2}\x${string:ip4:2}\x${string:ip6:2}\x${string:ip8:2}\x${string:ip10:2}\x${string:ip12:2}\x${string:ip14:2}" >> "$file"
+     done
+
+     ip2=$i+2; ip4=$i+4; ip6=$i+6; ip8=$i+8; ip10=$i+10; ip12=$i+12; ip14=$i+14
+     remainder=$len-$i
+     case $remainder in
+           2) echo -e -n "\x${string:i:2}" >> "$file" ;;
```
drwetter: Thank you. 

Two points: 
* printf would be more portable instead of echo -n -e
* asciihex_to_binary_file() doesn't look so elegant to me. That's for parsing the certificate?

drwetter:
To be honest: I couldn't reproduce on FreeBSD why ``echo -n -e`` failed.

dcooper16:
Okay, I created a PR to use `printf` in `asciihex_to_binary_file()` rather than `echo -e -n`.

In `parse_sslv2_serverhello()` this function is used to save the ASCII-HEX encoded certificate from the server's response and save it to a file.

In https://github.com/dcooper16/testssl.sh/blob/extended_tls_sockets/testssl.sh, `asciihex_to_binary_file()` is used many times to extract data from the ASCII-HEX encoded server's response and save the data to files in binary form: the server's ephemeral key, certificates, OCSP response, and the contents of the ALPN and NPN extensions. It is also used in `hmac()` and `derive-secret()` to help derive the symmetric key needed to decrypt parts of the server's response in TLS 1.3.

I realize that the function is a bit inelegant, but my goal was to make is efficient. Having a single loop that writes one byte at a time is much slower, and I couldn't think of another way to write the function that was more elegant, but still fast.

PR #493, opened 2016-10-11
Use printf in asciihex_to_binary_file

dcooper16:
Use `printf` in `asciihex_to_binary_file()` rather than `echo -e -n`


drwetter:
thank you, David

PR #494, opened 2016-10-11
Remove test of version tolerance

dcooper16:
PR #346 added a test for version tolerance to `run_protocols()`, but I think it may now be more appropriate to remove that test. [Draft -16 of TLS 1.3](https://tools.ietf.org/html/draft-ietf-tls-tls13-16), which was posted on September 22, changed the way that version negotiation is handled for TLS 1.3 and above. The current version tolerance test sends a ClientHello with the version field set to "03, 05", to represent a TLS 1.4 ClientHello. While this was consistent with [RFC 5246](https://www.rfc-editor.org/rfc/rfc5246.txt) and with drafts of TLS 1.3 up to -15, draft -16 changed the version field to `legacy_version` and declared that its value should be "03, 03" for TLS 1.2 and above. (For TLS 1.3 and above a Supported Versions extension is included to inform the server which versions of TLS the client supports.) The change in draft -16 was made as a result of the problems with servers not handling version negotiation correctly.

Since the current draft suggests that a server should never be presented with a ClientHello with a version higher than "03, 03" (even for clients that support TLS versions higher than 1.2), it seems there is no reason to include the version tolerance test anymore.

For servers that do not support TLS 1.2, the additional checks that were added by PR #346 will already detect if the server cannot perform version negotiation correctly, since it will display an error if the TLS 1.2 ClientHello doesn't result in a connection using the highest version of TLS that the server supports.


drwetter:
Seems logic consequence to me on the first glance.

But can't this wait a bit? I am hesitant because it is still a draft and the whole TLS 1.3 definition seems still in motion. Personally to start signaling higher TLS versions as 1.2 in a TLS extension sounds weird to me -- but to be honest I haven't followed any discussion.

Also for draft-16 PR #346  would be still a tolerance check. Question is only what it would be worth then (attention: a little bit of irony).

SSLlabs does that similar, only using different version numbers.

drwetter:
I would like to keep this open until things settle.

For now an idea could be to reflect the latest draft maybe by using instead of **Version tolerance** e.g. **Version tolerance (< draft-16)**.

BTW: Thanks very much for the fruitful discussion :-)

dcooper16:
It's okay with me if you don't want to merge this PR. I mainly submitted it now since version 2.8 isn't final yet, but will be soon. I've started to follow the IETF TLS WG mail list recently, and it seems that they are trying to wrap things up soon, so I would be surprised if they undid this change after all the discussion that just took place on the topic and the decision that there was consensus to make the change. There just seems to be a belief that there are fewer servers that have problems with unrecognized extensions than servers that don't handle version negotiation correctly.

So, while all existing servers **should** be able to pass the "Version tolerance" test, it seems likely that it won't matter in practice if servers fail when the ClientHello specifies a version higher than 1.2 (03, 03), so it's not clear that testssl.sh should flag it as a problem.

drwetter:
FYI: Probably I will make an official announcement soon about "stable" if I get some issues off the table. A few issues remain open which I make a 2.8-X later on. 2.9dev is already separated and will be the next default branch after 2.8-X.

Latest for 2.8-X probably #346 doesn't make so much sense, unless it's being considered as a some kind of tolerance test.

The more I read the draft I think I need to correct my initial impression.

dcooper16:
I was just reading through [draft-ietf-tls-tls13-21](https://datatracker.ietf.org/doc/draft-ietf-tls-tls13/) and noticed the following text that was added to draft-20 via a [PR](https://github.com/tlswg/tls13-spec/commit/5c09f1eb40354668a7594b6ed9df7a4b27729cbf):

> If [the supported versions] extension is not present, servers which are compliant with this specification MUST negotiate TLS 1.2 or prior as specified in [RFC5246], even if ClientHello.legacy_version is 0x0304 or later.  **Servers MAY abort the handshake upon receiving a ClientHello with legacy_version 0x0304 or later.**

So, when PR #346 was committed [draft-ietf-tls-tls13-14](https://tools.ietf.org/html/draft-ietf-tls-tls13-14), [RFC 5246](https://www.rfc-editor.org/rfc/rfc5246.txt), [RFC 4346](https://www.rfc-editor.org/rfc/rfc4346.txt), and [RFC 2246](https://www.rfc-editor.org/rfc/rfc2246.txt) all agreed that a server receiving a ClientHello with any version number higher than the highest version number supported by the server should return a ServerHello with a version number specifying the the highest version number supported by the server. So, the "Version tolerance" test was appropriate.

Now, [draft-ietf-tls-tls13-21](https://datatracker.ietf.org/doc/draft-ietf-tls-tls13/) explicitly permits a server receiving a version 0x0304 (or 0x0305) ClientHello to abort the handshake. So, a server could be compliant with [draft-ietf-tls-tls13-21](https://datatracker.ietf.org/doc/draft-ietf-tls-tls13/) and fail the "Version tolerance" test in testssl.sh version 2.8.

drwetter:
What? This is IMO going into the wrong direction security wise.

Actually I thought it would be a good idea to reimplement a tolerance test for 2.9dev.

drwetter:
maybe I was too fast, as it concerns a non-existing protocol.

drwetter:
I haven't follow the discussion there and I couldn’t find out. But did you got why this all of a sudden should be ok?

Aborting the connection has been the evil before and I think also if this is an exception for something which will not exist it seems to me like a lip service to the (formally) evil.

dcooper16:
I don't recall any discussion leading to the change specifically allowing servers to "abort the handshake upon receiving a ClientHello with legacy_version 0x0304 or later."

My recollection is that the decision to move to using the supported_versions extension was based on whole GREASE metaphor. Supposedly servers tended to be better about ignoring unknown extensions and unknown values in extensions than handling version negotiation as specified in TLSv1.2 and below. So, they thought that moving to a supported_versions extension would cause fewer problems with bad TLSv1.2 implementations than would sending a ClientHello that advertised {03, 04} in the version field.

While moving to supported_versions was a result of accommodating incorrectly implemented servers, I can't entirely fault the new language. There is a lot of language in the TLS standards saying that a server "MUST abort the handshake" if something is received from the client that no valid client should send. For example, servers are required to abort the handshake if the ClientHello message contains any data after the Extensions field (or the compression methods field, if there are no extensions).

drwetter:
> I don't recall any discussion leading to the change specifically allowing servers to "abort the handshake upon receiving a ClientHello with legacy_version 0x0304 or later."

IMO it sends the wrong signal -- perception could be that in general also for other, lower protocols it could be ok. And it just complicates matters...

> My recollection is that the decision to move to using the supported_versions extension was based on whole GREASE metaphor. Supposedly servers tended to be better about ignoring unknown extensions and unknown values in extensions than handling version negotiation as specified in TLSv1.2 and below. So, they thought that moving to a supported_versions extension would cause fewer problems with bad TLSv1.2 implementations than would sending a ClientHello that advertised {03, 04} in the version field.

My perception was that the broken (formally) intolerant servers were out there first. And people would have been afraid that the introduction to TLS 1.3 will take as long as TLS 1.2 if they use the record layer instead of supported_versions extension (TLS 1.2 was ready for 1-2 years in browsers but vendors like Mozilla were afraid tho switch them on because of (in other way) intolerant servers. )

I thought GREASE appeared a bit later then the draft with the supported_versions extension but I can be mistaken. You know probably better. Anyway I think it's a good thing.

As for the removal of version tolerance in the 2.8 stable branch -- I am hesitant. SSLlabs still has it and the TLS 1.3 draft is a draft. If something will change, I'll reconsider.

dcooper16:
> I thought GREASE appeared a bit later then the draft with the supported_versions extension but I can be mistaken. You know probably better. Anyway I think it's a good thing.

Actually, [draft-davidben-tls-grease-00](https://tools.ietf.org/html/draft-davidben-tls-grease-00) came out in July 2016 and the supported_versions extension wasn't added until September 2016, but I don't know when each of these ideas started to be discussed.

> As for the removal of version tolerance in the 2.8 stable branch -- I am hesitant. SSLlabs still has it and the TLS 1.3 draft is a draft. If something will change, I'll reconsider.

Understood. However, TLSv1.3 just underwent a [second working group last call](https://www.ietf.org/mail-archive/web/tls/current/msg23668.html) that ended yesterday, and that was limited to changes introduced to the document after draft 18. The TLS WG met earlier today (see https://www.ietf.org/proceedings/99/slides/slides-99-tls-sessb-tls-13-00.pdf and/or https://www.youtube.com/watch?v=fv1AIgRdKkY). There has been no discussions recently about version negotiation, so it seems safe to assume that this won't change.

Of course even without the specific "version tolerance" test, testssl.sh will still have some version tolerance testing. The current testssl.sh (2.8 and 2.9dev) will issue an error if a server that doesn't support TLSv1.2 can't handle a TLSv1.2 ClientHello. In the near future, 2.9dev will issue an error if a server that doesn't support TLSv1.3 can't handle a TLSv1.3 ClientHello. We can supplement that by adding a "version tolerance" test in which a TLSv1.3 ClientHello is sent in which the supported_versions extension includes additional versions (e.g., 0x0305).

dcooper16:
I recently came across an [article](https://lwn.net/Articles/701956/) that notes that the testssl.sh test for version intolerance is incomplete (as is the SSL Labs one). According to https://github.com/ssllabs/ssllabs-scan/issues/385 the problem is that the version tolerance test just checks whether the server responds with a ServerHello with an appropriate version number (rather than responding with an alert). However, with some version intolerant servers the failure is happening later:

> Handshake fails after client key exchange, change cipher spec stage.

PR #496, opened 2016-10-13
Increase HEADER_MAXSLEEP value to 10

teward:
There is a case where using this on a network with some (slight) instability or slowness can sometimes fail to detect or retrieve headers.  Currently, in my testing here, I've seen this range anywhere from 5.7 seconds to 7.8 seconds for a response from getting the headers, so maybe we should consider increasing the time slightly for the header max sleep value.  10 seconds is arbitrary, but still 'short' enough to die off quickly.

(Feel free to discuss or decline, but I'm hoping we can incorporate this to adapt for slightly slower networks)

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -162,7 +162,7 @@ SERVER_SIZE_LIMIT_BUG=false             # Some servers have either a ClientHello
 
 # tuning vars, can not be set by a cmd line switch
 EXPERIMENTAL=${EXPERIMENTAL:-false}
-HEADER_MAXSLEEP=${HEADER_MAXSLEEP:-5}   # we wait this long before killing the process to retrieve a service banner / http header
+HEADER_MAXSLEEP=${HEADER_MAXSLEEP:-10}   # we wait this long before killing the process to retrieve a service banner / http header
 readonly MAX_WAITSOCK=10                # waiting at max 10 seconds for socket reply
 readonly CCS_MAX_WAITSOCK=5             # for the two CCS payload (each)
 readonly HEARTBLEED_MAX_WAITSOCK=8      # for the heartbleed payload


drwetter:
Good question. 

Out of my guts: 5 secs should be enough for a HTTP server to respond. And as much as I love the idea that a tool does what the user wants (without doing too much assumptions) -- I believe a tool is a tool. And if hammer doesn't work as expected, the hammer can be held differently =>  providing the value from ENV.

OTOH there should be reasonable defaults. I am afraid that if we prolong this that could penalize those requests which will never come back. To be honest: I haven't done real profound data on this. So as a consequence: If I would have this (a lot of hosts, LAN, WAN, slow connections) I am happy to extend this period.

Opinions?

teward:
I agree, 5 seconds *should* be good for a response time, but depending on the congestion of a network, we may have it even *fractions* of a second longer.

In the last 50 tests I ran over the past hour since I submitted this, I'm getting most if not all on this network returning in 5.87.  If we do anything, I would suggest maybe just increasing by a second, to 6.  Depending on others' opinions.

(10 was pulled out of nowhere, but 6 could be acceptable and adapt for such slight slow response times)

drwetter:
Hi Thomas,

Am 10/13/2016 um 08:51 PM schrieb Thomas Ward:
> I agree, 5 seconds /should/ be good for a response time, but depending on the congestion of a
> network, we may have it even /fractions/ of a second longer.
> 
> In the last 50 tests I ran over the past hour since I submitted this,

if it was 50 tests over more then 1-2 host, would you mind putting this together as an
informal sheet?

> I'm getting most if not
> all on this network returning in 5.87. If we do anything, I would suggest maybe just increasing
> by a second, to 6. Depending on others' opinions.

(waiting for others to speak up)

BR, Dirk

teward:
I don't have the logs but I will try and replicate.  (We just had an incident that took more attention so i didn't get a chance to log all the results; I will run more tests soon to gauge the times to get headers based on the code being executed (I have a debug line locally giving me the full string to execute to pass to `time` on CLI))

teward:
Hm, looks like I know the reason for the delay, and it seems OpenSSL likes to behave slowly now compared to when it used to be faster.  Maybe I need to start looking there as the cause of the problem.

I'm going to close this for now; if the defaults need changed at a later time with a discussion, that'll be good.  (I'm also going to run some comparative tests from other systems on other nets to see if it's just this OpenSSL which I built from Peter's fork we use as a base for the testssl.sh binaries, or if it's all OpenSSLs, or if it's specific nets)

PR #498, opened 2016-10-19
Apple ATS9 client simulation

Sousaphone:
I had the need to test for Apple ATS9 compatibility last week, and found that only SSLLabs had the option. This client simulation has been created using the cipher/protocol spec from SSLLabs and the handshake bytes for that spec from the bundled openssl.


drwetter:
Thx, @Sousaphone .

There's one outstanding PR though from @seccubus which will cover that and much more too.

PR #499, opened 2016-10-25
Add option to retrieve entire server response

dcooper16:
In some cases the server's response to a ClientHello spans more than one packet. If the goal is just to determine whether the connection was successful and to extract a few pieces of information from the ServerHello message, then this is unlikely to be a problem. However, if there is a desire to extract the server's certificate chain (Certificate message) or to determine the type and size of the server's ephemeral public key (ServerKeyExchange message), then the entire response needs to be obtained, even if it spans multiple packets.

This PR adds a new function, `check_tls_serverhellodone()`, that checks whether the entire response has been received (e.g., whether the ServerHelloDone message has been received). If the response indicates that the response is incomplete, then `tls_sockets()` requests more data from the server until the response is complete or until the server doesn't provide any more data in response.

The PR only changes the behavior of `tls_sockets()` if the caller indicates that it wants to extract the ephemeral key or that it wants the entire response to be parsed. Otherwise, only the first packet returned by the server is sent to `parse_tls_serverhello()`. [The value of `$process_full` is not used at the moment, but will be in a subsequent PR that modifies `parse_tls_serverhello()`.]

This PR also changes `tls_sockets()` to send a close_notify to the server if the connection was successfully established.


testssl.sh
```diff
@@ -5761,10 +5761,98 @@ parse_sslv2_serverhello() {
...
+          remaining=$tls_hello_ascii_len-$i
+          [[ $msg_len -gt $remaining ]] && return 1
+
+          if [[ "$tls_content_type" == "16" ]]; then
+               tls_handshake_ascii+="${tls_hello_ascii:i:msg_len}"
+               tls_handshake_ascii_len=${#tls_handshake_ascii}
+               # the ServerHello MUST be the first handshake message
+               [[ $tls_handshake_ascii_len -ge 2 ]] && [[ "${tls_handshake_ascii:0:2}" != "02" ]] && return 2
+               if [[ $tls_handshake_ascii_len -ge 12 ]]; then
+                    DETECTED_TLS_VERSION="${tls_handshake_ascii:8:4}"
```
drwetter: Isn't that  pre-draft16 code where the TLS 1.3 is signaled in the Handshake layer, i.e. not extension?

dcooper16:
No, the ServerHello didn't change in draft-16, just the ClientHello. So, the client indicates the versions it supports in an extension, but the server still indicates the selected version in the first field of the ServerHello.

There are changes in the ServerHello in TLS 1.3, which will show up in `parse_tls_serverhello()`. The `random` is just 32 bytes of random data (no time field), the `session_id` and `compression_method` fields have been removed, and extensions are now mandatory.
```
 TLS 1.2                                           TLS 1.3
   struct {                                        struct {
       ProtocolVersion server_version;                 ProtocolVersion version;        
       Random random;                                  Random random;
       SessionID session_id;                           CipherSuite cipher_suite;
       CipherSuite cipher_suite;                       Extension extensions<0..2^16-1>;
       CompressionMethod compression_method;       } ServerHello;
       select (extensions_present) {
           case false:
               struct {};
           case true:
               Extension extensions<0..2^16-1>;
       };
      } ServerHello;

drwetter:
Thanks for the thorough explanation and your PR

PR #501, opened 2016-10-27
Fix for when no mapping file is present

dcooper16:
Now that the mapping file is no longer used, `$ADD_RFC_STR` should not be unset just because the mapping file cannot be found.

In addition, since `show_rfc_style()` is now used in `parse_tls_serverhello()`, it cannot return an empty string just because the user set "--mapping no-rfc" on the command line. Instead, `neat_list()` should check the value of `$ADD_RFC_STR` and not call `show_rfc_style()` if it has been unset.

Finally, since `show_rfc_style()` no longer returns strings with extra spaces, there is no need to call `strip_spaces()`

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1560,11 +1560,9 @@ rfc2openssl() {
 
 
 show_rfc_style(){
-     [[ -z "$ADD_RFC_STR" ]] && return 1
-     #[[ -z "$1" ]] && return 0
-
      local rfcname="" hexcode
      local -i i
+
      hexcode="$(toupper "$1")"
      case ${#hexcode} in
           3) hexcode="0x00,0x${hexcode:1:2}" ;;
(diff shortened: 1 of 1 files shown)


drwetter:
thx

drwetter:
Hi David,

can't find the right PR.... thus commenting here

 FYI: I need to swap our your mapping (currently https://github.com/drwetter/testssl.sh/blob/2.9dev/testssl.sh#L8515-L8873) to an external file. It's cleaner to separate code and data. Also the client simulation data will need to go to a different file.

Cheers, Dirk

PS: Do you have a twitter handle? So far I was only using your name...

dcooper16:
Okay, the current code dumps the mapping table to an external file and then reads the file to populate the arrays (TLS_CIPHER_*), so it should be fairly easy to just move the table to an external file. The only downside is that code would need to be reintroduced to handle that case in which the mapping file is missing.

No, I don't have a twitter handle. I don't have any social media accounts.

PR #506, opened 2016-11-03
Extend TLS ServerHello parsing (part 1)

dcooper16:
This PR extended `parse_tls_serverhello()` in a few ways:

* If the "full" response is to be parsed, then additional checks are performed to verify that `$tls_hello_ascii` contains the entire response

* The extensions field is parsed and the list of extensions found is placed in `$TLS_EXTENSIONS` (if the "full" response is being parsed).

* Initial support for TLS 1.3 is added:

  - Accounts for differences between TLS 1.2 ServerHello and TLS 1.3 ServerHello (as outlined in PR #499).

  - Recognizes new alerts and handshake message types.

  - Allows for server response to include message fragments of type "application data"

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -5851,24 +5851,26 @@ check_tls_serverhellodone() {
 }
 
 # arg1: ASCII-HEX encoded reply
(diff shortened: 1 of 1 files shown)


drwetter:
wow. Awesome!

PR #507, opened 2016-11-03
Allow caller to provide extra extensions to tls_sockets()

dcooper16:
This PR allows the caller to provide additional extensions to `tls_sockets()` to be included in the ClientHello. If the caller provides an extension that would have already been included in the ClientHello, then the caller's value for the extension is used rather than the default value.


drwetter:
Thank you

PR #508, opened 2016-11-03
Extend TLS ServerHello parsing (part 2)

dcooper16:
This PR adds initial parsing of the ServerKeyExchange message to `parse_tls_serverhello()`. For ephemeral DH keys, it extracts the length of the key. For ephemeral ECDH keys that are encoded using the named_curve option, it extracts the length of the key and the name of the curve.


drwetter:
Great. Thank you, David

PR #511, opened 2016-11-04
Use ephemeral public key in client simulations

dcooper16:
This PR makes similar changes to `run_client_simulation()` as were made to `tls_sockets()`, so that  `run_client_simulation()` retrieves the entire server response, even if it is split across multiple packets, and it has `parse_tls_serverhello()` extract information about the server's ephemeral public key, if present.

The PR also changes `run_client_simulation()` to use information about the ephemeral public key. It includes the length of the public key in the output and, if it is a DH public key, checks that the size is within the acceptable range (`${minDhBits[i]} <= dh_bits <= ${maxDhBits[i]}`).

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2348,7 +2348,8 @@ client_simulation_sockets() {
      local -i save=0
      local lines clienthello data=""
      local cipher_list_2send
-     local tls_hello_ascii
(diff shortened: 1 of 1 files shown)


drwetter:
Thx again!

PR #512, opened 2016-11-04
Use tls_sockets() in run_allciphers()

dcooper16:
The PR changes `run_allciphers()` to use `tls_sockets()` (and `sslv2_sockets()`)rather than `$OPENSSL` unless `$SSL_NATIVE` is set or `$STARTTLS` is non-empty. Using sockets allows `run_allciphers()` to test all ciphers, rather than just those supported by `$OPENSSL`.

Using sockets results in `run_allciphers()` running more slowly, partially since it is testing more ciphers, but mostly since `tls_sockets()` is currently slower than `$OPENSSL` (as noted in #413).

(Note that I broke this PR into two separate commits: the first just fixes some indentation errors in `run_allciphers()` without changing the code and the second contains the actual code changes.)

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2051,70 +2051,70 @@ run_allciphers() {
      done
 
      for (( bundle_size/=4; bundle_size>=1; bundle_size/=4 )); do
-         # Note that since the number of ciphers isn't a power of 4, the number
-         # of bundles may be may be less than 4**(round_num+1), and the final
-         # bundle may have fewer than bundle_size ciphers.
-         num_bundles=$nr_ciphers/$bundle_size
-         mod_check=$nr_ciphers%$bundle_size
-         [[ $mod_check -ne 0 ]] && num_bundles=$num_bundles+1
-         for ((i=0;i<num_bundles;i++)); do
-             # parent=index of bundle from previous round that includes this bundle of ciphers
-             parent=4**$round_num+$i/4
-             # child=index for this bundle of ciphers
-             child=4*4**$round_num+$i
-             if ${ciphers_found[parent]}; then
-                 ciphers_to_test=""
-                 end_of_bundle=$i*$bundle_size+$bundle_size
-                 [[ $end_of_bundle -gt $nr_ciphers ]] && end_of_bundle=$nr_ciphers
-                 for ((j=i*bundle_size;j<end_of_bundle;j++)); do
-                     [[ "${sslvers[j]}" != "SSLv2" ]] && ciphers_to_test="${ciphers_to_test}:${ciph[j]}"
-                 done
-                 ciphers_found[child]=false
-                 if [[ -n "${ciphers_to_test:1}" ]]; then
-                      $OPENSSL s_client -cipher "${ciphers_to_test:1}" $STARTTLS $BUGS -connect $NODEIP:$PORT $PROXY $SNI >$TMPFILE 2>$ERRFILE </dev/null
-                      sclient_connect_successful "$?" "$TMPFILE"
-                      [[ "$?" -eq 0 ]] && ciphers_found[child]=true
-                 fi
-             else
-                 # No need to test, since test of parent demonstrated none of these ciphers work.
-                 ciphers_found[child]=false
-             fi
+          # Note that since the number of ciphers isn't a power of 4, the number
+          # of bundles may be may be less than 4**(round_num+1), and the final
+          # bundle may have fewer than bundle_size ciphers.
+          num_bundles=$nr_ciphers/$bundle_size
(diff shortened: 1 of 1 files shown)


drwetter:
That PR is absolutely cool and the right thing to do as it gives on certain hosts more ciphers in return (for google even two post-quantum ciphers!).

As you mentioned the catch it is way slower (short tests with 2-4 tries per value):

```
Google:     3.3s / 1.0s
Facebook:   3.4s / 1.3s
Cloudflare: 2.3s / 1.0s
LAN:        4.5s / 1.4s
```

The # of ciphers is factor 2 (openssl vs. sockets). Except cloudflare though the speed penalty it is factor 3. So using sockets is still 50% slower.

I am afraid if there will be even more tests via sockets the whole thing will slow down considerably. Have you tried anything yet?

I haven't really tried hard yet. The things I tried weren't successful yet. Probably some profiling will help to get an idea ...

drwetter:
> The PR changes run_allciphers() to use tls_sockets() [..] unless $SSL_NATIVE is set or $STARTTLS is non-empty

Any reason for the latter?

dcooper16:
> > The PR changes run_allciphers() to use tls_sockets() [..] unless $SSL_NATIVE is set or $STARTTLS is non-empty
> 
> Any reason for the latter?

I just followed the logic from `run_protocols()`. I did the same thing in `run_client_simulation()`.

I haven't made much of an effort to speed up `tls_sockets()` either. You suggested in #413 that it "is probably due to the polling in `sockread_serverhello()`." So, I was hoping that someone who understood that part of the code better than I do might be able to work on it.

In the meantime, would you like to hold off on adding PRs like this one to 2.9dev, or perhaps add them, but only use the sockets implementation if the `$EXPERIMENTAL` flag is set?

drwetter:
>>    Any reason for the latter?
>
> I just followed the logic from run_protocols(). I did the same thing in run_client_simulation().

Oh, yes (cough). ;-)

Sockets work transparently over STARTTLS. There was a situation though where sockets or 
plain openssl mattered for ``run_protocols()``. It is somewhere documented here at github.

For ``run_client_simulation()`` though the logic seems to be the opposite.(``"$SSL_NATIVE" && using_sockets=false``), it's in a pending branch (client_simulation) though which I still need to merge.

> I haven't made much of an effort to speed up tls_sockets() either. You suggested in #413 that it "is probably due to the polling in sockread_serverhello()." So, I was hoping that someone who understood that part of the code better than I do might be able to work on it.

:-) hehe. Didn't do that either in the past, found some interesting links though which I wasn't able to follow up upon yet:

http://stackoverflow.com/questions/4336035/performance-profiling-tools-for-shell-scripts
http://stackoverflow.com/questions/5014823/how-to-profile-a-bash-shell-script-slow-startup

Maybe if somebody's listening it is a starting point.

> In the meantime, would you like to hold off on adding PRs like this one to 2.9dev, or perhaps add them, but only use the sockets implementation if the $EXPERIMENTAL flag is set?

Depends.... whether the PR using sockets provide additional use  -- in terms of results -- for the user. 

Here it does -- which is good -- however for the **now(!)** standard apache/nginx/IIS server it practically penalizes other users. As you submitted it I thought this would be a case for the EXPERIMENTAL flag  But then one still get's the wrong result e,g, on google when it's nobody using it. 

Maybe until this performance problem is solved we introduce another flag like "--thorough" and document this in ``help()``. Or the opposite thing like "--fast" which also spares other checks (no 2nd to N cipher in preferences e.g.)

I need to sleep over it myself. What's your stance?

Cheers, Dirk

dcooper16:
It seems like the path of least resistance would be to either not merge the PR at the moment or to add it, but only use the sockets version if the `$EXPERIMENTAL` flag is set.

This is in the 2.9dev branch and the 2.8 branch hasn't even been finalized yet, so there should be plenty of time to decide how to deal with this in the long term. Maybe before a decision needs to be made someone will figure out how to make `tls_sockets()` run faster.

drwetter:
This is great work. It will be merged.

I am only afraid if I merge it without any flags or kludges, it will be forgotten to improve ``tls_sockets()`` or whatever the culprit is.

drwetter:
Thanks again, David.

I settled for the --fast flag see d4ed7466ce4aff0682d616d0f117182cd22aad03. 

-E still now uses openssl only. :-(

dcooper16:
Thanks. I've been giving some thought to ways to speed up `run_all_ciphers()`. I think I have an idea that will work, but it involves rewriting most of the function, so it will need a lot of testing. I'll submit a PR as soon as I believe that it's ready.

PR #513, opened 2016-11-08
Curve X25519 fixes

dcooper16:
This PR fixes two issues related to curve X25519.

First, while OpenSSL 1.1.0 supports curve X25519, it is not included in the output of `$OPENSSL ecparam -list_curves`. I tried several versions of OpenSSL (and one version of LibreSSL), and every version outputs either "Error with command" or "unknown option" in response to `$OPENSSL s_client -curves $curve` if it either does not support the `-curves` option or does not support `$curve`. (When the `-curve` option is supported with `$curve`, a "connect" error is output.)

The second issue is that the "Server Temp Key" line in the output of `s_client` is different for curve X25519. For other elliptic curves, the output is
```
Server Temp Key: ECDH, P-256, 256 bits
```
For X25519 it is:
```
Server Temp Key: X25519, 253 bits
```
So, `read_dhbits_from_file()` needs to allow for `$what_dh` being "X25519" rather than "ECDH" and `run_pfs()` needs to allow for the possibility that the curve name will be the first field rather than the second.  (In `read_dhbits_from_file()`, I change `$what_dh` from "X25519" to "ECDH", but an alternative would be to change the code later in `read_dhbits_from_file()` to output `pr_done_good "253 bit X25519"` or `pr_done_good "253"`.)

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3503,6 +3503,10 @@ read_dhbits_from_file() {
      grep -q bits <<< $bits || bits=$(awk -F',' '{ print $2 }' <<< $temp)
      bits=$(tr -d ' bits' <<< $bits)
 
+     if [[ "$what_dh" == "X25519" ]] || [[ "$what_dh" == "X448" ]]; then
+          what_dh="ECDH"
+     fi
+
      debugme echo ">$HAS_DH_BITS|$what_dh|$bits<"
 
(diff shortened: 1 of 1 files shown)


drwetter:
Thx for the fix and extensive explanation

PR #514, opened 2016-11-08
Curve X25519 fixes for 2.9dev

dcooper16:
This PR fixes the same issues as were fixed in PR #513, but also makes two changes to `parse_tls_serverhello()`:

* It changes the number of bits for curve X25519 from 256 to 253 to match OpenSSL.

* It removes the "ECDH, " from the "Server Temp Key: " line in order to match OpenSSL's output.


drwetter:
thx again

PR #515, opened 2016-11-08
Find more extensions in run_server_defaults()

dcooper16:
This PR uses `tls_sockets()` to determine whether a server supports certain extensions that may not be supported by `$OPENSSL`. At the moment it checks for max_fragment_length, client_certificate_url, truncated_hmac, ALPN, signed_certificate_timestamp, encrypt_then_mac, and extended_master_secret.

In https://github.com/dcooper16/testssl.sh/blob/extended_tls_sockets/testssl.sh, `run_server_defaults()` is re-written to use `tls_sockets()` instead of `$OPENSSL`, with just one call to `$OPENSSL s_client` to get the session ticket, which reduces the dependence on `$OPENSSL`, but this PR limits the number of calls to `tls_sockets()`, which is still slow.

Note: I included ALPN in the `tls_sockets()` ClientHello since a single call to `tls_sockets()` cannot test for both NPN and ALPN, and since support for NPN was added to OpenSSL before support for ALPN was added, I figured it was more likely that `determine_tls_extensions()` had already determined whether the server supported NPN.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -839,6 +839,23 @@ asciihex_to_binary_file(){
      return 0
 }
 
+# arg1: text string
+# Output a comma-separated ASCII-HEX string resprestation of the input string.
+string_to_asciihex() {
+     local string="$1"
+     local -i i eos
+     local output=""
+
+     eos=${#string}-1
+     for (( i=0; i<eos; i++ )); do
+          output+="$(printf "%02x," "'${string:i:1}")"
+     done
+     [[ -n "$string" ]] && output+="$(printf "%02x" "'${string:eos:1}")"
+     out "$output"
+     return 0
+     
+}
+
 ###### check code starts here ######
 
 # determines whether the port has an HTTP service running or not (plain TLS, no STARTTLS)
@@ -5135,6 +5152,8 @@ run_server_defaults() {
      local -a ocsp_response ocsp_response_status sni_used
      local -a ciphers_to_test success
      local cn_nosni cn_sni sans_nosni sans_sni san
+     local alpn_proto alpn="" alpn_list_len_hex alpn_extn_len_hex success
+     local -i alpn_list_len alpn_extn_len
 
      # Try each public key type once:
      # ciphers_to_test[1]: cipher suites using certificates with RSA signature public keys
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks.

See our discussion in #512. Need to do some tests for any performance impact.

Then please note that with openssl I have added tests for both ALPN and NPN (``testssl.sh -S testssl.sh``)

```

 TLS extensions (standard)    "server name/#0" "renegotiation info/#65281" "EC point formats/#11" "session ticket/#35" "status request/#5" "heartbeat/#15" "application layer protocol negotiation/#16" "next protocol/#13172"
````

 Takes two calls.

Probably it needs also some rearragements. The original function for this is ``determine_tls_extensions()``.

**Comment by dcooper16** [comment]
I think this PR is a little different than #512. Using sockets for `run_allciphers()` involves many calls to `tls_sockets()`; testing google.com involves 109 calls! This PR involves just one call to `tls_sockets()`, so it creates **much** less of a performance impact.

At the moment, `determine_tls_extensions()` is called more than 7 times, since it is trying to find all of the server's certificates. On a site like facebook.com, which has both an RSA and ECC certificate, `determine_tls_extensions()` is called 9 times. Since I just wanted to find additional TLS extensions by making a single call to `tls_sockets()` (with the ClientHello listing all cipher suites), I placed the call in `run_server_defaults()`.

I realize that the current testssl.sh will test for both the NPN and ALPN extensions, if the version of OpenSSL being used supports those extensions. However, there are other extensions that this PR will detect. For example, using the OpenSSL from the testssl.sh website, the current testssl.sh finds the following extensions for google.com:
```
 TLS extensions (standard)    "renegotiation info/#65281" "server name/#0" "session ticket/#35" "application layer protocol negotiation/#16" "EC point formats/#11" "next protocol/#13172"
```
However, using this PR, the list of extensions detected is:
```
TLS extensions (standard)    "renegotiation info/#65281" "server name/#0" "session ticket/#35" "application layer protocol negotiation/#16" "EC point formats/#11" "next protocol/#13172" "extended master secret/#23" "signed certificate timestamps/#18"
```
So, the PR finds two extensions (extended master secret and signed certificate timestamps) that the current testssl.sh doesn't. With the PR, all eight of these extensions are found, even if an older version of OpenSSL that doesn't support ALPN is used.

Even using OpenSSL 1.1.0 with the current testssl.sh doesn't find the signed certificate timestamps extension.

**Comment by drwetter** [comment]
Ok, got it now. Thanks.

My bigger picture is that I want in the future to avoid N calls from ``run_whatever()`` to ``subfunction()``. 
Instead of it there should be a global status variable which keeps the result of a subfunction like 

``
[[ -z "$TLS_EXTENSIONS" ]] && determine_tls_extensions
``

or 

```
if [[ ! -s $HEADERFILE ]]; then
          run_http_header "$1" || return 3
     fi
```

so that the caller just checks whether there's a result in the needed global variable and if not: populate this variable. If yes:  just use it.

It avoids redundant connect calls and makes it easier readable. 

So for your code code here maybe a better place is determine_tls_extensions.

Issue #520, opened 2016-11-15
False positive with OpenSSL version < 1.0


dcooper16:
Since `CIPHERS_BY_STRENGTH_FILE` includes the OpenSSL names for the cipher suites, I was looking into whether testssl.sh could do more with old versions of OpenSSL that do not support the `-V` option for `$OPENSSL ciphers`. However, I ran into a problem, as shown below. With OpenSSL 0.9.8zh (the version that comes with OS X), `$OPENSSL s_client` seems to return `0` even if the connection is not successful. As a result a call to `$OPENSSL s_client` followed by `sclient_connect_successful()` indicates that the connection was successful, even if it wasn't.

Perhaps this isn't worth fixing, since the use of such an old version of OpenSSL is strongly discouraged, but I wanted to at least raise the issue.

```
bash-3.2$ openssl version
OpenSSL 0.9.8zh 14 Jan 2016
bash-3.2$ openssl s_client -cipher RC4-MD5 -connect 81.169.199.25:443
CONNECTED(00000003)
47879:error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure:/BuildRoot/Library/Caches/com.apple.xbs/Sources/OpenSSL098/OpenSSL098-59.60.1/src/ssl/s23_clnt.c:593:
bash-3.2$ echo $?
0
```

drwetter:
Yes, that's breaking it and other places, too. ``-V`` is not suppported either.

0.9.8<whatever> is out of support, at least from OpenSSL. Don't know what Apple users are using it nowadays. It basically boils down to this question whether it's worth considering it.

PR #521, opened 2016-11-15
OpenSSL 0.9.8 does not support the pkey command

dcooper16:
The `tls_sockets()` and `sslv2_sockets()` use `get_pub_key_size()` to extract the size of the server's public key if the full response is being processed, and `get_pub_key_size()` uses `$OPENSSL pkey` to extract the server's public key from the certificate. However, OpenSSL 0.9.8 does not support the "pkey" command. This PR changes `get_pub_key_size()` to suppress the error message displayed by OpenSSL when the "pkey" command is not supported.


drwetter:
Thanks.

Well, ok. See #520 .

PR #522, opened 2016-11-15
Faster version of run_allciphers() 

dcooper16:
This PR speeds up the implementation of `run_allciphers()` by introducing a number of changes:

* Rather than check for implemented ciphers in a hierarchical manner (as introduced in #326), this PR follows the approach of `cipher_pref_check()`. Testing a block of ciphers, marking the selected cipher as implemented, and then testing same block of ciphers, minus those that have previously been selected, until a test fails. Thus the number of calls to `$OPENSSL s_client` is just one more than the number of ciphers implemented. (Since some servers cannot handle ClientHellos with more than 128 messages, the tests are performed on blocks of 128 or fewer ciphers. So, if OpenSSL supports 197 ciphers, the number of calls to `$OPENSSL s_client` is 2 plus the number of ciphers supported by the server.

* If `$using_sockets` is true, then OpenSSL is used first to find all supported ciphers that OpenSSL supports (since OpenSSL is faster than `tls_sockets()`), and then `tls_sockets()` is only used to test those cipher suites that were not found to be supported by OpenSSL.

* The `prepare_debug()` function, which reads in the mapping file determines which ciphers are supported by the version of OpenSSL being used. If a version of OpenSSL older than 1.0 is being used, then this is used to determine which ciphers to test using OpenSSL rather than using `$OPENSSL ciphers -V`.

Following the approach of `cipher_pref_check()` reduces the number of queries to the server. Using OpenSSL before `tls_sockets()` reduces the number of calls to `tls_sockets()` to 3 plus the number of ciphers supported by the server that are not supported by OpenSSL, so the cost penalty over just using OpenSSL is fairly small.

If this PR is accepted, then a similar change can be made to `run_cipher_per_proto()`.


drwetter:
Sounds cool, thanks.

Gimme a bit to recover and to look at it.

Cheers, Dirk 
-- 

drwetter:
Excellent job! Thank you

|   Host           | PR 522   | PR 512  | openssl |
|-----------------|---------|-------------|------------|
|   tesstsl.net    |   6   |  14    | 5.4 |
|   dev.tetsssl.sh |  8.0  | 24     |9 |
|   google.de      |   11.5|  33    | 9 |
|   LAN            |  5.8  | 45.5   |4.4 |
|   facebook       |  9    | 33.3   |13 |

PR #523, opened 2016-11-15
Additional fixes for no cipher mapping file

dcooper16:
Client simulations can still use sockets even if the cipher mapping file is missing. If the cipher file is present, then `parse_tls_serverhello()` write the RFC name for the cipher and then `run_client_simulation()` converts that to the OpenSSL name (so that the output is the same as if OpenSSL were used).  This PR changes `parse_tls_serverhello()` so that it writes the OpenSSL name for the cipher if the mapping file is missing, which `run_client_simulation()` can then just display.

This PR also unsets `ADD_RFC_STR` if the mapping file is missing, so that `neat_list()` won't try to display the RFC names for the ciphers.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3317,13 +3317,10 @@ run_client_simulation() {
      requiresSha2+=(false)
 
      outln
-     [[ $TLS_NR_CIPHERS == 0 ]] && using_sockets=false
      if "$using_sockets"; then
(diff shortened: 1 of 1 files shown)


drwetter:
Thank you for your improvement to my fix!

The dh_bits are still not shown, maybe because of #531.

Q: can you improve ``run_allciphers()`` in the same manner?

PR #525, opened 2016-11-16
fix a duplicate test id in the client sim section

tecknicaltom:
The Apple ATS 9 iOS 9 client simulation test had an incorrect short name
which resulted in a duplicate ID in the JSON output


drwetter:
Thx for catching this

PR #528, opened 2016-11-18
Extend TLS ServerHello parsing (part 5)

dcooper16:
When `tls_sockets()` is used with the "full" option and the chosen cipher suite involves an ephemeral finite-field DH key (DH), this PR extracts the public key from the ServerKeyExchange message and adds it to `$TMPFILE`. In addition (and the primary reason for this PR), it compares the ephemeral public key's parameters to those specified in [RFC 7919](https://www.rfc-editor.org/rfc/rfc7919.txt), and indicates whether one the groups from that RFC was used. This will allow `run_pfs()` to be extended to indicate which, if any, [RFC 7919](https://www.rfc-editor.org/rfc/rfc7919.txt) DH groups a server supports.

This PR extracts the [RFC 7919](https://www.rfc-editor.org/rfc/rfc7919.txt) DH group parameters from the key pairs in `$TLS13_KEY_SHARES`. The primary purpose for `$TLS13_KEY_SHARES is in support of TLS 1.3 (as can be seen in https://github.com/dcooper16/testssl.sh/blob/extended_tls_sockets/testssl.sh). In TLS 1.3, the ClientHello must include a key_shares extension with the client's ephemeral public keys and the corresponding private keys are needed to decrypt parts of the server's response.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -343,6 +343,202 @@ c0,0d, c0,03, 00,0a, 00,63, 00,15, 00,12, 00,0f, 00,0c,
 00,62, 00,09, 00,65, 00,64, 00,14, 00,11, 00,0e, 00,0b,
 00,08, 00,06, 00,03, 00,ff"
 
+readonly -a TLS13_KEY_SHARES=(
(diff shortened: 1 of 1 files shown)


drwetter:
Good work, thank you!

PR #529, opened 2016-11-18
Extend TLS ServerHello parsing (part 6)

dcooper16:
This PR extends `tls_sockets()` by parsing the contents of the ALPN and NPN extensions, if the "full" response is being processed.


drwetter:
Thx again

PR #530, opened 2016-11-19
Revert "Fix neat_list() for Camellia GCM"

drwetter:
Reverts drwetter/testssl.sh#524


drwetter:
Thanks for the PR and noticing.

I was too fast and reverted it for now (fix follows later) It's kind of painful adding (2) more columns but as we have AES and AESGCM and do not have the MAC somewhere it would be more correct to have also CamelliaGCM

Issue #531, opened 2016-11-19
fix run_client_simulation()


drwetter:
```
 "$using_sockets" && [[ -n "${handshakebytes[i]}" ]] && has_dh_bits=$HAS_DH_BITS && HAS_DH_BITS=true
"$HAS_DH_BITS" && read_dhbits_from_file $TMPFILE
 "$using_sockets" && [[ -n "${handshakebytes[i]}" ]] && HAS_DH_BITS=$has_dh_bits
```

drwetter:
should be resolved

PR #533, opened 2016-11-21
More fixes for missing cipher mapping file

dcooper16:
> The dh_bits are still not shown, maybe because of #531.

This PR fixes the issue of dh_bits not being shown if the cipher-mapping.txt file is missing. The problem is that the code in `parse_tls_serverhello()` that parses the ServerKeyExchange message assumes that `$rfc_cipher_suite` has the RFC version of the name the cipher suite. However, if the cipher-mapping.txt file is missing, `$rfc_cipher_suite` will have the OpenSSL name of the cipher suite. This PR changes the code to recognize either the RFC or OpenSSL names for ciphers with ephemeral DH or ECDH keys.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7102,7 +7102,8 @@ parse_tls_serverhello() {
 
      # Now parse the server key exchange message
      if [[ $tls_serverkeyexchange_ascii_len -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


drwetter:
Thank you

PR #534, opened 2016-11-21
un-disable SNI for STARTTLS protocol connections

typingArtist:
SNI works for STARTTLS protocols just as for standard
TLS protocols like HTTPS. There is no reason to disable
it.

I am actually in the process of setting up such a service and would like to test it with testssl.sh.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1063,7 +1063,6 @@ service_detection() {
      local addcmd=""
 
      if ! $CLIENT_AUTH; then
-          # SNI is nonsense for !HTTPS but fortunately for other protocols s_client doesn't seem to care
           [[ ! "$1" =~ ssl ]] && addcmd="$SNI"
           printf "$GET_REQ11" | $OPENSSL s_client $1 -quiet $BUGS -connect $NODEIP:$PORT $PROXY $addcmd >$TMPFILE 2>$ERRFILE &
           wait_kill $! $HEADER_MAXSLEEP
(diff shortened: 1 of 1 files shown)


typingArtist:
@drwetter, any chance to have this merged? Do you have any explicit objections? Anything else you want to discuss or have documented for that change?

drwetter:
Thx, @typingArtist . 

I researched it a bit ~3 weeks ago. As a matter of fact, yes, some servers support SNI over STARTTLS, like exim and dovecot, postfix (until recently?) and ejabbered e.g. do not support it. The majority of server applications does not IMO. And I couldn't find an official internet standard for it. 

OTOH if it allows more servers to be tested correctly I would be fine with it. At the moment I don't know about side effects (needs testing). And: Looking at it from the user perspective I am not sure whether it is the right thing to do as they might confuse it with other options (xmpp domain) or whether it causes wrong expectations --> from the client you can't tell whether there's SNI support on the server side. So there's some uncertainty interpreting results correctly.

I would like to have this PR sitting here for a bit until higher priority issues/PRs are resolved and then I'll take care about it.

typingArtist:
Thanks for looking into it. I don’t think that SNI in STARTTLS requires any additional standard to work or to be allowed.TLS means: Connect, then do TLS handshake (with or without SNI). START

drwetter:
Cleaning up. That (and more) was done in #1280

PR #535, opened 2016-11-21
Reorganize run_server_defaults()

dcooper16:
This function reorganizes `run_server_defaults()` based on the suggestion in #515.

The current `determine_tls_extensions()` is renamed to `get_server_certificate()`, and two changes are made to it:
* it no longer includes an extra call to `$OPENSSL s_client` to check for the ALPN extension; and
* rather than setting `$TLS_EXTENSIONS` to be the extensions found during this call to the function, it adds any newly found extensions to those already in `$TLS_EXTENSIONS`.

The PR then adds a new function, `determine_tls_extensions()`, which borrows some logic from the old `determine_tls_extensions()`, but this new `determine_tls_extensions()` only looks for additional TLS extensions, including ALPN.

`run_server_defaults()` makes multiple calls to `get_server_certificate()` (as it previously did to `determine_tls_extensions()`) in order to collect all of the server's certificates, and then it makes one call to `determine_tls_extensions()`, which checks for support for extensions that were not checked for by `get_server_certificate()` (e.g., ALPN and, if sockets are being used, extensions not supported by `$OPENSSL`).

The new `determine_tls_extensions()` will check for most of the extensions that are checked for by 
`run_server_defaults()`, including the heartbeat extension, so the call to `determine_tls_extensions()` from `run_heartbleed()` will still work.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4691,17 +4691,89 @@ sclient_connect_successful() {
      return 1
 }
 
+# Note that since, at the moment, this function is only called by run_server_defaults()
+# and run_heartbleed(), this function does not look for the status request or NPN
+# extensions. For run_heartbleed(), only the heartbeat extension needs to be detected.
+# For run_server_defaults(), the status request and NPN would already be detected by
+# get_server_certificate(), if they are supported. In the case of the status extension,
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks again for this

Issue #519, opened 2016-11-15
Speed up sockets


drwetter:
Currently and as more and more socket-based tests will be implemented it is recommended that some profiling and speed improvements will be done (see also discussion under #512).

drwetter:
...copying over from #512 the hints:

http://stackoverflow.com/questions/4336035/performance-profiling-tools-for-shell-scripts
http://stackoverflow.com/questions/5014823/how-to-profile-a-bash-shell-script-slow-startup

drwetter:
Note: ``read -t `` might be at least more elegant, see #536

drwetter:
As a first start I committed 27d0570fb5f15e1636e571b166d9879a62e15e82 which should help here ...

drwetter:
see also #667

drwetter:
Heads up: I've changed this week the socket output to a fifo instead of a file which gave a good performance gain, up to 20% in LAN+WAN,  Google and Cloudflare less (~10 /~5%) for some reason. Testing testssl.sh was 56s/70s (Linux client).

Fifo is a good compromise as it doesn't do read and writes to disk. All file actions taking place are in-memory -- besides creating the fifo in the filesystem and using dd. One cannot use variables as they have a problem with NULL chars

It's not committed yet, as there need to be more changes and testing needs to be done.

PR #536, opened 2016-11-22
Implement proper STARTTLS handshake for ftp, smtp, pop, imap and nntp

typingArtist:
The current logic doesn’t work with some SMTP servers which are picky about protocol order, so I always had to go with `FAST_STARTTLS=false`. Additionally the current code already tries to work around some of the problems instead of just performing STARTTLS properly, which is actually not as hard as it seems – at least for the ftp, smtp, pop and imap, as well as the partially-supported nntp.

Still not converted to a proper STARTTLS dialog are partially-supported ldap and xmpp. ldap is a beast on its own, and xmpp requires XML parsing for compliance. I left them at the bad state for now.

It would be cool if openssl had support for picking up a live session in a fd, and starting TLS from where it was left. This way, testssl.sh could implement STARTTLS, and openssl could take over for protocols unknown to openssl. However, we won’t see that anytime soon – at least openssl 1.1 introduces unix domain socket access.

Tested smtp, imap and pop3 against some of my servers extensively (smtp is meta1, imap is cyrus-imapd, pop3 is dovecot). Tested nntp against public service nntp.aioe.org:119 and ftp against test.rebex.net:21

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -853,6 +853,7 @@ fileout() { # ID, SEVERITY, FINDING
 
 debugme() {
      [[ "$DEBUG" -ge 2 ]] && "$@"
+     return 0
 }
 
 hex2dec() {
--- a/testssl.sh
+++ b/testssl.sh
@@ -6073,8 +6073,8 @@ starttls_line() {
 }
 
 starttls_just_send(){
-     debugme echo -e "\n=== sending \"$1\" ..."
-     echo -e "$1" >&5
+     debugme echo -e "C: $1"
+     echo -ne "$1\r\n" >&5
 }
 
 starttls_just_read(){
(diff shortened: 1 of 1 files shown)


drwetter:
The point of the existing, deliberately not clean, but most of the time good working handshake is that it is fast.

A STARTTLS handshake alone is overhead. If you do a proper handshake you loose a considerable amount of more time. 

Note for every single test of testssl you have to restart the handshake. 







Am 22. November 2016 19:11:54 MEZ, schrieb typingArtist <notifications@github.com>:
>The current logic doesn’t work with some SMTP servers which are picky
>about protocol order, so I always had to go with `FAST_STARTTLS=false`.
>Additionally the current code already tries to work around some of the
>problems instead of just performing STARTTLS properly, which is
>actually not as hard as it seems – at least for the ftp, smtp, pop and
>imap, as well as the partially-supported nntp.
>
>Still not converted to a proper STARTTLS dialog are partially-supported
>ldap and xmpp. ldap is a beast on its own, and xmpp requires XML
>parsing for compliance. I left them at the bad state for now.
>
>It would be cool if openssl had support for picking up a live session
>in a fd, and starting TLS from where it was left. This way, testssl.sh
>could implement STARTTLS, and openssl could take over for protocols
>unknown to openssl.
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/drwetter/testssl.sh/pull/536
>
>-- Commit Summary --
>
>* Implemented proper STARTTLS handshake for ftp, smtp, pop, imap and
>nntp
>
>-- File Changes --
>
>    M testssl.sh (131)
>
>-- Patch Links --
>
>https://github.com/drwetter/testssl.sh/pull/536.patch
>https://github.com/drwetter/testssl.sh/pull/536.diff


-- 

typingArtist:
My patch improves the handshake in comparison to your dirty one in three ways, and speed is even among them:

* correct STARTTLS order of operations client/server
* analyzes FEAT/EHLO/CAPABILITY/CAPABILITIES for proper handshake – there is room for improvements on the  user interface, however
* doesn’t sleep, so it is done when the server is done (typically way faster than the 1 second of your final `starttls_line()`)

So all these things are improvements with exactly which downside? Where is the advantage of the dirty implementation?

Please also remember that already quite some tests are done through `openssl s_client` and not `fd_socket()`. openssl actually _does_ go through the full STARTTLS handshake each time, and it’s not slow either –granted, it’s a little faster anyway as it’s written in C.

drwetter:
Okay, fair enough. Let me look into it.

Thx, Dirk

Am 23. November 2016 06:39:26 MEZ, schrieb typingArtist <notifications@github.com>:
>My patch improves the handshake in comparison to your dirty one in
>three ways, and speed is even among them:
>
>* correct STARTTLS order of operations client/server
>* analyzes FEAT/EHLO/CAPABILITY/CAPABILITIES for proper handshake –
>there is room for improvements on the  user interface, however
>* doesn’t sleep, so it is done when the server is done (typically way
>faster than the 1 second of your final `starttls_line()`)
>
>So all these things are improvements with exactly which downside? Where
>is the advantage of the dirty implementation?
>
>Please also remember that already quite some tests are done through
>`openssl s_client` and not `fd_socket()`. openssl actually _does_ go
>through the full STARTTLS handshake each time, and it’s not slow either
>–granted, it’s a little faster anyway as it’s written in C.


-- 

typingArtist:
Thanks. Actually there was a small bug in the script which only shows up when the script is run without `--debug=`, so please look at the current version.

(please also consider #534).

drwetter:
Hi @typingArtist,

thanks a bunch. It's definitely faster (still trying to find the reason for). Given this and avoiding the dirty shortcut it's a good step forward!

Only thing: what if the loop ``while read -r -t $STARTTLS_SLEEP one_line; do`` blocks?

Cheers, Dirk

drwetter:
Hmm, I assumed ``wait_kill`` won't actually wait if the socket reply from the server comes back immediately. But maybe this is part of this and other speed problem.

typingArtist:
The code is quite broken without the patch. It’s running like that:

```
connect
send bli
send bla
send STARTTLS
cat >tmp & # will receive initial greeting, reaction to bli, reaction to bla, reaction to STARTTLS
wait_kill cat after 1 second
```

All the ‘benefit’ of sending without listening is gone when you `cat` and `wait_kill` for 1 second. Every connection will have to take this one second, because without proper protocol analysis you cannot arbitrarily break the `cat`, so you have to wait speculatively for 1 second, hoping that all of the server’s response has ben caught by `cat` until then. Leaving some reaction unread in the socket can break the following TLS handshake, so we should always swallow what was received.

Doing the protocol correctly doesn’t require any unneeded waiting, and killing. Just wait (or timeout) for the next line from the server, it will come, analyze it and continue with the protocol. In the end, no waiting for 1 second is necessary.

Why should `while read -r -t $STARTTLS_SLEEP one_line; do` block? It reads a line, or waits for one line, or waits for a socket error, or waits for the now misnamed sleep time (actually, it’s a timeout). A read error or timeout will automatically terminate `read` and `while`, as will the analysis inside the `while` do if the result is not an expected continuation, e.g. starting with `250-` or if it’s the final, expected result, e.g. starting with `250[space]`. Additionally, the loop can wait for an expected property, e.g. in the list of features returned for `EHLO`, it waits for `STARTTLS`. This way I can easily follow the protocol by checking the availability of STARTTLS and stop immediately if there is something wrong, not overrunning the server with improper commands.

I would actually want to rename `$STARTTLS_SLEEP` to `$STARTTLS_SOCKET_TIMEOUT` and raise its value to something larger, as it will not influence normal quick server handshakes. OTOH we could better communicate with tar pits. We might even consider backing off to single-byte reads after hitting a protocol line timeout, with smaller timeouts between these byte reads for better testing of tar pits that do make progress, but just on a single-byte basis. But let’s first get this change done :grin:

drwetter:
> Why should while read -r -t $STARTTLS_SLEEP one_line; do block?

I meant the while loop which is getting input from the file descriptor. It waits for a server reply. If the server just doesn't reply, testssl hangs. Admittedly that isn't often the case but if it happens it's a problem.

Because I had sometimes to deal with hanging replies, this is why I put the read in the background and polled the read process. Except cleaning up all the legacy STARTTLS handshakes -- which is appreciated -- your suggested implementation doesn't address this.

The whole thing is related to #519 (that is originally meant without upgrading from plaintext to TLS). I spend some time improving this but w/o any success so far. The amount of sleep btw doesn't seem to be relevant, also sleeping a fraction of a second doesn't make it faster.

The order of commands you mentioned is not quite correct though but it doesn't help for this discussion.

typingArtist:
The `read` gets input from the socket (or it does not). If it gets sufficient input within the timeout to return a full line, it writes to the variable `$one_line` and returns zero, so the `while` stays true. If it could not get enough data to form a full line in time (note the `-t $STARTTLS_SLEEP`), or the socket is closed or otherwise erroneous, `read` will return with an error so that `while` immediately terminates. Otherwise the commands inside the loop are executed, analyzing the returned line in `$one_line`, eventually returning on either a completely unexpected line or the actual end of a return sequence.

The `debugme` calls should kind of document what leads to the various exit reasons and `return` statements.

And, no, this cannot hang. `read` will return after one second if called with `-t 1`. The logic *does* address this. Have you actually seen that it breaks? I would be really surprised.

A larger timeout only helps testing slow servers, be it we have a very bad connection to them, or on purpose being a tar pit. As soon as my logic runs into a timeout, the connection breaks. It must never run into a timeout for running successful.

drwetter:
sorry, my bad. Should get more sleep.

Thanks for the cleanup/speedup PR, much appreciated.

PR #537, opened 2016-11-23
don't parse invalid server reply for SSLv2

typingArtist:
SSLv2 is dying. However, we test against it. Let’s not waste precious cycles on parsing an already invalid response against 300+ possible ciphers.

Note there is another block of code around [L6390](https://github.com/drwetter/testssl.sh/compare/2.9dev...typingArtist:parsing_sslv2_shortcut?expand=1#diff-2b772c01e86977452d2db64b3031f913R6392) that could be guarded as well … what do you think @drwetter?

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6349,6 +6349,7 @@ parse_sslv2_serverhello() {
      # [cipher spec length] ==> ciphers GOOD: HERE ARE ALL CIPHERS ALREADY!
 
      local ret=3
+     local invalid=0
      if [[ "$2" == "true" ]]; then
           echo "======================================" > $TMPFILE
      fi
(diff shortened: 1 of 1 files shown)


typingArtist:
btw, with this patch applied, `testssl -e` on one of my IMAP servers with 6 ciphers enabled (and SSLv2 disabled, of course) runs in 10.9 seconds compared to 40.8 s without the patch.

btw, this patch plus #536 makes a total run time of 4.9 seconds. :grinning:

typingArtist:
@drwetter, is there anything missing in this patch you want to see?

dcooper16:
Hi @typingArtist,

I agree that `parse_sslv2_serverhello()` shouldn't be trying to extract the certificate or the list of ciphers from the server's reply if the reply isn't valid. However, I believe that should apply whether the return value for `parse_sslv2_serverhello()` is 0, 1, or 7, not just if it is 1. So, I'd like to propose a slightly different change to the code to address this issue. Please look at PR #551 and let me know if you are okay with it.

Thanks,

David

typingArtist:
@dcooper16, thanks for your PR #551. Please go on with that.

PR #540, opened 2016-11-29
add Android 7.0 handshake

typingArtist:
This is from a dump of K-9 Mail running on Nougat 7.0, patch level 5 Nov 2016, said to be the same as FolderSync running on the same device. It appears that [Android 7.0 has lost all but one elliptic curve](http://stackoverflow.com/questions/39133437/sslhandshakeexception-handshake-failed-on-android-n-7-0), so I thought it would be cool to test for that.

@drwetter, it appears we have to decide about [the same problem as SSL Labs](https://community.qualys.com/thread/16833-testing-android-x-vs-chrome-on-android-x) regarding naming our tests appropriately and differentiating between Android x capabilities and that of Chrome on Android x. This patch is clearly identifying Android 7.0 (not Chrome).


drwetter:
Hi @typingArtist,

thx. There's a script though in a pending branch still to be merged, see ``client-simulation-data.sh`` in ``https://github.com/drwetter/testssl.sh/tree/client_simulation/utils`` which pulls the API from SSLlabs.

Cheers, Dirk

PS: cool, still RC4 ;-)

typingArtist:
Well, my addition is based on a real connection of a client app, not Chrome, it is grepped from a tcpdump.

What’s currently on SSL Labs named "Android x" is only Chrome for Android x, not Android x, the system stack itself. See my question [here](https://community.qualys.com/thread/16833-testing-android-x-vs-chrome-on-android-x). I think that we should ask ourselves the same question, what we really want to test and how to name it.

drwetter:
@typingArtist : was your q answered? 

Point is that the API from SSLlabs gives us all the data via their API https://api.dev.ssllabs.com/api/v3/getClients  -- that belongs to the merge I have on my table. I don't want to change that offering from SSLlabs.

Add more clients manuall? Maybe, but not now. Also I think it's smarter to wait whether your q will be addressed by SSLlabs.

Did you do a cross check with @LeeBrotherston's tls-fingerprinting?

typingArtist:
no, not answered yet. Seems nobody cares about that differentiation. Yet. The differentiation between Safari and the Apple stack has not been in SSL Labs from the beginning, either.

I am completely fine if you go with the SSL Labs API. Actually I added that just for my testing against a friend’s service because he owns Android 7.0 devices he couldn’t make work with it. Only after days I thought about making that a PR.

Perhaps we can create an issue where we can summarize new clients and even support for STARTTLS client checks.

Thanks for the pointer to tls-fingerprinting. That’s new for me, a quick crosscheck doesn’t reveal a match, but I have to look more closely.

drwetter:
@typingArtist : admittedly this section needs a helping hand -- I wasn't able yet to merge the branch "client_simulation" in yet.

I would have loved to merge yours before but then if I look at the data collected it seems not correct. I mentioned SSLlabs before and their API -- I also do not believe that Android 7 had no old CHACHA/POLY ciphers (cc13-15). Then Google states that ARCFOUR has been removed: https://developer.android.com/about/versions/nougat/android-7.0-changes.html#tls-ssl -- unless one uses the ``SSLSocketFactory()``. So maybe the _real connection of a client app_ as your phrased it is doing this or similar.

SSLlabs is supposed to use the standard browser .

typingArtist:
I raised that on SSL Labs forum [once more](https://community.qualys.com/thread/17423-rant-why-is-there-no-real-android-70-testing). The situation was marginally affecting me having only one friend of mine with a Google phone by that time, which is upgraded to 7.1.x for quite some time, but now we have two Huawei Mate 9’s in the family, and they are proudly running Android 7.0 with no estimate for an OS upgrade. Really annoying. Please, consider adding that additional test, even without SSL Labs enhancing their client connection tests towards the Android TLS library.

drwetter:
We can always add stuff. Have you tried _utils/wireshark2ciphers.pl_  ?

The change you are asking for may sound small as it is only for Android 7 but it seems a different approach -- which is not something which I want to waive but needs some consideration upfront. 

For now there are browsers in the handshake and universally used TLS-stacks (openssl, java).  We can change that. My pain is that the list in a default run  shouldn't be too big as is eats time and every tiny little bit of added information is probably not what everybody wants, considering the costs mentioned. 

The I have the concern that somebody needs to maintain this. Dumping once in a while the API from SSLlabs is straightforward. Gathering specific fingerprints and maintain them is not as easy.

But I see now @ivanr answered. You might want to checkout https://github.com/drwetter/testssl.sh/blob/2.9dev/utils/wireshark2ciphers.pl . Also @LeeBrotherston's [tls-fingerprinting](https://github.com/LeeBrotherston/tls-fingerprinting) might come in handy

typingArtist:
No, haven’t tested wireshark2ciphers.pl. Have to take a look at it.

I think at least we should support two sets of clients: those we import from SSL Labs, and those on top of it, maintained by us. Perhaps we can make some of them optional in some way, or we can allow the user to specify a file like with --clienthello=android7.0-hello.bin which shall be tested against this server.

I am fully with you regarding run-time for the client tests. However, Android 7.0 is big enough a client to warrant this test to be done, I think.

Let’s hope the issue gets some traction at SSL Labs … so we just import their client simulation fixing it for us as well. ;)

drwetter:
@typingArtist : I am closing this only for house keeping reasons. There needs to be a general approach. 

Feel free to open an issue regarding this general approach for client handshaking. Maybe somebody is willing to step in and maintain such data.

PR #541, opened 2016-11-29
run_cipher_per_proto() speedup + sockets

dcooper16:
This PR implements `run_cipher_per_proto()` in a manner similar to `run_allciphers()`.


drwetter:
Wow! google.com is 3x faster now

drwetter:
Ah. Maybe you want to have a look and let me know what you think...

If I run this off a different directory where the cipher mapping file cannot be found it is faster. I assume by looking at the result this is because it has less ciphers to test -- I was the culprit which put the cipher list to an external file. ;-)

As testssl.sh gets bigger and bigger IMO it becomes more important to at least separate code from data (there's more to do) But I guess the warning in the beginning ``No cipher mapping file in $TESTSSL_INSTALL_DIR/etc/ found`` is flying by too fast. It also doesn't inform about  consequences

Opinions?

PR #542, opened 2016-11-29
run_rc4() sockets implementation

dcooper16:
This PR implements `run_rc4()` in a similar manner to `run_allciphers()` and `run_cipher_per_proto()` (in PR #541). The change doesn't seem to have much of an impact on speed, but when sockets are used it can detect ciphers that aren't locally supported by OpenSSL.


drwetter:
Thank you.

PR #551, opened 2016-12-06
Don't parse SSLv2 ServerHello unless successful response

dcooper16:
This PR is a proposed alternative to #537. It only attempts to extract the certificate and list of ciphers from the SSLv2 ServerHello is `ret=3`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6514,9 +6514,9 @@ parse_sslv2_serverhello() {
           fi
      fi
 
-     [[ "$2" == "true" ]] && [[ -e $HOSTCERT ]] && rm $HOSTCERT
-     [[ "$2" == "true" ]] && [[ -e $TEMPDIR/intermediatecerts.pem ]] && rm $TEMPDIR/intermediatecerts.pem
-     if [[ "$2" == "true" ]] && [[ $ret -eq 3 ]]; then
+     [[ "$2" == "true" ]] || return $ret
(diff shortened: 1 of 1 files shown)


typingArtist:
I like this one, thanks @dcooper16. perhaps you could refine `[[ -e file ]] && rm file` into `rm -f file` instead and tie the two `rm`’s into one?

Additionally I’d make a `[[ "$2" == "true" ]] || return $ret` before the changed block so $2 doesn’t have to be evaluated three times in a row.

Then the code would read something like:

```
[[ "$2" == "true" ]] || return $ret
rm -f $HOSTCERT $TEMPDIR/intermediatecerts.pem
if [[ $ret -eq 3 ]]; then
  …
fi
return $ret
```
```

drwetter:
> ``[[ "$2" == "true" ]] || ``

is way too long. ``"$2" ||`` suffices

dcooper16:
Thanks @typingArtist , I made the changes that you suggested.

> >    [[ "$2" == "true" ]] ||
> 
> is too long. "$2" || suffices. Same for &&

I tried making this change, but it didn't work. It seems that for this to work "$2" would have to always be either `true` or `false`, but in this case it will likely be either `true` or the empty string. If "$2" is the empty string, then `"$2" ||` results in a "command not found" error.

drwetter:
Thx @dcooper16 and also @typingArtist for discussing and implementing

PR #553, opened 2016-12-06
Fix bug in reading of cipher mapping file

dcooper16:
When the cipher-mapping.txt file is read, the contents of the "Mac=..." column is placed in `TLS_CIPHER_EXPORT` rather than the contents of the "export" column. This PR fixes that.


drwetter:
Thank you, David.

PR #555, opened 2016-12-08
Align elliptic curve lengths with OpenSSL

dcooper16:
For several elliptic curves the number of bits, as indicated by OpenSSL, is slightly different than the name implies. For example, for sect239k1 OpenSSL outputs: `Server Temp Key: ECDH, sect239k1, 238 bits`.

This PR aligns the output created by `parse_tls_serverhello()` with OpenSSL.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7281,25 +7281,25 @@ parse_tls_serverhello() {
                     # http://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
                     case $named_curve in
                          1) dh_bits=163 ; named_curve_str="K-163" ;;
-                         2) dh_bits=163 ; named_curve_str="sect163r1" ;;
+                         2) dh_bits=162 ; named_curve_str="sect163r1" ;;
                          3) dh_bits=163 ; named_curve_str="B-163" ;;
                          4) dh_bits=193 ; named_curve_str="sect193r1" ;;
                          5) dh_bits=193 ; named_curve_str="sect193r2" ;;
-                         6) dh_bits=233 ; named_curve_str="K-233" ;;
+                         6) dh_bits=232 ; named_curve_str="K-233" ;;
                          7) dh_bits=233 ; named_curve_str="B-233" ;;
-                         8) dh_bits=239 ; named_curve_str="sect239k1" ;;
-                         9) dh_bits=283 ; named_curve_str="K-283" ;;
-                         10) dh_bits=283 ; named_curve_str="B-283" ;;
-                         11) dh_bits=409 ; named_curve_str="K-409" ;;
+                         8) dh_bits=238 ; named_curve_str="sect239k1" ;;
+                         9) dh_bits=281 ; named_curve_str="K-283" ;;
+                         10) dh_bits=282 ; named_curve_str="B-283" ;;
+                         11) dh_bits=407 ; named_curve_str="K-409" ;;
                          12) dh_bits=409 ; named_curve_str="B-409" ;;
-                         13) dh_bits=571 ; named_curve_str="K-571" ;;
-                         14) dh_bits=571 ; named_curve_str="B-571" ;;
-                         15) dh_bits=160 ; named_curve_str="secp160k1" ;;
-                         16) dh_bits=160 ; named_curve_str="secp160r1" ;;
-                         17) dh_bits=160 ; named_curve_str="secp160r2" ;;
+                         13) dh_bits=570 ; named_curve_str="K-571" ;;
+                         14) dh_bits=570 ; named_curve_str="B-571" ;;
+                         15) dh_bits=161 ; named_curve_str="secp160k1" ;;
+                         16) dh_bits=161 ; named_curve_str="secp160r1" ;;
+                         17) dh_bits=161 ; named_curve_str="secp160r2" ;;
                          18) dh_bits=192 ; named_curve_str="secp192k1" ;;
                          19) dh_bits=192 ; named_curve_str="P-192" ;;
-                         20) dh_bits=224 ; named_curve_str="secp224k1" ;;
+                         20) dh_bits=225 ; named_curve_str="secp224k1" ;;
                          21) dh_bits=224 ; named_curve_str="P-224" ;;
                          22) dh_bits=256 ; named_curve_str="secp256k1" ;;
                          23) dh_bits=256 ; named_curve_str="P-256" ;;


drwetter:
Not sure OpenSSL is right, see http://www.secg.org/sec2-v2.pdf.

Where is that in OpenSSL defined. A quick grep didn't return much

dcooper16:
I tried finding the information in the OpenSSL source code, but couldn't. The way I found the information was to add a line to the part of `run_pfs()` that uses OpenSSL to find supported curves:
```
grep "Server Temp Key" "$TMPFILE"
```

I then ran it using a couple of different versions of OpenSSL against some servers that support a lot of different curves. Between fpki-graph.fpki-lab.gov, www.nginx.com, and www.google.com I was able to get all of the curves defined in NamedCurve except x448, which isn't yet supported by OpenSSL.

I don't know from where OpenSSL got its numbers. So, it doesn't really matter to me whether you close or merge this PR. None of the values differ by more than 2 bits, which is insignificant. However, without this PR, testssl.sh will display slightly different results depending on whether the results are obtained using OpenSSL or `tls_sockets()`.

For fpki-graph.fpki-lab.gov the results  of the `grep` line were:
```
Server Temp Key: ECDH, K-163, 163 bits
Server Temp Key: ECDH, sect163r1, 162 bits
Server Temp Key: ECDH, B-163, 163 bits
Server Temp Key: ECDH, sect193r1, 193 bits
Server Temp Key: ECDH, sect193r2, 193 bits
Server Temp Key: ECDH, K-233, 232 bits
Server Temp Key: ECDH, B-233, 233 bits
Server Temp Key: ECDH, sect239k1, 238 bits
Server Temp Key: ECDH, K-283, 281 bits
Server Temp Key: ECDH, B-283, 282 bits
Server Temp Key: ECDH, K-409, 407 bits
Server Temp Key: ECDH, B-409, 409 bits
Server Temp Key: ECDH, K-571, 570 bits
Server Temp Key: ECDH, B-571, 570 bits
Server Temp Key: ECDH, secp160k1, 161 bits
Server Temp Key: ECDH, secp160r1, 161 bits
Server Temp Key: ECDH, secp160r2, 161 bits
Server Temp Key: ECDH, secp192k1, 192 bits
Server Temp Key: ECDH, P-192, 192 bits
Server Temp Key: ECDH, secp224k1, 225 bits
Server Temp Key: ECDH, P-224, 224 bits
Server Temp Key: ECDH, secp256k1, 256 bits
Server Temp Key: ECDH, P-256, 256 bits
Server Temp Key: ECDH, P-384, 384 bits
Server Temp Key: ECDH, P-521, 521 bits
```

For www.nginx.com the results were:
```
Server Temp Key: ECDH, P-256, 256 bits
Server Temp Key: ECDH, P-521, 521 bits
Server Temp Key: ECDH, brainpoolP512r1, 512 bits
Server Temp Key: ECDH, brainpoolP384r1, 384 bits
Server Temp Key: ECDH, P-384, 384 bits
Server Temp Key: ECDH, brainpoolP256r1, 256 bits
Server Temp Key: ECDH, secp256k1, 256 bits
Server Temp Key: ECDH, B-571, 570 bits
Server Temp Key: ECDH, K-571, 570 bits
Server Temp Key: ECDH, K-409, 407 bits
Server Temp Key: ECDH, B-409, 409 bits
Server Temp Key: ECDH, K-283, 281 bits
Server Temp Key: ECDH, B-283, 282 bits
```

and for www.google.com (using OpenSSL 1.1.0b):
```
Server Temp Key: X25519, 253 bits
Server Temp Key: ECDH, P-256, 256 bits
```

drwetter:
@dcooper16, thx. 

Gotta run, thus in short: 

The right approach IMO is first to find out what's going on and then determine the right definition. :-)

In ``apps/s_cb.c`` is ``ssl_print_tmp_key()`` which calls ``EVP_PKEY_bits`` in ``crypto/evp/p_lib.c`` and the value in the server hello stems from ``pkey->ameth->pkey_bits(pkey)``

drwetter:
@jurajsomorovsky , master of all elliptic curves, ;-) : you know by any chance how the -- on the first glance -- inconsistency of curve name and bit size comes from? "bits" above are the values which the serverhello returns, e.g. ``openssl s_client -curves B-571  -connect www.nginx.com:443 | grep 'Server Temp Key'``?

jurajsomorovsky:
We briefly took a look at it (with @neverhub), and we think the problem is that the OpenSSL server strictly counts the number of bits in the integer. There are two cases:
- The value is smaller than the elliptic curve size: if the server sends a key, which starts with 0-bits, OpenSSL removes the first 0-bits and prints out the number of bits counting from the first 1-bit.
- The value is larger than the elliptic curve size: in your trace we can see the 160-bit curve, where OpenSSL reports 161 bits. I believe (but I am not sure) that this is connected with some signing bit, which is used by the conversion of the byte array to a different OpenSSL structure.

jurajsomorovsky:
Btw, to view the EC bytes, you can also use TLS-Attacker:
`java -jar TLS-Attacker-1.2.jar client -connect  fpki-graph.fpki-lab.gov:443 -cipher TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA -named_curve SECT571R1`
or
`java -jar TLS-Attacker-1.2.jar client -connect  fpki-graph.fpki-lab.gov:443 -cipher TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA -named_curve SECP160R1`

You will get something like this in the log:
`17:27:53.285 [main] DEBUG de.rub.nds.tlsattacker.tls.protocol.handshake.ECDHEServerKeyExchangeHandler - Parsed the following EC domain parameters: 
17:27:53.285 [main] DEBUG de.rub.nds.tlsattacker.tls.protocol.handshake.ECDHEServerKeyExchangeHandler -   Curve order: 1461501637330902918203687197606826779884643492439
17:27:53.285 [main] DEBUG de.rub.nds.tlsattacker.tls.protocol.handshake.ECDHEServerKeyExchangeHandler -   Parameter A: ffffffffffffffffffffffffffffffff7ffffffc
17:27:53.286 [main] DEBUG de.rub.nds.tlsattacker.tls.protocol.handshake.ECDHEServerKeyExchangeHandler -   Parameter B: 1c97befc54bd7a8b65acf89f81d4d4adc565fa45
17:27:53.286 [main] DEBUG de.rub.nds.tlsattacker.tls.protocol.handshake.ECDHEServerKeyExchangeHandler -   Base point: (4a96b5688ef573284664698968c38bb913cbfc82,23a628553168947d59dcc912042351377ac5fb32,1) 
17:27:53.286 [main] DEBUG de.rub.nds.tlsattacker.tls.protocol.handshake.ECDHEServerKeyExchangeHandler -   Public key point Q: (86fbe32b25b57ab5a8db2a96a3e277dd7b81044d,6d09f25391694b6178f64dcc03aeb96703b00b48,1) 
17:27:53.287 [main] DEBUG de.rub.nds.tlsattacker.tls.workflow.GenericWorkflowExecutor - The following message was parsed: 
SERVER_KEY_EXCHANGE
  Handshake Message Length: 305
  Curve Type: NAMED_CURVE
  Named Curve: SECP160R1
  Public Key: 
  04 86 FB E3 2B 25 B5 7A  B5 A8 DB 2A 96 A3 E2 77
  DD 7B 81 04 4D 6D 09 F2  53 91 69 4B 61 78 F6 4D 
  CC 03 AE B9 67 03 B0 0B  48 
  Signature Algorithm: SHA512 RSA`

neverhub:
Well i guess we were wrong. I think the output does merely print the strength of the corresponding private value of received public point. The output thus equals the size/order n of the used curve. While the order of the curve is close to their corresponding field, they are not necessarily equal ([Hasse's theorem](http://en.wikipedia.org/wiki/Hasse%27s_theorem_on_elliptic_curves)). 

Let me give 2 quick examples:

1. `Server Temp Key: ECDH, K-283, 281 bits`
K-283 has order `0x1ffffffffffffffffffffffffffffffffffe9ae2ed07577265dff7f94451e061e163c61`
size of the curve => 281 bits
2. `Server Temp Key: ECDH, secp160r1, 161 bits`
secp160r1 has order `0x100000000000000000001F4C8F927AED3CA752257`
size of the curve => 161 bits

Correct me if i'm wrong, but that should explain both cases.

drwetter:
@neverhub & @jurajsomorovsky Thanks!

drwetter:
And: Thank you, David!

Issue #548, opened 2016-12-04
Mark weak curves in PFS section as weak


drwetter:

drwetter:
solved see PR

Issue #464, opened 2016-09-07
NamedCurves with ffdhe


drwetter:
Copy/cited from #122

--snip
On a related note, I just noticed that draft-ietf-tls-negotiated-ff-dhe extends NamedCurve in TLSv1.0 - TLSv1.2 to allow finite field Diffie-Hellman groups to be listed (as the draft of TLS v1.3 does for TLSv1.3:

```
       enum {
       // other already defined elliptic curves (see RFC 4492)
           ffdhe2048(256), ffdhe3072(257), ffdhe4096(258),
           ffdhe6144(259), ffdhe8192(260),
       //
       } NamedCurve;
```

Should I go ahead and add these to the Supported Groups (formerly Supported Elliptic Curves) extension, in the case that the client is offering any TLS_DHE_... ciphers, or wait until draft-ietf-tls-negotiated-ff-dhe becomes an RFC?
--snap

--> has been released, see https://tools.ietf.org/html/rfc7919. Can't tell who's using it yet. @dcooper16 ?

Topic for 2.9dev.

dcooper16:
I made some changes to https://github.com/dcooper16/testssl.sh/blob/extended_tls_sockets/testssl.sh. If a server chooses a TLS_DHE_* cipher suite, then `parse_tls_serverhello()` will extract the server's ephemeral public key from the ServerKeyExchange message and compare the public key's parameters to the parameters specified in RFC 7919.

I then added code to `run_pfs()` that (for servers that support TLS_DHE_* cipher suites) would send ClientHello messages with just TLS_DHE_* cipher suites and with supported_groups extensions that specified the FFDHE groups from RFC 7919.

None of the servers that I tested used any of groups from RFC 7919, and they all seem to just ignore the presence of the supported_groups extension when using a TLS_DHE_* cipher suite.

So, perhaps we shouldn't add the FFDHE groups from RFC 7919 to the ClientHello as a general rule. But, maybe in the future we can add something like the code that I added to `run_pfs()` that will, if the server supports RFC 7919 (or TLS 1.3), list the groups from RFC 7919 that it supports.

drwetter:
Thanks for the heads up, appreciated!

I guess we just need to watch when servers will start using it. Maybe a separate check at a certain time will make sense.

PR #556, opened 2016-12-08
run_pfs() speedup + sockets

dcooper16:
This PR implements `run_pfs()` in a manner similar to `run_allciphers()`. It uses OpenSSL followed by `tls_sockets()` to test for both supported PFS cipher suites as well as elliptic curves offered.

I made an attempt at addressing #548 by using different colors to print the different curve names, depending on strength. The colors chosen are exactly the same as those that would be chosen by `read_dhbits_from_file()`:
```
     # bits <= 163:       pr_svrty_medium
     163 < # bits <= 193: pr_svrty_minor
     193 < # bits <= 224: out
     # bits > 224:        pr_done_good
```

I also added code for #464 to create a list of the DH groups from RFC 7919 that a server supports. However, since no servers seem to support this at the moment (except with TLS 1.3), I marked this code to only run if the $EXPERIMENTAL flag is set.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -5781,38 +5781,98 @@ run_server_defaults() {
 
 run_pfs() {
      local -i sclient_success
-     local pfs_offered=false ecdhe_offered=false
-     local tmpfile
-     local dhlen
-     local hexcode dash pfs_cipher sslvers kx auth enc mac curve
-     local pfs_cipher_list="$ROBUST_PFS_CIPHERS"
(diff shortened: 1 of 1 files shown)


drwetter:
Thx! Bear w me, next week I'll be back
-- 
Sent from my mobile. Excuse the autocorrection, my brevity&typos

drwetter:
awesome, thx!

PR #565, opened 2016-12-19
Use sockets in run_server_preference()

dcooper16:
This PR adds the use of sockets to `run_server_preference()` to determine the "Negotiated cipher per proto." It only uses sockets in two cases:
* For SSLv2, if $OPENSSL does not support SSLv2.
* For SSLv3, if $OPENSSL does not support SSLv3.

This PR will have no effect if the provided OpenSSL binaries are used.

Note that this may create some "odd" results since `tls_sockets()` may choose a different cipher for SSLv3 than $OPENSSL does for TLSv1 and TLSv1.1, even though the server likely offers the same set of ciphers for SSLv3, TLSv1, and TLSv1.1. This seems to be unavoidable, given that different versions of OpenSSL will negotiate different ciphers.


drwetter:
Thank you.

How about >= TLS 1.0? The output is sometimes a bit odd too see ``./testssl.sh -E -P -q --ip=one testssl.net`` or ``./testssl.sh -E -P -q --ip=one google.com`` as there are the new (RFC) CHACHA/POLY ciphers which the shipped openssl doesn't support.

dcooper16:
I don't understand your comment on this one. The code in this PR only runs if the server does not specify a cipher order, and both testssl.net and google.com specify a cipher order.

This PR doesn't do anything for >= TLS 1.0, since I assume that whatever version of OpenSSL is being used it supports, TLSv1.0, TLSv1.1, and TLSv1.2, and this PR does nothing unless the server doesn't support one of the protocols (SSLv2 and/or SSLv3).

Were you confusing this PR with the one for `cipher_pref_check()`?

drwetter:
yes, sorry, forget about my comment. ;-)

PR #567, opened 2016-12-20
Mark export ciphers in run_rc4()

dcooper16:
This PR adds ",exp" to the bits column when `run_rc4()` is run in the "--wide" mode and the cipher is an export cipher. This makes the wide mode of `run_rc4()` align with other functions, such as `run_allciphers()`.


drwetter:
thank you!

PR #570, opened 2016-12-21
Use sockets for run_ssl_poodle()

dcooper16:
This PR changes `run_ssl_poodle()` to use sockets. This PR is particularly useful when $OPENSSL is OpenSSL 1.1.0, since OpenSS 1.1.0 does not support SSLv3 by default. But, it is also useful if $OPENSSL supports some, but not all, of the CBC ciphers.

As with `run_beast()`, there is a small change to `$cbc_ciphers`. The following two ciphers were added:
```
          0x00,0x0B - EXP-DH-DSS-DES-CBC-SHA  SSLv3 Kx=DH/DSS   Au=DH   Enc=DES(40)   Mac=SHA1 export
          0x00,0x0E - EXP-DH-RSA-DES-CBC-SHA  SSLv3 Kx=DH/RSA   Au=DH   Enc=DES(40)   Mac=SHA1 export
```
The ciphers that were removed are all SSLv2 ciphers:
```
     0x07,0x00,0xC0 - DES-CBC3-MD5            SSLv2 Kx=RSA      Au=RSA  Enc=3DES(168) Mac=MD5 
     0x06,0x00,0x40 - DES-CBC-MD5             SSLv2 Kx=RSA      Au=RSA  Enc=DES(56)   Mac=MD5 
     0x04,0x00,0x80 - EXP-RC2-CBC-MD5         SSLv2 Kx=RSA(512) Au=RSA  Enc=RC2(40)   Mac=MD5  export
     0x05,0x00,0x80 - IDEA-CBC-MD5            SSLv2 Kx=RSA      Au=RSA  Enc=IDEA(128) Mac=MD5 
     0x03,0x00,0x80 - RC2-CBC-MD5             SSLv2 Kx=RSA      Au=RSA  Enc=RC2(128)  Mac=MD5
```
(EXP-RC2-CBC-MD5 is both an SSLv2 and an SSLv3 cipher. Previously it was listed twice in `$cbc_ciphers`, now it appears once.)


drwetter:
Hi David,

for those and the other PRs regarding sockets you need to be a bit of patient with me. I need to do some testing for speed and depended on the outcome I will at the moment reuse the experimental flag for this. 

Unfortunately I need to dedicate time for this -- which until Xmas (here: 24-27, so 2 days longer as in the US) is a rare currency for me. I don't want to get you frustrated but if you have other PRs for me I am happy to resolve them sooner.

Cheers, Dirk

dcooper16:
Hi Dirk,

It's no problem. I'm trying to finish work on adding the use of sockets wherever $OPENSSL is likely to be lacking, but it's fine with me if the PRs don't get accepted for a while. Enjoy the holidays.

I understand the concern about speed, and based on your feedback I've tried to write functions like `run_allciphers()`, `run_cipher_per_proto()`, etc. to use $OPENSSL as much as possible, and then only use `tls_sockets()` to check whether $OPENSSL missed anything. So, the speed penalty over using `--ssl-native` tends to be fairly small, unless `tls_sockets()` finds something that would have been missed by $OPENSSL. There is also now the `--fast` option, allowing the user to choose speed over completeness.

It seems that without sockets the testing will become less and less complete. The provided OpenSSL binaries are great for testing whether servers support old, weak ciphers or protocols, but as was noted in #379, they don't support some of the newer ciphers that are increasingly being used by servers. On the other hand, the latest OpenSSL supports these newer ciphers, but has dropped support for older, weaker ciphers and protocols, making OpenSSL 1.1.0 inappropriate for verifying that servers don't support these things.

drwetter:
Hi David,

thank you!

Besides testing vulnerable and not vulnerable hosts I do measurements for your PRs. This one is ~0,2s slower using sockets which is ok. 

Just to give you my picture:

1. tests should be complete regarding ciphers (maybe really seldom used ciphers could be exempt (PSK, KRB5, SRP))
2. it should be as fast as possible
3. if there's a choice which is not good or the test is not complete, it should definitely echo a warning
4. a nice to have is if testssl is smart enough to decide on it's own what the best approach is (it shouldn't try to be smarter than the user. testssl is a tool and not supposed to take away control)
5. people are supposed to use the openssl version shipped with testssl.

1 and 2 are sometimes contradicting. 3 I implemented for poodle right now, older editions of run_XXX have this in an older edition. 4 is the automagic in ``run_allciphers()`` e.g.. I appreciate the good work you have done on this for getting this done. That is totally cool but for smaller tests like this it would be overdoing it.  5 is something users sometimes do not follow.

I didn't manage this week to get further with performance analysis for sockets. Next time slot is in 2 weeks for this.

Cheers, Dirk

PR #571, opened 2016-12-21
Use sockets for run_freak()

dcooper16:
This PR changes `run_freak()` to use sockets.

I added two ciphers to `$exportrsa_cipher_list`: EXP1024-RC4-MD5 (0x00,0x60) and EXP1024-RC2-CBC-MD5 (0x00,0x61). So, the list is now as follows:
```
      0x00,0x62 - EXP1024-DES-CBC-SHA            TLS_RSA_EXPORT1024_WITH_DES_CBC_SHA                SSLv3      Kx=RSA(1024)   Au=RSA     Enc=DES(56)                    Mac=SHA1     export    
      0x00,0x61 - EXP1024-RC2-CBC-MD5            TLS_RSA_EXPORT1024_WITH_RC2_56_MD5                 SSLv3      Kx=RSA(1024)   Au=RSA     Enc=RC2(56)                    Mac=MD5      export    
      0x00,0x64 - EXP1024-RC4-SHA                TLS_RSA_EXPORT1024_WITH_RC4_56_SHA                 SSLv3      Kx=RSA(1024)   Au=RSA     Enc=RC4(56)                    Mac=SHA1     export    
      0x00,0x60 - EXP1024-RC4-MD5                TLS_RSA_EXPORT1024_WITH_RC4_56_MD5                 SSLv3      Kx=RSA(1024)   Au=RSA     Enc=RC4(56)                    Mac=MD5      export    
      0x00,0x14 - EXP-EDH-RSA-DES-CBC-SHA        TLS_DHE_RSA_EXPORT_WITH_DES40_CBC_SHA              SSLv3      Kx=DH(512)     Au=RSA     Enc=DES(40)                    Mac=SHA1     export    
      0x00,0x08 - EXP-DES-CBC-SHA                TLS_RSA_EXPORT_WITH_DES40_CBC_SHA                  SSLv3      Kx=RSA(512)    Au=RSA     Enc=DES(40)                    Mac=SHA1     export    
      0x00,0x06 - EXP-RC2-CBC-MD5                TLS_RSA_EXPORT_WITH_RC2_CBC_40_MD5                 SSLv3      Kx=RSA(512)    Au=RSA     Enc=RC2(40)                    Mac=MD5      export    
 0x04,0x00,0x80 - EXP-RC2-CBC-MD5                SSL_CK_RC2_128_CBC_EXPORT40_WITH_MD5               SSLv2      Kx=RSA(512)    Au=RSA     Enc=RC2(40)                    Mac=MD5      export    
      0x00,0x0E - EXP-DH-RSA-DES-CBC-SHA         TLS_DH_RSA_EXPORT_WITH_DES40_CBC_SHA               SSLv3      Kx=DH/RSA      Au=DH      Enc=DES(40)                    Mac=SHA1     export    
      0x00,0x03 - EXP-RC4-MD5                    TLS_RSA_EXPORT_WITH_RC4_40_MD5                     SSLv3      Kx=RSA(512)    Au=RSA     Enc=RC4(40)                    Mac=MD5      export    
 0x02,0x00,0x80 - EXP-RC4-MD5                    SSL_CK_RC4_128_EXPORT40_WITH_MD5                   SSLv2      Kx=RSA(512)    Au=RSA     Enc=RC4(40)                    Mac=MD5      export
```

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -8811,18 +8811,27 @@ run_tls_fallback_scsv() {
 # Factoring RSA Export Keys: don't use EXPORT RSA ciphers, see https://freakattack.com/
 run_freak() {
      local -i sclient_success=0
-     local -i nr_supported_ciphers=0
-     # with correct build it should list these 7 ciphers (plus the two latter as SSLv2 ciphers):
-     local exportrsa_cipher_list="EXP1024-DES-CBC-SHA:EXP1024-RC4-SHA:EXP-EDH-RSA-DES-CBC-SHA:EXP-DH-RSA-DES-CBC-SHA:EXP-DES-CBC-SHA:EXP-RC2-CBC-MD5:EXP-RC2-CBC-MD5:EXP-RC4-MD5:EXP-RC4-MD5"
-     local addcmd="" addtl_warning=""
+     local -i i nr_supported_ciphers=0
+     # with correct build it should list these 9 ciphers (plus the two latter as SSLv2 ciphers):
+     local exportrsa_cipher_list="EXP1024-DES-CBC-SHA:EXP1024-RC2-CBC-MD5:EXP1024-RC4-SHA:EXP1024-RC4-MD5:EXP-EDH-RSA-DES-CBC-SHA:EXP-DH-RSA-DES-CBC-SHA:EXP-DES-CBC-SHA:EXP-RC2-CBC-MD5:EXP-RC4-MD5"
+     local exportrsa_tls_cipher_list_hex="00,62, 00,61, 00,64, 00,60, 00,14, 00,0E, 00,08, 00,06, 00,03"
+     local exportrsa_ssl2_cipher_list_hex="04,00,80, 02,00,80"
+     local addcmd="" addtl_warning="" hexc
      local cve="CVE-2015-0204"
      local cwe="CWE-310"
      local hint=""
+     local using_sockets=true
+
+     "$SSL_NATIVE" && using_sockets=false
 
      [[ $VULN_COUNT -le $VULN_THRESHLD ]] && outln && pr_headlineln " Testing for FREAK attack " && outln
      pr_bold " FREAK"; out " ($cve)                     "
 
-     nr_supported_ciphers=$(count_ciphers $(actually_supported_ciphers $exportrsa_cipher_list))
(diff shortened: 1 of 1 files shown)


dcooper16:
I tested it on a lot of servers and didn't get any false positives.

Do you have any more information on the second `-E` output above. I can't understand all the error messages under SSLv2. `run_cipher_per_proto()` should only call `sslv2_sockets()` once, and it should only try to load `$HOSTCERT` if `$SHOW_SIGALGO` is `true`. So, why are there multiple error messages indicating that there is an error opening `$HOSTCERT`?

Can you tell whether the false positives in `run_freak()` are happening as a result of it "detecting" SSLv3-TLSv1.2 ciphers or SSLv2 ciphers?

drwetter:
they were public hosts I found (my fat server I simulate those things has a hw problem).

As I don't have your email address and I don't want people blame publicly I posted it some place else and you should have received a mail, then removed it. The IPv4 addresses were "encoded". Got it?

dcooper16:
I tried one of the servers you mentioned from my computer at home, and I am now seeing the same results that you are. Perhaps there is a firewall where I work that is blocking SSLv2, but that makes it appear as if the server is returning an empty response.

I looked at the full ServerHello being returned in response to the SSLv2 ClientHello in `run_freak()` and the last part of the response is:
```
000006f0  cd f6 83 01 00 80 07 00  c0 9e 40 49 90 46 f8 01  |..........@I.F..|
00000700  57 43 14 5b 26 e4 33 c5  2b                       |WC.[&.3.+|
```
So, the server is responding that it supports 0x010080 and 0x0700c0, even though those cipher suites are not in the ClientHello. That's the reason for the false positive. I'll have to modify the sockets version of `run_freak()` to check whether any of the ciphers specified in the ClientHello appear in the ServerHello, not just whether the server responded with a non-empty list of ciphers. I'll have to make a similar change to the sockets version of `std_cipherlists()`.

drwetter:
Thank you. Works for me now ;-)

This is still slower than doing it natively -- for the two examples given ~1 second, for the other one ~2 seconds.

This bugs me and needs to be addressed. Best would be to do some performance analysis, see #519 .
Maybe I can start with that this week.

Issue #120, opened 2015-06-11
common primes test (LOGJAM (CVE-2015-4000))


drwetter:
discussion see #107 .

How to test it properly?

drwetter:
Ubuntu 12/.04 LTS at least has now also 2048 Bit DH:

     Changelog for libssl1.0.0 (http://changelogs.ubuntu.com/changelogs/pool/main/o/openssl/openssl_1.0.1-4ubuntu5.31/changelog) [115 kB]
     openssl (1.0.1-4ubuntu5.31) precise-security; urgency=medium

      * SECURITY IMPROVEMENT: reject dh keys smaller than 768 bits
       - debian/patches/reject_small_dh.patch: reject small dh keys in
         ssl/s3_clnt.c, ssl/ssl.h, ssl/ssl_err.c, update documentation in
         doc/ssl/SSL_CTX_set_tmp_dh_callback.pod, make s_server use 2048-bit
         dh in apps/s_server.c, clarify docs in doc/apps/dhparam.pod,
         switch defaut dh to 2048-bit in apps/dhparam.c, apps/gendh.c.

While I can't tell that those pregenerated primes (`apps/dh2048.pem`) will pose a threat in the future -- this would be good for fingerprinting the patch status : 1024 Bit on identified 12.04 LTS means likely missing patches. Debian wheezy and opensuse 13.1 are still @ 1024 Bit DH (no backport also of the SSLCerfiticateKeyFile)

Still the question remains how to test externally for those precomputed primes.

jpluimers:
Is there a way to get the hash of these primes externally? If so, then you could have a table of (hashes of) common pre-generated primes in testssl.sh and test them.

drwetter:
Am 15. August 2015 10:50:31 MESZ, schrieb Jeroen Wiert Pluimers <notifications@github.com>:
>Is there a way to get the hash of these primes externally? If so, then
>you could have a table of (hashes of) common pre-generated primes in
>testssl.sh and test them.

If you could do that,  it would be great. 

The first set of primes is in openssl source code. Lookout for, *512*\.pem *1024*\.pem (currently on the road, sorry about being vague).

Cheers, Dirk

jpluimers:
Are you looking for something like this?

MD5:

    RetinaMBPro1TB:openssl jeroenp$ md5 `find ./apps/dh*.pem`
    MD5 (./apps/dh1024.pem) = 1ccf6d71bb16d2d04cb81b21ff724778
[...]

PR #572, opened 2016-12-21
Use sockets for run_logjam()

dcooper16:
This PR changes `run_logjam()` to use sockets.


drwetter:
Thx, David.

Speed is as expected similar here and it compensates the deficiency of a (nowadays) standard openssl where no  DHE EXPORT ciphers have been compiled in. A good step fwd again!

 Do you think you can address #120, too? The -msg option in https://superuser.com/questions/905557/openssl-display-dh-parameters doesn't seem to be nice to me. The best collection of common primes looks seems to me the one from the NSE script (https://svn.nmap.org/nmap/scripts/ssl-dh-params.nse)

Issue #554, opened 2016-12-07
Definitions of cipher lists in run_std_cipherlists()


dcooper16:
Hi @drwetter ,

I'm working to try to modify `run_std_cipherlists()` to use sockets, and I have a question about the definitions of the lists. Some are straightforward, but I am unclear about others.

```
     std_cipherlists 'NULL:eNULL'                       " Null Ciphers             " 1 "NULL"
```
The "Null Ciphers" list is easy; it is just those for which "Enc=NULL."

```
     std_cipherlists 'aNULL'                            " Anonymous NULL Ciphers   " 1 "aNULL"
```
"Anonymous NULL Ciphers" is those for which "Au=None."
```
     std_cipherlists 'ADH'                              " Anonymous DH Ciphers     " 1 "ADH"
```

 "Anonymous DH Ciphers" is those for which "Au=None" and "Kx=DH" (rather than "Kx=ECDH") 

```
     std_cipherlists 'EXPORT40'                         " 40 Bit encryption        " 1 "EXPORT40"
```
"40 Bit encryption" is those for which the "Enc=..." contains "(40)."

```
     std_cipherlists 'EXPORT56'                         " 56 Bit encryption        " 1 "EXPORT56"
```

This is the first difficult one, since 'EXPORT56' is not the same as ciphers with 56-bit encryption. Should the list be all ciphers that use 56-bit encryption, as the text that is displayed in the output implies, or only those that use 56-bit encryption that are marked as "export"?

```
> ./openssl.Linux.x86_64.krb5 ciphers -V EXPORT56
          0x00,0x63 - EXP1024-DHE-DSS-DES-CBC-SHA SSLv3 Kx=DH(1024) Au=DSS  Enc=DES(56)   Mac=SHA1 export
          0x00,0x62 - EXP1024-DES-CBC-SHA     SSLv3 Kx=RSA(1024) Au=RSA  Enc=DES(56)   Mac=SHA1 export
          0x00,0x61 - EXP1024-RC2-CBC-MD5     SSLv3 Kx=RSA(1024) Au=RSA  Enc=RC2(56)   Mac=MD5  export
          0x00,0x65 - EXP1024-DHE-DSS-RC4-SHA SSLv3 Kx=DH(1024) Au=DSS  Enc=RC4(56)   Mac=SHA1 export
[...]

PR #573, opened 2016-12-22
Use sockets for run_std_cipherlists()

dcooper16:
This PR changes `run_std_cipherlists()` to use sockets. As noted in isse #554, I have some questions about the definitions of the cipher lists, but I wrote the code so that the ciphers that are tested when using sockets are the same as those that are tested when using OpenSSL. For a few of the cipherlists, the sockets version tests a few additional ciphers; but these are ciphers that are not supported by OpenSSL, and whose definitions are consistent with the ciphers that OpenSSL includes.

As written, `std_cipherlists()` will use sockets for testing by default, except in two cases:
* If the `$SSL_NATIVE` is true, then only OpenSSL is used, and if OpenSSL doesn't support any ciphers in the cipherlist, then a "Local Problem" warning is displayed.
* If `$FAST` is true (but `$SSL_NATIVE` is false), then OpenSSL is used whenever it supports at least one cipher from the cipherlist, and `tls_sockets()` (or `sslv2_sockets()`) is only used when OpenSSL doesn't support any ciphers from the cipherlist.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4440,105 +4440,39 @@ run_protocols() {
 run_std_cipherlists() {
      local hexc hexcode strength
      local -i i
-     local null_ciphers="" anon_ciphers="" adh_ciphers="" exp40_ciphers=""
-     local exp56_ciphers="" exp_ciphers="" low_ciphers="" des_ciphers=""
-     local medium_ciphers="" tdes_ciphers="" high_ciphers=""
-     local sslv2_null_ciphers="" sslv2_anon_ciphers="" sslv2_adh_ciphers="" sslv2_exp40_ciphers=""
-     local sslv2_exp56_ciphers="" sslv2_exp_ciphers="" sslv2_low_ciphers="" sslv2_des_ciphers=""
-     local sslv2_medium_ciphers="" sslv2_tdes_ciphers="" sslv2_high_ciphers=""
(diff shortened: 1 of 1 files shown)


drwetter:
Hi David,

this is an important PR.

Q: Why do you compile the ciphers from the internal list? Except the HIGH / high_ciphers do you expect many changes over time?

dcooper16:
Hi Dirk,

I'm okay with changing the code to hard-code each of the list of ciphers, once there is agreement on what the lists should contain. It would have the advantage that the sockets version would work even if the cipher-mapping.txt file were missing. I think it would be helpful, though, to keep the current code (either commented out or in a separate file), in case the lists needed to be updated or modified at some point in the future.

drwetter:
Hi David,

my suggestion is to just output what your magic does so that it generates the standard lists.

Your proposal with the separate file sounds good. We could use this then as a generator for this standard lists.

Cheers, Dirk

dcooper16:
Okay, I replaced the code to compile the cipher lists with hard-coded values, and a created a separate utility (submitted in a different PR) that will generate these cipher lists as well as all of the other cipher lists used in testssl.sh (e.g., robust PFS ciphers, CBC ciphers, weak DH ciphers, export RSA ciphers).

Some of these lists may never change, but I think it's still useful for the utility to compile these lists, since the source code can be referenced by anyone who wants a precise definition of the contents of any of the lists.

drwetter:
Thank you. 

That was the first occasion when "Local problem: No XXXXX encryption configured in /usr/bin/openssl" popped up when doing a default run. 

Now: Gone 👍

PR #574, opened 2016-12-23
Use sockets for run_http2()

dcooper16:
This PR changes `run_http2()` so that it uses `tls_sockets()` rather than failing, if `$OPENSSL` does not support the `-alpn` option. If `$OPENSSL` supports the `-alpn` option (or if `$SSL_NATIVE` is true), then this PR has no effect.


drwetter:
Thx!

PR #579, opened 2017-01-05
Use sockets for run_crime()

dcooper16:
This PR changes `run_crime()` to use `tls_sockets()` rather than failing if `$OPENSSL` lacks zlib support, unless `$SSL_NATIVE` is `true`.

At the moment, the ClientHello created by `socksend_tls_clienthello()` only specifies the NULL compression method. So, this PR adds a new parameter to `socksend_tls_clienthello()` and `tls_sockets()` to allow the caller to request that additional compression methods (DEFLATE and LZS) be specified in the ClientHello.

This PR makes another change to `run_crime()`. At the moment, if `$OPENSSL s_client` fails to connect to the server, `run_crime()` will report that the server is not vulnerable, since the output from `$OPENSSL s_client` includes the line "Compression: NONE" (see below). This PR changes that by checking whether the connection was successful, and reporting a "test failed (couldn't connect)" warning if it wasn't successful, rather than reporting "not vulnerable (OK)".

```
CONNECTED(00000003)
140338777061024:error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version:s23_clnt.c:769:
---
no peer certificate available
---
---
SSL handshake has read 7 bytes and written 389 bytes
---
New, (NONE), Cipher is (NONE)
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1
    Cipher    : 0000
    Session-ID: 
    Session-ID-ctx: 
    Master-Key: 
    Key-Arg   : None
    PSK identity: None
    PSK identity hint: None
    SRP username: None
    Start Time: 1483645971
    Timeout   : 300 (sec)
    Verify return code: 0 (ok)
---
```

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -8998,6 +8998,9 @@ run_crime() {
           fi
      else
           [[ "$OSSL_VER" == "0.9.8"* ]] && addcmd="-no_ssl2"
+          if [[ $OSSL_VER_MAJOR.$OSSL_VER_MINOR == "1.1.0"* ]] || [[ $OSSL_VER_MAJOR.$OSSL_VER_MINOR == "1.1.1"* ]]; then
+               addcmd="-comp"
+          fi
           $OPENSSL s_client $OPTIMAL_PROTO $BUGS $addcmd $STARTTLS -connect $NODEIP:$PORT $PROXY $SNI </dev/null &>$TMPFILE
           sclient_connect_successful $? $TMPFILE
           sclient_success=$?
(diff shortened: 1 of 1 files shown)


dcooper16:
I just discovered a problem with `run_crime()` when using OpenSSL 1.1.0. Starting with OpenSSL 1.1.0 the ClientHello created by s_client will only specify the NULL compression method even if OpenSSL is compiled with zlib support, unless the `-comp` flag is specified on the command line. (For versions of OpenSSL prior to 1.1.0, specifying `-comp` on the command line will result in failure.)

So, at the moment, if OpenSSL 1.1.0 is being used and it has been compiled with zlib support, then `run_crime()` will report "not vulnerable (OK)" even if the server is actually vulnerable.

So, I modified the PR so that if it is determined that OpenSSL has zlib support then `-comp` will be added to the `$OPENSSL s_client` command line if the OpenSSL version is 1.1.0* or 1.1.1*.

drwetter:
Thank you again

PR #580, opened 2017-01-05
Detect $TESTSSL_INSTALL_DIR when testssl is a symlinked command

gniltaws:
I have a symlink to testssl.sh in /usr/local/bin/testssl so I can run it outside the git repository.  With the latest pull from the 2.9dev branch I wasn't able to do that anymore.  This works without having realpath installed.  I also removed trailing whitespace since that is a pet peeve of mine.


drwetter:
Thank you.

PR #582, opened 2017-01-12
Create static cipher lists for testssl.sh

dcooper16:
This PR adds a new utility that generates the various static cipher lists that appear in testssl.sh.

This utility serves two purposes:
* It can be run whenever new ciphers are added to cipher-mapping.txt to see if any of the lists in testssl.sh need to be updated. (This includes if cipher-mapping.txt is modified to add OpenSSL-style names for ciphers that are currently listed, but that have not yet been assigned such names.)
* The source code can be used as a reference in order to understand how each of the lists is defined.


drwetter:
Thx!

PR #583, opened 2017-01-12
run_client_simulation() bugfix

dcooper16:
There are two places in `run_client_simulation()` in which `$OPENSSL s_client` is called, after which there is a `debugme echo` line to display the `$OPENSSL s_client` command line when testssl.sh is being run in debug mode, and then `sclient_connect_successful $? $TMPFILE` is called to determine whether `$OPENSSL s_client` successfully established a connection.

So, `sclient_connect_successful()` is being passed the result of the `debugme()` call, which always returns 0, rather than the result of the `$OPENSSL s_client` call.

This PR fixes the problem by moving the `debugme()` line to before the call to `$OPENSSL s_client`, so that  `sclient_connect_successful()` is passed the results of the `$OPENSSL s_client` call.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3991,8 +3991,8 @@ run_client_simulation() {
                fi
           else
                ! "$HAS_NO_SSL2" && protos[i]="$(sed 's/-no_ssl2//' <<< "${protos[i]}")"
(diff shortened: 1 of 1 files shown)


drwetter:
oh, yes, thx :-)

PR #585, opened 2017-01-13
Show selected curve

dcooper16:
This PR changes `read_dhbits_from_file()` so that, when the "quiet" parameter is absent, the selected curve is shown in addition to the number of bits. This PR only affects the output of `run_client_simulation()` and the "Negotiated cipher" in `run_server_preference()`.


drwetter:
yes, that's good additional information. Thx

PR #590, opened 2017-01-18
Generate list of all DHE ciphers

dcooper16:
This PR adds a function that generates a list of all DHE ciphers for `run_logjam()`.


drwetter:
thx! 

How about changing DHE to DH? Especially for LOGJAM one could find this irritating as I could be interpreted as "DH Export". Also we don't use DH and DHE consistently within testssl.

PR #595, opened 2017-01-24
Rename cipher lists for run_logjam()

dcooper16:
This PR renames the cipher lists for `run_logjam()` in generate_static_cipher_lists.sh to align with their names in testssl.sh, as requested in #590.

I think these names are still open for misinterpretation, however, since its not clear whether "dh_cipher" refers to ciphers that use static DH keys, ephemeral DH keys, or both.


drwetter:
> I think these names are still open for misinterpretation, however, since its not clear whether "dh_cipher" refers to ciphers that use static DH keys, ephemeral DH keys, or both.

Good point. I am under the impression that this distinction is not correct in the OpenSSL naming scheme either.

What do you suggest?

drwetter:
thx

PR #600, opened 2017-01-25
Print unsupported ciphers in light grey

dcooper16:
This PR changes testssl.sh so that when ciphers are being listed in wide mode (i.e., using `neat_list()`) and the `--show-each` option is set, ciphers that are not available are printed in light grey, whereas ciphers that are available continue to be printed in black. This makes it easier to distinguish between ciphers that are available and those that are not (the "available/"not a/v" column remains).

This PR does not change the way that ciphers that are available are printed, but it includes a hook that would allow that to change. For example, for ciphers that are available, the name of the cipher suite could be printed in a different color depending on its quality (as is done for the "Negotiated cipher" in `run_server_preference()`). The same could be done for the "Encryption" and "Bits" columns.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2203,8 +2203,7 @@ neat_header(){
 neat_list(){
      local hexcode="$1"
      local ossl_cipher="$2" tls_cipher=""
-     local kx enc strength
-     local -i i str_len
+     local kx enc strength line
 
      kx="${3//Kx=/}"
      enc="${4//Enc=/}"
@@ -2220,6 +2219,12 @@ neat_list(){
 
      [[ -n "$ADD_RFC_STR" ]] && tls_cipher="$(show_rfc_style "$hexcode")"
 
+     if [[ "$5" == "false" ]]; then
+          line="$(printf -- " %-7s %-33s %-10s %-10s%-8s${ADD_RFC_STR:+ %-49s}${SHOW_EACH_C:+  %-0s}" "$hexcode" "$ossl_cipher" "$kx" "$enc" "$strength" "$tls_cipher")"
+          pr_litegrey "$line"
+          return 0
+     fi
+
      #printf -- "%q" "$kx" | xxd | head -1
      # length correction for color escape codes (printf counts the escape color codes!!)
      if printf -- "%q" "$kx" | egrep -aq '.;3.m|E\[1m' ; then     # here's a color code which screws up the formatting with printf below
@@ -2230,46 +2235,6 @@ neat_list(){
           while [[ ${#kx} -lt 13 ]]; do                   # so it'll be filled up ok
                kx="$kx "
           done
-     fi
-     if [[ -n "$5" ]]; then
-          # FIXME: When $5 is true, highlight ciphers based on quality.
-          if [[ "$5" == "false" ]]; then
-               str_len=${#hexcode}
-               hexcode="$(pr_litegrey "$hexcode")"
-               for (( i=str_len; i < 7; i++ )); do
-                    hexcode+=" "
-               done
-
-               str_len=${#kx}
-               kx="$(pr_litegrey "$kx")"
-               for (( i=str_len; i < 10; i++ )); do
-                    kx+=" "
-               done
-
-               str_len=${#enc}
-               enc="$(pr_litegrey "$enc")"
-               for (( i=str_len; i < 10; i++ )); do
-                    enc+=" "
-               done
-
-               str_len=${#strength}
-               strength="$(pr_litegrey "$strength")"
-               for (( i=str_len; i < 8; i++ )); do
-                    strength+=" "
-               done
-
-               str_len=${#tls_cipher}
-               tls_cipher="$(pr_litegrey "$tls_cipher")"
-               for (( i=str_len; i < 49; i++ )); do
-                    tls_cipher+=" "
-               done
-
-               str_len=${#ossl_cipher}
-               ossl_cipher="$(pr_litegrey "$ossl_cipher")"
-               for (( i=str_len; i < 33; i++ )); do
-                    ossl_cipher+=" "
-               done
-          fi
      fi
      #echo "${#kx}"                            # should be always 20 / 13
      printf -- " %-7s %-33s %-10s %-10s%-8s${ADD_RFC_STR:+ %-49s}${SHOW_EACH_C:+  %-0s}" "$hexcode" "$ossl_cipher" "$kx" "$enc" "$strength" "$tls_cipher"
--- a/testssl.sh
+++ b/testssl.sh
@@ -2187,10 +2187,14 @@ neat_header(){
 # arg2: cipher in openssl notation
 # arg3: keyexchange
 # arg4: encryption (maybe included "export")
+# arg5: "true" if the cipher's "quality" should be highlighted
+#       "false" if the line should be printed in light grey
+#       empty if line should be printed in black
 neat_list(){
      local hexcode="$1"
      local ossl_cipher="$2" tls_cipher=""
      local kx enc strength
+     local -i i str_len
 
      kx="${3//Kx=/}"
      enc="${4//Enc=/}"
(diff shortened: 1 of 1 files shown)


drwetter:
Thx, David. I would need some instruction here, can you help me out pls? ;-)

``./testssl.sh --ssl-native --show-each -e <NODE>`` doesn't do it.

testssl.sh
```diff
@@ -2217,6 +2221,46 @@ neat_list(){
...
+          # FIXME: When $5 is true, highlight ciphers based on quality.
+          if [[ "$5" == "false" ]]; then
+               str_len=${#hexcode}
+               hexcode="$(pr_litegrey "$hexcode")"
+               for (( i=str_len; i < 7; i++ )); do
+                    hexcode+=" "
+               done
+
+               str_len=${#kx}
+               kx="$(pr_litegrey "$kx")"
```
drwetter: If it is forseeable that the line is printed in grey I think it's better collecting the output of a line in a variable and then eventually do a  ``pr_litegrey $whole_line``.

If I got this right...

dcooper16:
Okay, I changed `neat_list()` so that if the line is to be printed in light grey, the formatted line is placed in a variable, which is then printed using `pr_litegrey()`.

I can't explain why it doesn't seem to be working for you. I've tried it on a couple of different operating systems, using multiple versions of OpenSSL (with and without `--ssl-native`), and it always works as expected for me.

If it's still not working for you, could you run it with `bash -x` and send me the results?

drwetter:
Should have looked at the ``--log`` file. Sorry.

Catch is: It doesn't work on black background -- at least not for me. See attached (left konsole, right xterm). If you don't have a better suggestion I'll
black
 change that to grey later.

dcooper16:
Hi Dirk,

I hadn't thought to try out different color schemes for the terminal. It definitely complicates things. ):

I've just tried experimenting with Konsole, and for me grey works better than litegrey with a black background, but not very well with a white background (including when I run the output through `aha` and view it in a web browser). Things really look bad with "Solorized" when using grey or "Solorized Light" when using light grey.

Color schemes like "Green on Black" or "Red on Black" make it difficult to distinguish text printed using `out` versus text printed using `pr_done_good()/pr_done_best()` or `pr_svrty_high()/pr_svrty_critical()`.

Unfortunately, there doesn't seem to be a way to find out what background color the terminal is using. I don't thing there's anything we can do to accommodate all color schemes, but I'll see if I can find a alternative to light grey (or grey) that seems to work better across a range of terminal color schemes.

drwetter:
Hi David,

Yes,  the folks which do not use a default scheme are lost unfortunately. I rather blame the scheme,  Solarized seems to be one of the worst. I tried that last week for different means and quickly switched back. 

My idea is more supporting black and white background only. Unless we go for 256 colors which has little usage out there I see no way yet.

Let me know if you come to a different conclusion...

Cheers,  Dirk 


⁣-- 

AlGreed:
Just small idea: how about to experiment with the font? make them italic, underlined or strikethrough or...?

Using escape characters (echo -e "\e[3mitalic\e[0m") or tput.

drwetter:
@AlGreed : italic and underlined are already defined in testssl.sh -- for both code and usage. (IIRC tput didn't work in BSD).

All font experiments are not really appropriate here.

AlGreed:
Ok. good to know. Thanks!

dcooper16:
I can't really think of anything at the moment. Before I noticed the last few comments in this PR, I found https://en.wikipedia.org/wiki/ANSI_escape_code#Colors and tried out the different grayscale colors in the 256-bit mode (the script below displays them all). Using one of the colors in the middle of this range (e.g., 245) seems to work well in the terminal, but when the output is run through `aha`, only the current `pr_litegrey()` results in the unavailable ciphers being printed in a different color than the available ones.

I tried printing the unavailable ciphers in _italics_, but it doesn't help much in the terminal, and when using `aha` to create HTML, the italics gets lost.

So, at the moment I can't think of a good way to make the available and unavailable ciphers easily distinguishable that works with HTML (as generated by `aha`) as well as in terminals with black backgrounds and terminals with white backgrounds.

```
#!/usr/bin/env bash
#
# vim:ts=5:sw=5:expandtab
# we have a spaces softtab, that ensures readability with other editors too

[ -z "$BASH_VERSINFO" ] && printf "\n\033[1;35m Please make sure you're using \"bash\"! Bye...\033[m\n\n" >&2 && exit 245
[ $(kill -l | grep -c SIG) -eq 0 ] && printf "\n\033[1;35m Please make sure you're calling me without leading \"sh\"! Bye...\033[m\n\n"  >&2 && exit 245

# debugging help:
readonly PS4='${LINENO}> ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'

out(){
          printf -- "%b" "${1//%/%%}"
}
outln() { out "$1\n"; }

pr_off()          { out "\033[m"; }

out "\033[0;37mHello" ; pr_off; outln " Hello (current litegrey)"
out "\033[1;30mHello" ; pr_off; outln " Hello (current grey)"
out "\033[38;5;232mHello" ; pr_off; outln " Hello (232)"
out "\033[38;5;233mHello" ; pr_off; outln " Hello (233)"
out "\033[38;5;234mHello" ; pr_off; outln " Hello (234)"
out "\033[38;5;235mHello" ; pr_off; outln " Hello (235)"
out "\033[38;5;236mHello" ; pr_off; outln " Hello (236)"
out "\033[38;5;237mHello" ; pr_off; outln " Hello (237)"
out "\033[38;5;238mHello" ; pr_off; outln " Hello (238)"
out "\033[38;5;239mHello" ; pr_off; outln " Hello (239)"
out "\033[38;5;240mHello" ; pr_off; outln " Hello (240)"
out "\033[38;5;241mHello" ; pr_off; outln " Hello (241)"
out "\033[38;5;242mHello" ; pr_off; outln " Hello (242)"
out "\033[38;5;243mHello" ; pr_off; outln " Hello (243)"
out "\033[38;5;244mHello" ; pr_off; outln " Hello (244)"
out "\033[38;5;245mHello" ; pr_off; outln " Hello (245)"
out "\033[38;5;246mHello" ; pr_off; outln " Hello (246)"
out "\033[38;5;247mHello" ; pr_off; outln " Hello (247)"
out "\033[38;5;248mHello" ; pr_off; outln " Hello (248)"
out "\033[38;5;249mHello" ; pr_off; outln " Hello (249)"
out "\033[38;5;250mHello" ; pr_off; outln " Hello (250)"
out "\033[38;5;251mHello" ; pr_off; outln " Hello (251)"
out "\033[38;5;252mHello" ; pr_off; outln " Hello (252)"
out "\033[38;5;253mHello" ; pr_off; outln " Hello (253)"
out "\033[38;5;254mHello" ; pr_off; outln " Hello (254)"
out "\033[38;5;255mHello" ; pr_off; outln " Hello (255)"

exit 0
```

drwetter:
> I tried printing the unavailable ciphers in italics, but it doesn't help much in the terminal, and when using aha to create HTML, the italics gets lost.

Italic, bold, underline and other font modifications I'd like rather to reserve for emphasizing output, headlines and similar. This all has been taken. I think also testssl should not overdo it with respect to font outlines or color. One of the good features of testssl is that it is clearly giving the user feedback what's good and what not. Overdoing it would mean blurring this information. So I am careful here. Think of a good flyer/brochure or a good web site which is done by a graphics designer: fonts and colors are not being used randomly and only being used for specific purposes.

> So, at the moment I can't think of a good way to make the available and unavailable ciphers easily distinguishable that works with HTML (as generated by aha) as well as in terminals with black backgrounds and terminals with white backgrounds.

As far as HTML is concerned: aha is a cool thing -- however I still consider it as a kludge -- a good working one though. My idea like two years back was to do this natively or via JSON-->HTML converter. Now as JSON is there I have my doubts that this is going to fly as the ideas of both are different: one is for humans, the other one is more a machine readable format. Conversion thus seems too complicated and HTML output would probably be easier if done natively by testssl.

Ok, this doesn't help us now. ;-)

Bottomline: I would suggest to either roll this back or to leave it in there as a hook and replace ``pr_litegrey`` by ~``pr_weak`` which will be defined as ``out`` for now.

drwetter:
PS: Maybe this  https://stackoverflow.com/questions/28096697/how-to-get-current-terminal-color-pair-in-bash/28334701#28334701 could help reading the background color. Don't know how portable that is though or whether it works at all (haven't tried)

dcooper16:
I believe that I came across the stackoverflow.com page when looking for a solution, but the comments seemed to suggest there's no standard way to do this, even if there might be a solution that would work for one particular terminal emulator.

I understand the desire to not overdo it with the use of colors, but I think more use of colors could make the output more readable. When `--show-each` is used, I think printing the unavailable ciphers in light gray makes it much easier to see which ciphers are available and which are not.

Below is an example of Cipher Order and Client Simulation in which the ciphers are color-coded based on the rules currently used to print the "Negotiated Cipher." While I think that too many of the ciphers are labeled as `pr_svrty_medium`, I like (particularly for TLSv1.2) that the colors allow me to quickly see whether the "best" ciphers appear first in the list and the "worst" ciphers appear last in the list. [The main problem I see with the coloring scheme below is that all CBC ciphers are marked as `pr_svrty_medium`. Given that for SSLv3 - TLSv1.1 the only choices are CBC ciphers (`pr_svrty_medium`) or RC4 ciphers (`pr_svrty_high`), the output seems overly negative for SSLv3 - TLSv1.1. It seems that at least some ciphers that can be used with SSLv3 - TLSv1.1 need to be marked as not bad (`out`) or even good (`pr_done_good`).]

cipherorder
browsersim

drwetter:
@dcooper16 , that color scheme is interesting. However:
* too much IMO for an average user. But maybe ``--color=3|COLOR=3`` could be a hook for those which prefer this
* maybe using a 256color scheme looks better here?
* As a general idea I do not like testssl to use the same color for different purposes.
* Consequenlty we would need to re-discuss the color schme: Currently we have for BEAST/CBC ciphers yellow -- supposed there's TLS >=1.1. Magenta is reserved currently for local problems, i.e. warnings That could be changed with cyan which is only used for one purpose (wide mode to emphasize available ciphers). While writing this: I haven't put any consideration into it... 

For the lightgrey thing: During the weekend I'll swap that with pr_weak as discussed.

drwetter:
strikethrough was not a bad idea but it's ugly and doesn't work on konsole. Litegrey actually emphasizes the text on a white xterm.

So I'll leave the hook in here for the future.

Issue #10, opened 2014-07-05
replace echo -e with printf for portability between Linux and BSD/Darwin


dmitris:
echo -e is not supported on BSD-like systems including Mac OS [1] which leads to garbled output.  It would be great to replace it with printf. https://github.com/dmitris/testssl.sh/blob/crossplat/testssl.sh has the first attempt on the change (in branch cross plat).

[1] http://en.wikipedia.org/wiki/Echo_command
Some variants of Unix, such as Linux, support the options -n and -e, and do not process escape sequences unless the -e option is supplied. For example, FGRED=`echo -e "\033[31m"` might be used under Linux. Unfortunately, such options are non standard[3] due to historical incompatibilities between BSD and System V; the printf command can be used in situations where this is a problem. It is therefore recommended that printf be used to ensure that escape sequences are processed.

PR #609, opened 2017-02-01
Handle renaming of the Supported Elliptic Curves Extension

dcooper16:
[RFC 4492](https://www.rfc-editor.org/rfc/rfc4492.txt) introduced the Supported Elliptic Curves Extension, but this extension was renamed Supported Groups in [RFC 7919](https://www.rfc-editor.org/rfc/rfc7919.txt). Following [RFC 7919](https://www.rfc-editor.org/rfc/rfc7919.txt) (and [TLSv1.3](https://www.ietf.org/id/draft-ietf-tls-tls13-18.txt)), `parse_tls_serverhello()` refers to this extension as "supported groups/#10". Since, at the moment, OpenSSL's s_client refers to this extension as "elliptic curves/#10", the extension sometimes appears twice in the "TLS extensions" line, if it is detected by both OpenSSL (in `get_server_certificate()`) and `tls_sockets()` (in `determine_tls_extensions()`):
```
 TLS extensions (standard)    "renegotiation info/#65281" "elliptic curves/#10" "EC point formats/#11" "supported groups/#10"
```
This PR fixes the problem of the extension appearing twice in the "TLS extensions" line by replacing any instances of "elliptic curves/#10" with "supported_groups/#10" in the `$tls_extensions` line extracted from `$OPENSSL s_client`. This PR also changes "supported groups/#10" to "supported_groups/#10" in `parse_tls_serverhello()`, since the current development branch of OpenSSL uses "supported_groups" to refer to this extension (see https://github.com/openssl/openssl/pull/1825).

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -5427,7 +5427,10 @@ determine_tls_extensions() {
                success=$?
           fi
           if [[ $success -eq 0 ]]; then
-               tls_extensions=$(grep -a 'TLS server extension ' $TMPFILE | sed -e 's/TLS server extension //g' -e 's/\" (id=/\/#/g' -e 's/,.*$/,/g' -e 's/),$/\"/g')
+               tls_extensions=$(grep -a 'TLS server extension ' $TMPFILE | \
+                    sed -e 's/TLS server extension //g' -e 's/\" (id=/\/#/g' \
+                        -e 's/,.*$/,/g' -e 's/),$/\"/g' \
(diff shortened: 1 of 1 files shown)


drwetter:
Thx, David.

A good idea, not sure whether it's the right moment yet. Thus the question to you: which hosts in general have as of today both "#10" as tested by testssl?

Issue #587, opened 2017-01-13
Let's Encrypt certificate reports CRL issue 


gerhard-tinned:
I have a certificate from Let's Encrypt and the testssl.sh report the following (in red).

<pre>
 Certificate Revocation List  --
</pre>

Why is it reported like an error??

drwetter:
yes, thank you. 

It should be checked whether there's one mechanism of certificate revocation and if not then it should shout out loud.

gerhard-tinned:
That sounds perfect. Can I help in any way?

drwetter:
Thx. In general yes but I would prefer to have a look myself at this piece of code as it was written some time back I would do it differently today.

gerhard-tinned:
No problem. ;) 

Let me know when I can test something. ;-)

drwetter:
now e.g. ;-)

gerhard-tinned:
I have tested it, works for me! Thanks for the quick fix. :)

I use the red colour to determine the number of errors. Or is there a better way?

drwetter:
Thx for reporting!

At the moment there's no other elegant way. JSON or CSV output could better suit your needs..

Cheers, Dirk


⁣-- 
Sent via mobile. Excuse my brevity, my typos and the autocorrection​

gerhard-tinned:
Well, I guess the standard would be json. But for me, a simple return code with the number of errors would be more than enough. 

I run the script for a number of hosts and would like to know for each of them if they passed(0 errors) or not (more than 0 errors).

drwetter:
I wouldn't call them errors as in the sense of the program it is a malfunction. 

In any case, yes, sounds reasonable -- see #327.

PR #612, opened 2017-02-02
Update fix to 587

dcooper16:
In the revised code that was created to address #587, nothing is printed after "OCSP URI" if there is neither an OCSP URI nor a CRL URI. Instead, "--" should be printed (perhaps it should be printed in red, just as the "--" for the CRL is).

I still believe that there is an inconsistency in the reporting of the output, however. At the moment, the "Certificate Revocation List" and "OCSP URI" lines indicate it is acceptable as long as the certificate contains either a CRL URI or an OCSP URI. However, the "OCSP stapling" line reports a minor finding if an OCSP response was not included in the server's reply. Shouldn't we just assume that if the certificate doesn't include an OCSP URI, then it wouldn't be possible for the server to obtain an OCSP response to staple to its reply? If so, then it seems that no OCSP stapling should only be considered a finding if an OCSP URI is present.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6198,7 +6198,7 @@ certificate_info() {
                fileout "${json_prefix}crl" "INFO" "No CRL provided"
           else
                pr_svrty_highln "-- (NOT ok)"
-               fileout "${json_prefix}crl" "HIGH" "Neither CRL nor  OCSP URL provided"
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, David.

> If so, then it seems that no OCSP stapling should only be considered a finding if an OCSP URI is present.

yes

Issue #605, opened 2017-01-30
openssl binary under LIS (bash for Windows)


marcstern:
Under LIS, I get a "Segmentation fault (core dumped)" with the openssl.Linux.x86_64

Any idea? LIS is based on Ubuntu

Thanks

teward:
Bash on Windows requires you to compile the OpenSSL fork yourself - it isn't the exact same as the environment the static builds were made for and has been an issue in the past.  The downside is you lose a lot of the ciphers/curves this way.  You may wish to just use the bundled OpenSSL that comes with the Linux subsystem.

marcstern:
Thanks for the quick reply.
Any idea about the reasons/limitation why a standard Linux bin cannot be 
used under LIS?

teward:
Probably has to do with the libraries and what is available.  I have not had to work with it on the linux subsystem for windows but in most cases it would be missing shared libraries, etc. that aren't bundled in WSL but are available in the full Ubuntu installs.

drwetter:
gdb program core ;-)

marcstern:
can't find it under LIS :-(
    
    
      
    
    On 01-02-2017 22:15, Dirk Wetter wrote:
    
    
      
      gdb program core ;-)
      —
        You are receiving this because you authored the thread.
        Reply to this email directly, view
          it on GitHub, or mute
          the thread.
      
        
          
          
          
          
        
        
        
      
      {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/drwetter/testssl.sh","title":"drwetter/testssl.sh","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/drwetter/testssl.sh"}},"updates":{"snippets":[{"icon":"PERSON","message":"@drwetter in #605: gdb program core ;-)"}],"action":{"name":"View Issue","url":"https://github.com/drwetter/testssl.sh/issues/605#issuecomment-276784420"}}}

drwetter:
@teward  Is there a variable one can use to prevent testssl's binary to be picked up?

[...]

PR #620, opened 2017-02-07
Attempt to force system binaries for WSL

teward:
This is an attempt to force the use of the system binaries for OpenSSL if on WSL.  (See #605)

The evil thing is, this (testssl.sh) is *slow* on WSL.  Not sure why, but it is...

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -10769,10 +10769,12 @@ find_openssl_binary() {
           outln " Looking some place else ..."
      elif [[ -x "$OPENSSL" ]]; then
           :    # 1. all ok supplied $OPENSSL was found and has excutable bit set -- testrun comes below
+     elif [ -e "/mnt/c/Windows/System32/bash.exe" ] && test_openssl_suffix "$(dirname "$(which openssl)")"; then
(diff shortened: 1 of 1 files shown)


teward:
Some notes:

It *does* run slow, I'm running a test on google.com in my Windows 10 + Linux Subsystem VM on my hypervisors to get a better idea, but it's properly using the system binaries here.

teward:
With this commit, on my VM, which has 3GB RAM and 4 vCPUs (each at 3.199GHz) available to it, possibly due to the limitations on CPU usage being enforced by my vSphere settings, a standard run of `time ./testssl.sh google.com` returns the following run time:

    real    34m29.453s
    user    0m15.203s
    sys     34m18.781s


I mean, I knew WSL was slow, but... geez.  Anyways, this 'works' though it's not speedy.

drwetter:
Thank you. Mind the double square brackets, so that it is bash internal. 
'[' could be just a binary, try ``which [``

teward:
@drwetter Yep, my fault.  Oopsies.  (Typos are evil aren't they)

drwetter:
Fault? c'mon.. Thx for the PR. :)

⁣-- 

Issue #623, opened 2017-02-07
Formatting for BEAST needs proper wrapping


drwetter:
First four lines have 2-3 ciphers and are properly indented, the last line is not wrapped or indented

![screenshot_20170207_230201](https://cloud.githubusercontent.com/assets/8036727/22713602/7c25b332-ed89-11e6-8ed3-ac94138b3e18.png)

(the first 4 lines can as well be a bit broader)

drwetter:
Maybe this is the chance to write a proper function which takes a string, the # of spaces for indentation and at which line something should be (word) wrapped

dcooper16:
I'm no expert with `sed`, `awk`, etc., but this seems to work:

```
out_row_aligned_max_width() {
     local text="$1"
     local spaces="$2"
     local -i max_width="$3"
     local -i i len cut_point
     local cr=$'\n'
     local formatted_text=""

     max_width=$max_width-${#spaces}
     len=${#text}
     while true; do
          i=$max_width
          if [[ $i -ge $len ]]; then
               i=$len
          else
               while true; do
                    [[ "${text:i:1}" == " " ]] && break
                    [[ $i -eq 0 ]] && break
                    i=$i-1
               done
               if [[ $i -eq 0 ]]; then
                    i=$max_width+1
                    while true; do
                         [[ "${text:i:1}" == " " ]] && break
                         [[ $i -eq $len ]] && break
                         i+=1
                    done
               fi
          fi
          if [[ $i -eq $len ]]; then
               if [[ -z "$formatted_text" ]]; then
                    formatted_text="$text"
               else
                    formatted_text+="${cr}${spaces}${text}"
               fi
               break
          else
               if [[ -z "$formatted_text" ]]; then
                    formatted_text="${text:0:i}"
               else
                    formatted_text+="${cr}${spaces}${text:0:i}"
               fi
               len=$len-$i-1
               i=$i+1
               text="${text:i:len}"
               [[ $len -eq 0 ]] && break
          fi
     done
     out "$formatted_text"
     return 0
}
```

drwetter:
see commit above

PR #628, opened 2017-02-09
Wrap long lines

dcooper16:
This PR addresses the issue raised in #623. This PR is based on the function `out_row_aligned_max_width()` that I proposed in #623, but the `out_row_aligned_max_width()` in this PR is a little different. It takes a fourth parameter, which is the function to use to print each word in the text string to be printed. This is used in `run_pfs()` so that the "Elliptic curves offered" can be printed using this function (some servers support 25 curves), while still having the curves printed using color-coding to indicate the quality of each curve.

I somewhat arbitrarily have each line wrap at 120 characters, but that could be changed (e.g., to `$TERM_WIDTH`).

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -975,6 +975,64 @@ out_row_aligned() {
      done
 }
 
+# prints text over multiple lines, trying to make no line longer than $max_width.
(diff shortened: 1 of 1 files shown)


drwetter:
awesome! Thx

PR #630, opened 2017-02-13
Option to show RFC cipher names

dcooper16:
When a list of cipher suites is being displayed using `neat_list()`, testssl.sh shows the cipher suite's OpenSSL name and (in most cases) the RFC name as well. However, in all other cases only the OpenSSL name is shown.

This PR adds the option to have cipher suite's RFC names shown instead of the OpenSSL name, by including `--mapping rfc` in the command line. [Note: if the cipher-mapping.txt file cannot be found, then the `--mapping rfc` option is ignored and the OpenSSL names are shown.]

This PR seems to be related to issue #9, but #9 may be been referring to the output created by `neat_list()`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -229,6 +229,7 @@ HAS_PROXY=false
 HAS_XMPP=false
 HAS_POSTGRES=false
 ADD_RFC_STR="rfc"                       # display RFC ciphernames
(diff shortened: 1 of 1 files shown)


drwetter:
great you realized this old issue.

Thanks for this PR, again!

PR #632, opened 2017-02-14
Negotiated cipher per proto bugfix

dcooper16:
I have a test server that I configured to support only SSLv3 and TLSv1.2. When I set `SSLHonorCipherOrder` to `off` I get the following results:
```
 Negotiated cipher per proto  (limited sense as client will pick)
 (SPDY is an HTTP protocol and thus not tested here)
     ECDHE-RSA-AES256-SHA:          SSLv3     ECDHE-RSA-AES256-GCM-SHA384:   TLSv1.2
```
The current code, when printing TLSv1.2 checks whether `${cipher[4]}` is empty, and since it is assumes no previous protocol (SSLv2, SSLv3, TLSv1, TLSv1.1) was supported, and so doesn't output a newline before outputting the cipher and protocol for TLSv1.2.

This PR fixes that by changing to code to look at the previous non-empty cipher (if there is one), even if that does not come from the previous protocol. It also rearranges the code a bit so that the code for printing both the cipher and protocol doesn't need to be repeated.


drwetter:
thank you

Issue #619, opened 2017-02-07
Output HTML


dcooper16:
> > So, at the moment I can't think of a good way to make the available and unavailable ciphers easily distinguishable that works with HTML (as generated by aha) as well as in terminals with black backgrounds and terminals with white backgrounds.
> 
> As far as HTML is concerned: aha is a cool thing -- however I still consider it as a kludge -- a good working one though. My idea like two years back was to do this natively or via JSON-->HTML converter. Now as JSON is there I have my doubts that this is going to fly as the ideas of both are different: one is for humans, the other one is more a machine readable format. Conversion thus seems too complicated and HTML output would probably be easier if done natively by testssl.

I decided to look into getting testssl.sh to create HTML output, and I believe that I'm mostly there: https://github.com/dcooper16/testssl.sh/blob/generate_html/testssl.sh.

My basic idea was to change functions like `out()`, `pr_warning()`, etc. to output the input to the terminal and, optionally, to an HTML file. In order for this to work I needed to ensure that these functions were only called for output that should go to both the terminal and the HTML file. However, in some cases testssl.sh currently calls one of these functions and "captures" the output in a variable. So, I had to modify the code so that none of these functions were called for that purpose.

I also had to change the code a bit to ensure that all output that was intended for both the terminal and the HTML file was printed using one of these functions (e.g., not using `echo`).

As a general rule I assumed that debugging output should only be displayed on the terminal, so I call separate functions (e.g., `pr_warning_term()`) when the output should be displayed on the terminal but not in the HTML.

The code seems to be working okay, but needs more testing. The main problem at the moment is that the caller needs to provide the name of the file in which to output the HTML, and testssl.sl will append to that file. So, if you run testssl.sh multiple times and want the output to go to the same file name, you may have to delete the file yourself before each call.

As noted, the code is available at https://github.com/dcooper16/testssl.sh/tree/generate_html (just add an `--htmlfile <filename>` option to the command line). Please let me know if you think this is the right direction to go.

Thanks,

David

drwetter:
👍  excellent! Ran it, looks good.

Let me have a look at the code. Call it a day now...

dcooper16:
I've done some more work on this code over the past week and believe that it is ready to go.

The main issue at the moment is that all the output is sent to a single file, the one specified in the command line. This should work with the current code, but it would not work with `run_mass_testing_parallel()`. If that function were in use, then there would be no choice but to have the HTML for each server go into a separate file.

[...]

PR #634, opened 2017-02-14
Generate html

dcooper16:
As described in #619, this PR adds the option for testssl.sh to create HTML. The option may be invoked using either `--html` or `--htmlfile <filename>`. If the latter is used, "\<filename\>" may be the name of a file or a directory.

If a file name is provided, then all of the output is sent to that file, and the HTML output looks very similar to what is displayed in the terminal. If a file name is not provided and mass testing is being performed, then the output for each line in the file provided to `--file` is placed in a separate HTML file.

The colors that are used in the HTML output are specified in the `pr_...()` functions and in `emphasize_stuff_in_headers()`. As the choice of colors is subjective, this could certainly use some minor adjustments. In some cases I tried to closely match the colors that appear in the terminal. However, that isn't always possible since testssl.sh creates different colors in different terminals.

One issue to be noted is that when output is created using `out_row_aligned_max_width()`, the width used for the HTML output is `$TERM_WIDTH`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -158,6 +158,7 @@ LOGFILE=${LOGFILE:-""}                  # logfile if used
 JSONFILE=${JSONFILE:-""}                # jsonfile if used
 CSVFILE=${CSVFILE:-""}                  # csvfile if used
 HTMLFILE=${CSVFILE:-""}                 # HTML if used
+HTMLHEADER=true                         # include HTML headers and footers in HTML file, if one is being created
 APPEND=${APPEND:-false}                 # append to csv/json file instead of overwriting it
 GIVE_HINTS=false                   # give an addtional info to findings
 HAS_IPv6=${HAS_IPv6:-false}             # if you have OpenSSL with IPv6 support AND IPv6 networking set it to yes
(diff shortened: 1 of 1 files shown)


drwetter:
see 8d66786e42cc296937ba6b9bbdfe73d9be8209e2 . More later

drwetter:
Thanks a lot for the PR, @dcooper16!
 
Work will continue on the branch 2.9dev_html.

Already changed:
* removed HTMLHEADER. We always want that (in fact for flat JSON this is missing and needs to be added)
* changing of names (out/pr/term). They were carrying redundant information sometimes (pr_*_term ) and it was anyway maybe a good idea to start with that before.
* some formatting for readability


Open points and some ideas:
* italic in ``testssl.sh -S`` doesn't work in the terminal
* HTMLHEADER: does that work with --file. If not I rather propose to handle this like JSON/CSV.
* the former sed statement around L1900 for the header was way more readable. The combined
  html+terminal version is just too much. First idea: Maybe a switch whether HTML is requested is good is better so that this can be separated. Maybe is worth to try to locally define color variables with either HTML like ``html_yellow=" <span style=\"color:AAA;font-weight:bold\">`` but one needs to be careful with sed, it might hiccup and BSD's sed is different to GNU's.
* if we're not sed'ing the color with the variable it's anyway good keeping some repeated markups in a local variable.
* what went into main here is too much. Actually what I put already in there bothered
  me as too much logic and not obvious dependencies are in here. Now it's worse :-)
  Can't this be just similar to JSON or CSV -- a separate function with hooks
  not in main()?
* minor thing: TERM_WIDTH is for HTML is maybe not the best. But that can be
  tackled later

PR #635, opened 2017-02-14
run_protocols() bug fix

dcooper16:
Since the test for TLS 1.2 in `run_protocols()` now uses `tls_sockets()` whenever `$ssl_native` is `true` (i.e., there is no longer a requirement for `$EXPERIMENTAL` to be true as well), the `$EXPERIMENTAL` flag should no longer be checked if the return value is 1.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4506,7 +4506,7 @@ run_protocols() {
                ;;                                  # GCM cipher in TLS 1.2: very good!
           1)
                pr_svrty_mediumln "not offered"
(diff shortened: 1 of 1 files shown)


dcooper16:
I added a second bug fix.

Unlike in the TLS 1 and TLS 1.1 tests, in the TLS 1.2 test, if the server failed the connection rather than downgrading, but it should have downgraded, then the "not offered" and error messages were displayed on separate lines.

drwetter:
thx

PR #637, opened 2017-02-17
Printing Negotiated cipher

dcooper16:
`run_server_preference()` prints out the server's Negotiated cipher in a different color depending on the quality of the cipher. However, there is a "FIXME" since CBC ciphers are supposed to be flagged, but it is not easy to identity all CBC ciphers from their OpenSSL names.

This PR partially addresses this. It creates a separate function for printing a cipher based on its quality. Whenever possible it determines the quality of the cipher based on the RFC name. However, if it is provided an OpenSSL name and no cipher-mapping.txt file is available, it will follow the current (imperfect) logic for determining the cipher's quality.

The function also returns a value that indicates the quality of the cipher provided, with higher numbers indicating better ciphers. This return value is used by `run_server_preference()` to determine how to populate the "severity" field when calling `fileout()`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4660,6 +4660,80 @@ pr_ecdh_curve_quality() {
      fi
 }
 
+# Print $2 based on the quality of the cipher in $1. If $2 is empty, print $1.
+# The return value is an indicator of the quality of the cipher in $1:
+#   0 = $1 is empty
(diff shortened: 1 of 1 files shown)


drwetter:
Good you took this on, thank you!

PR #639, opened 2017-02-17
OCSP must staple

dcooper16:
[RFC 7633](https://tools.ietf.org/html/rfc7633) introduces the TLS Features certificate extension, which contains "Features":
> The object member "Features" is a sequence of TLS extension identifiers (features, in this specification's terminology) as specified in the IANA Transport Layer Security (TLS) Extensions registry.  If these features are requested by the client in its ClientHello message, then the server MUST return a ServerHello message that satisfies this request.

The main purpose of this certificate extension is to implement "must staple." If the extension is present in a TLS server's certificate and it includes status_request, then the server MUST include a stapled OCSP response if the client requests one. (The same applies for the status_request_v2 extension.)

This PR adds a check to `certificate_info()` of whether the server supports "must staple" (i.e., whether its certificate includes a TLS Features extension with "status_request"). It also changes the output for "OCSP stapling" in the case that the server did not staple an OCSP response. It indicates that:
* it is a critical issue if the certificate specifies "must staple"
* it is a low severity issue if the certificate does not specify "must staple," but the certificate does include an OCSP URI.
* it is not an issue at all if the certificate does not specify "must staple" and certificate does not include an OCSP URI.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -5894,6 +5894,50 @@ compare_server_name_to_cert()
      return $ret
 }
 
+must_staple() {
+     local json_prefix="$2"
+     local cert extn
+     local -i extn_len
+     local supported=false
+     
+     # Note this function is only looking for status_request (5) and not
+     # status_request_v2 (17), since OpenSSL seems to only include status_request (5)
+     # in its ClientHello when the "-status" option is used.
+     
+     # OpenSSL 1.1.0 supports pretty-printing the "TLS Feature extension." For any
+     # previous versions of OpenSSL, OpenSSL can only show if the extension OID is present.
+     if $OPENSSL x509 -in "$HOSTCERT" -noout -text 2>>$ERRFILE | grep -A 1 "TLS Feature:" | grep -q "status_request"; then
+          # FIXME: This will indicate that must staple is supported if the
+          # certificate indicates status_request or status_request_v2. This is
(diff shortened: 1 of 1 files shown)


dcooper16:
I should note that I haven't yet found a certificate that includes the TLS Features extension. However, https://community.qualys.com/thread/16290 indicates that Let's Encrypt supports it and that https://dev.ssllabs.com now includes a test for must staple.

drwetter:
~2 weeks ago I migrated from startssl to LE (testssl.sh). I missed that !

⁣-- 

dcooper16:
As was noted at https://community.qualys.com/thread/16290, Chrome isn't currently planning to support must staple, and as can be seen at https://groups.google.com/a/chromium.org/forum/#!msg/security-dev/-pB8IFNu5tw/5CEbCaKTGwAJ, at least one commenter believes that support for OCSP isn't very good at the moment and so implementing must staple could break things. I don't know if the commenter is just being overly pessimistic, but it's something to think about before requesting this extension in a TLS server certificate.

The main reason that I implemented this test is I wanted to see if it made sense to recommend inclusion of this extension in TLS server certificates, and it seemed that a good starting point was to find out if anyone was actually using it yet.

drwetter:
Thank you!

Anybody with a site which has the "must staple" extension?

PR #642, opened 2017-02-21
Wrap long lines in display_rdns_etc()

dcooper16:
Some sites have a long list of IP addresses and some IP addresses have a long list of DNS names that map to them.

This PR changes `display_rdns_etc()` to use `out_row_aligned_max_width()` to print the other IP addresses in `$IP46ADDRs` and to print `$rDNS`.


drwetter:
Thank you

Issue #643, opened 2017-02-21
testssl.sh fails at LOGJAM test


shoopdawoop:
I am seeing this when testing a server since after I changed my cipher settings (perhaps I have misconfigured nginx?).

The script runs fine up to this point, throws this and then quits:

```
LOGJAM (CVE-2015-4000), experimental      head: illegal line count -- -1
./testssl.sh: line 10236: 1
2
3
4
5
[...count goes on up to...]
371
372: syntax error in expression (error token is "2
3
4
5
[...count goes on up to...]
371
372")

user$
```

The nginx directive I set:

```
ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
ssl_ciphers 'ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:!DSS';
ssl_prefer_server_ciphers on;
```

My fault or should the script catch this?

teward:
That looks like an ugly error in the script; what OS are you running the script on?  And what server are you testing against?  As well, which branch of testssl.sh are you using?

bknowles:
This is exactly the same error that I had found while working on Issue #638 but had not yet gotten around to filing another bug report on.

bknowles:
[...]

PR #644, opened 2017-02-21
Fix issue #643

dcooper16:
It seems that the head command on OS X does not accept a negative number as a value for the "-n" parameter. This PR provides an alternative method for removing the "generator:  " line without using "head."


dcooper16:
Following @bknowles suggestion, I just replaced

```tail -n +2 | head -n -1```

with

```egrep -v "prime|generator"```

drwetter:
Thx David!

Issue #631, opened 2017-02-14
RFC names in wide output / -e , -E


drwetter:
``neat_header()`` and ``neat_list()`` need to be changed too after #630 .

drwetter:
I think the switches for ``--mapping`` should be clarified like:

rfc-only | rfc ==> displays only RFC  names in output
openssl-only | openssl ==>  displays only openssl names in output
both  ==> default

bknowles:
👍

dcooper16:
I tried to address this issue with PR #646.

I don't believe having three options of "rfc-only", "openssl-only", and "both" works. These three options might work for information that is printed in wide mode, but not in other places (e.g., client simulation, pfs, negotiated cipher, cipher order). In these places, testssl.sh needs to display either the OpenSSL name or the RFC name.

That's why #630 only added an "rfc" choice. I wanted a way to print the RFC name rather than the OpenSSL name in places where testssl.sh could only print one or the other, but I don't see a problem with just printing both in wide mode, when both can be printed.

By providing 4 choices for the `--mapping` option (including the default one), the user can not only request that only the OpenSSL name or only the RFC name be printed, but can request that both be printed in wide mode while choosing which name type should be used in places where only one or the other can be printed.

drwetter:
see #646

PR #646, opened 2017-02-23
Add option to print only the RFC cipher names

dcooper16:
This PR attempts to address #631. It allows four choices for the `--mapping` option. "openssl" (default), "rfc", "no-openssl", and "no-rfc".

* "openssl" is the current default

* "no-rfc" shows only the OpenSSL names (just as it currently does)

* "rfc" shows the RFC name rather than the OpenSSL name for things that are not in wide now (just as it currently does). But now, in wide mode, it shows the RFC name first (further to the left) and the OpenSSL name second.

* "rfc-only" shows only the RFC name.

If the `--mapping` option is not provided, this is the same as "openssl". If the "cipher-mapping.txt" file cannot be found, then testssl.sh runs as if "no-rfc" had been requested.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -227,8 +227,7 @@ HAS_FALLBACK_SCSV=false
 HAS_PROXY=false
 HAS_XMPP=false
 HAS_POSTGRES=false
-ADD_RFC_STR="rfc"                       # display RFC ciphernames
-SHOW_RFC=""                             # display RFC ciphernames instead of OpenSSL ciphernames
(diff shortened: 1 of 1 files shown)


drwetter:
Good work, thanks!

Other than what you described "rfc-only" doesn't exist, it's equal to no-openssl. I was about to add this as an equivalent to the command line but it would confuse more than it helps. Then openssl-only would need to be an equivalent to "no-rfc"

PR #647, opened 2017-02-24
Add missing herestring redirect (<<<) in find_openssl_binary function

gniltaws:
With the latest pull, I kept getting a `tr` error when I ran testssl.sh:
`$ ./testssl.sh -q -U google.com
tr: extra operand ‘1.0.2-chacha’
Only one string may be given when deleting without squeezing repeats.
Try 'tr --help' for more information.`
I tracked it down to a missing `<<<` in the `find_openssl_binary` function.

I also removed trailing whitespace since that is a pet peeve of mine.


drwetter:
Thanks

PR #652, opened 2017-02-28
Fix HTML headers and footers

dcooper16:
This PR ensures that each HTML file produced by testssl.sh only includes a single header, at the top, and a single footer, at the end. It also tries to ensure that the short-version banner is only placed at the top of the HTML file if (1) mass testing is being performed and (2) the results of each test is being placed in a separate file.

It also moves some of the logic out of main and into `html_header()`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -160,6 +160,7 @@ LOGFILE=${LOGFILE:-""}                  # logfile if used
 JSONFILE=${JSONFILE:-""}                # jsonfile if used
 CSVFILE=${CSVFILE:-""}                  # csvfile if used
 HTMLFILE=${HTMLFILE:-""}                # HTML if used
+HTMLHEADER=true                         # include HTML headers and footers in HTML file, if one is being created
 APPEND=${APPEND:-false}                 # append to csv/json file instead of overwriting it
 GIVE_HINTS=false                        # give an addtional info to findings
 HAS_IPv6=${HAS_IPv6:-false}             # if you have OpenSSL with IPv6 support AND IPv6 networking set it to yes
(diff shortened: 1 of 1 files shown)


drwetter:
thx: does that have the same prob as #653 ?

dcooper16:
> thx: does that have the same prob as #653 ?

No. I tried this out in many different scenarios, and it seemed to work in all of them.

The logic that I used is very different from what is used for the JSON and CSV output. In my code, the call to (possibly) insert the header is made at the very beginning. So, if the `--file` option is being used and all the output is being placed in a single file, the "parent" call to testssl.sh inserts the header and footer, whereas with JSON it is the first "child" call to testssl.sh that inserts the footer.

It seems that the reason the JSON file gets two footers is that `fileout_footer()` is called by both `run_mass_testing()` and `cleanup ()`, and both insert a footer. When each "child" creates its own JSON file, the `"$APPEND" || fileout_footer` logic prevents a footer from being added.

The logic I used is a follows. `html_header()` is called at the very beginning, and it uses this logic to determine whether to add a header:
```
     # Don't create HTML headers and footers in the following scenarios:
     #  * HTML output is not being created.
     #  * mass testing is being performed and each test will have its own HTML file.
     #  * this is an individual test within a mass test and all HTML output is being placed in a single file.
     if ! "$do_html" || \
        ( "$do_mass_testing" && ( [[ -z "$HTMLFILE" ]] || [[ -d "$HTMLFILE" ]] ) ) || \
        ( "$APPEND" && [[ -n "$HTMLFILE" ]] && [[ ! -d "$HTMLFILE" ]] ); then
          HTMLHEADER=false
          return 0
     fi
```

`html_footer()` is called by `cleanup ()` (and only by `cleanup ()`, and it adds a footer if and only if a header was added (i.e., `$HTMLHEADER` is `true`).

In the logic in `html_header()` the second line says that the "parent" call to testssl.sh should not create headers and footers if the `--file <testfile>` is used and each line of "testfile" will have its own HTML file (each "child" call with create headers and footers for its file). The third line says that if this is a "child" call (`$APPEND` is `true`) and the output of all the "children" is being placed in a single file, then don't create a header or footer (since the "parent" will create the header and footer for the file).

The `HTMLHEADER` is set to indicate to `html_footer()` whether or not to add a footer, since a footer should be added if and only if a header is added.

drwetter:
👍

PR #661, opened 2017-03-15
Speedup normalize_ciphercode()

dcooper16:
While doing some performance testing I discovered that `normalize_ciphercode()` is very slow. This PR simplifies the function and speeds it up significantly. This PR also addresses the TODO item in `normalize_ciphercode()` by eliminating the global variable HEXC.


drwetter:
Thank you

PR #663, opened 2017-03-15
Use -s option with OpenSSL ciphers

dcooper16:
With OpenSSL 1.1.0 (and maybe other versions), the `ciphers` function lists many cipher suites that are not actually supported by the `s_client` option. This PR fixes that by using the `-s` option whenever `$OPENSSL ciphers` is used to obtain a list of cipher suites supported by OpenSSL. According to https://www.openssl.org/docs/manmaster/man1/ciphers.html:
```
-s

    Only list supported ciphers: those consistent with the security level, and minimum and
    maximum protocol version. This is closer to the actual cipher list an application will
    support.
```
When the `-s` option is used along with `-tls1`, OpenSSL 1.1.0 will not list any ciphers that only work with TLSv1.2. So, `prepare_debug()` needed to be changed to correctly populate `ossl_supported_tls`, which is supposed to be a list of all non-SSLv2 ciphers supported by the server.

LibreSSL issues an "unknown option" error if the `-s` option is provided, so the `-s` option is only included in the command line if `$OPENSSL` has been determined to support it.

This PR is needed so that `prepare_debug()` can correctly determine which cipher suites are or are not supported by `$OPENSSL`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -349,6 +349,7 @@ HAS_IRC=false
 HAS_CHACHA20=false
 HAS_AES128_GCM=false
 HAS_AES256_GCM=false
+OSSL_CIPHERS_S=""
 PORT=443                                # unless otherwise auto-determined, see below
 NODE=""
 NODEIP=""
@@ -1181,11 +1182,11 @@ actually_supported_ciphers() {
 
      [[ "$2" != "ALL" ]] && tls13_ciphers="$2"
      if "$HAS_CIPHERSUITES"; then
-          $OPENSSL ciphers $3 -ciphersuites "$tls13_ciphers" "$1" 2>/dev/null || echo ""
+          $OPENSSL ciphers $3 $OSSL_CIPHERS_S -ciphersuites "$tls13_ciphers" "$1" 2>/dev/null || echo ""
      elif [[ -n "$tls13_ciphers" ]]; then
-          $OPENSSL ciphers $3 "$tls13_ciphers:$1" 2>/dev/null || echo ""
+          $OPENSSL ciphers $3 $OSSL_CIPHERS_S "$tls13_ciphers:$1" 2>/dev/null || echo ""
      else
-          $OPENSSL ciphers $3 "$1" 2>/dev/null || echo ""
+          $OPENSSL ciphers $OSSL_CIPHERS_S $3 "$1" 2>/dev/null || echo ""
      fi
 }
 
@@ -3037,7 +3038,11 @@ openssl2hexcode() {
      local -i i
 
      if [[ $TLS_NR_CIPHERS -eq 0 ]]; then
-          hexc="$(actually_supported_ciphers 'ALL:COMPLEMENTOFALL:@STRENGTH' 'ALL' "-V" | awk '/ '"$1"' / { print $1 }')"
+          if "$HAS_CIPHERSUITES"; then
+               hexc="$($OPENSSL ciphers -V -ciphersuites "$TLS13_OSSL_CIPHERS" 'ALL:COMPLEMENTOFALL:@STRENGTH' | awk '/ '"$1"' / { print $1 }')"
+          else
+               hexc="$($OPENSSL ciphers -V 'ALL:COMPLEMENTOFALL:@STRENGTH' | awk '/ '"$1"' / { print $1 }')"
+          fi
      else
           for (( i=0; i < TLS_NR_CIPHERS; i++ )); do
                [[ "$1" == ${TLS_CIPHER_OSSL_NAME[i]} ]] && hexc="${TLS_CIPHER_HEXCODE[i]}" && break
@@ -5281,11 +5286,11 @@ listciphers() {
 
      [[ "$2" != "ALL" ]] && tls13_ciphers="$2"
      if "$HAS_CIPHERSUITES"; then
-          $OPENSSL ciphers $3 -ciphersuites "$tls13_ciphers" "$1" &>$TMPFILE
+          $OPENSSL ciphers $OSSL_CIPHERS_S $3 -ciphersuites "$tls13_ciphers" "$1" &>$TMPFILE
      elif [[ -n "$tls13_ciphers" ]]; then
-          $OPENSSL ciphers $3 "$tls13_ciphers:$1" &>$TMPFILE
+          $OPENSSL ciphers $OSSL_CIPHERS_S $3 "$tls13_ciphers:$1" &>$TMPFILE
      else
-          $OPENSSL ciphers $3 "$1" &>$TMPFILE
+          $OPENSSL ciphers $OSSL_CIPHERS_S $3 "$1" &>$TMPFILE
      fi
      ret=$?
      debugme cat $TMPFILE
@@ -11841,8 +11846,10 @@ parse_tls_serverhello() {
           else
                rfc_cipher_suite="$(show_rfc_style "x${tls_cipher_suite:0:4}")"
           fi
+     elif "$HAS_CIPHERSUITES"; then
+          rfc_cipher_suite="$($OPENSSL ciphers -V -ciphersuites "$TLS13_OSSL_CIPHERS" 'ALL:COMPLEMENTOFALL' | grep -i " 0x${tls_cipher_suite:0:2},0x${tls_cipher_suite:2:2} " | awk '{ print $3 }')"
      else
-          rfc_cipher_suite="$(actually_supported_ciphers 'ALL:COMPLEMENTOFALL' 'ALL' "-V" | grep -i " 0x${tls_cipher_suite:0:2},0x${tls_cipher_suite:2:2} " | awk '{ print $3 }')"
+          rfc_cipher_suite="$($OPENSSL ciphers -V 'ALL:COMPLEMENTOFALL' | grep -i " 0x${tls_cipher_suite:0:2},0x${tls_cipher_suite:2:2} " | awk '{ print $3 }')"
      fi
      echo "Cipher    : $rfc_cipher_suite" >> $TMPFILE
      if [[ $dh_bits -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


drwetter:
I needed some time to research this myself and still I am lost a bit here, 
It seems you missed a part of the explanation: 

```
   PSK and SRP ciphers are not enabled by default: they require -psk  or -srp to enable them.
```

If I diff the output of ``ciphers -V`` and ``ciphers -V -s`` for 1.1.x it seems that in fact drops the SRP and PSK ciphers.

Are you suggesting not to test those anymore?

dcooper16:
Hi Dirk,

I did miss the part about PSK and SRP ciphers only being enabled if the `-psk` and `-srp` flags are used, but I don't understand what it means. What we really want to know is what ciphers are supported by `s_client`, and I don't know what (if any) flags are needed for `s_client` to support the PSK and SRP ciphers. I just tried running `openssl ciphers -V -psk -srp -s ALL:COMPLEMENTOFALL:@STRENGTH` and then using only ciphers that it lists in `s_client`:
```
openssl s_client -connect 127.0.0.1:443 --cipher RSA-PSK-AES256-GCM-SHA384:SRP-RSA-AES-256-CBC-SHA:AES256-GCM-SHA384
```
I looked at the resulting ClientHello (using ssldump), and it only included AES256-GCM-SHA384.

The reason that I submitted this PR, however, had nothing to do with PSK or SRP ciphers. The issue I encountered had to do with other ciphers. Using a version of OpenSSL 1.1.0e compiled to support as any ciphers as possible, this is what I get:
```
cmd 1> openssl ciphers -V ALL:COMPLEMENTOFALL:@STRENGTH > l1.txt
cmd 2> openssl ciphers -V -s ALL:COMPLEMENTOFALL:@STRENGTH > l2.txt
cmd 3 > diff l1.txt l2.txt | grep "<" | grep -v PSK | grep -v SRP 

<     0x00,0xA7 - ADH-AES256-GCM-SHA384   TLSv1.2 Kx=DH       Au=None Enc=AESGCM(256) Mac=AEAD
<     0x00,0x6D - ADH-AES256-SHA256       TLSv1.2 Kx=DH       Au=None Enc=AES(256)  Mac=SHA256
<     0x00,0xC5 - ADH-CAMELLIA256-SHA256  TLSv1.2 Kx=DH       Au=None Enc=Camellia(256) Mac=SHA256
<     0xC0,0x19 - AECDH-AES256-SHA        SSLv3 Kx=ECDH     Au=None Enc=AES(256)  Mac=SHA1
<     0x00,0x3A - ADH-AES256-SHA          SSLv3 Kx=DH       Au=None Enc=AES(256)  Mac=SHA1
<     0x00,0x89 - ADH-CAMELLIA256-SHA     SSLv3 Kx=DH       Au=None Enc=Camellia(256) Mac=SHA1
<     0x00,0xA6 - ADH-AES128-GCM-SHA256   TLSv1.2 Kx=DH       Au=None Enc=AESGCM(128) Mac=AEAD
<     0x00,0x6C - ADH-AES128-SHA256       TLSv1.2 Kx=DH       Au=None Enc=AES(128)  Mac=SHA256
<     0x00,0xBF - ADH-CAMELLIA128-SHA256  TLSv1.2 Kx=DH       Au=None Enc=Camellia(128) Mac=SHA256
<     0xC0,0x18 - AECDH-AES128-SHA        SSLv3 Kx=ECDH     Au=None Enc=AES(128)  Mac=SHA1
<     0x00,0x34 - ADH-AES128-SHA          SSLv3 Kx=DH       Au=None Enc=AES(128)  Mac=SHA1
<     0x00,0x9B - ADH-SEED-SHA            SSLv3 Kx=DH       Au=None Enc=SEED(128) Mac=SHA1
<     0x00,0x46 - ADH-CAMELLIA128-SHA     SSLv3 Kx=DH       Au=None Enc=Camellia(128) Mac=SHA1
<     0xC0,0x16 - AECDH-RC4-SHA           SSLv3 Kx=ECDH     Au=None Enc=RC4(128)  Mac=SHA1
<     0x00,0x18 - ADH-RC4-MD5             SSLv3 Kx=DH       Au=None Enc=RC4(128)  Mac=MD5 
<     0x00,0x04 - RC4-MD5                 SSLv3 Kx=RSA      Au=RSA  Enc=RC4(128)  Mac=MD5 
<     0xC0,0x17 - AECDH-DES-CBC3-SHA      SSLv3 Kx=ECDH     Au=None Enc=3DES(168) Mac=SHA1
<     0x00,0x1B - ADH-DES-CBC3-SHA        SSLv3 Kx=DH       Au=None Enc=3DES(168) Mac=SHA1
<     0xC0,0x06 - ECDHE-ECDSA-NULL-SHA    SSLv3 Kx=ECDH     Au=ECDSA Enc=None      Mac=SHA1
<     0xC0,0x10 - ECDHE-RSA-NULL-SHA      SSLv3 Kx=ECDH     Au=RSA  Enc=None      Mac=SHA1
<     0xC0,0x15 - AECDH-NULL-SHA          SSLv3 Kx=ECDH     Au=None Enc=None      Mac=SHA1
<     0x00,0x3B - NULL-SHA256             TLSv1.2 Kx=RSA      Au=RSA  Enc=None      Mac=SHA256
<     0x00,0x02 - NULL-SHA                SSLv3 Kx=RSA      Au=RSA  Enc=None      Mac=SHA1
<     0x00,0x01 - NULL-MD5                SSLv3 Kx=RSA      Au=RSA  Enc=None      Mac=MD5
```

I was testing against a server that supports RC4-MD5, and yet OpenSSL wasn't detecting it. This isn't a big problem at the moment. I believe that the only issue is that if one tested this server using `--ssl-native` and `--show-each`, the output would suggest that RC4-MD5 was tested for and it was found that the server doesn't implement it, but it wasn't actually tested.

Where this was really a problem for me was in the `cipher_pref_check()` function in my [save_time](https://github.com/dcooper16/testssl.sh/tree/save_time) branch. At first, I wrote the code so that if server only supported ciphers supported by `$OPENSSL`, then `cipher_pref_check()` would just use `$OPENSSL`. However, if I used OpenSSL 1.1.0 and tested this server that supported RC4-MD5, then RC4-MD5 wasn't showing up in the list. I fixed this by checking whether `$OPENSSL` had actually found all of the ciphers supported by the server, and then starting over with `tls_sockets()` if it hadn't. But, the code would work better if the information in the `TLS_CIPHER_OSSL_SUPPORTED` array corrected indicated which ciphers `s_client` did and did not support.

drwetter:
Hi David,

> I did miss the part about PSK and SRP ciphers only being enabled if the -psk and -srp flags are used, but I don't understand what it means.

I am confused by this too. :-) But I am looking more into the direction of openssl here... And I didn't realize the ADH + RC4 part though (see below)

> [..} to support the PSK and SRP ciphers. I just tried running openssl ciphers -V -psk -srp -s ALL:COMPLEMENTOFALL:@STRENGTH and then using only ciphers that it lists in s_client: [..] I looked at the resulting ClientHello (using ssldump), and it only included AES256-GCM-SHA384.

Isn't here -psk or -srp missing?

For the rest I need to look deeper into it to as it seems another oddity -- at least not by looking at ciphers.pod in the openssl tree.

dcooper16:
Hi Dirk,

> Isn't here -psk or -srp missing?

No. Adding `-psk` or `-srp` to `s_client` doesn't seem to work:
```
cmd 1> ./openssl version
OpenSSL 1.1.0e  16 Feb 2017

cmd 2> ./openssl s_client -srp

s_client: Option unknown option -srp
s_client: Use -help for summary.

cmd3 > ./openssl s_client -psk
s_client: Option -psk needs a value
s_client: Use -help for summary.
```
https://www.openssl.org/docs/manmaster/man1/s_client.html doesn't mention anything about SRP, and for PSK the options are:
```
-psk_identity identity

    Use the PSK identity identity when using a PSK cipher suite.
-psk key

    Use the PSK key key when using a PSK cipher suite. The key is given as a hexadecimal number without leading 0x, for example -psk 1a2b3c4d.
```

I've been thinking about this some more, and it seems that OpenSSL cannot be used to test for SRP and PRK cipher suites.

According to [RFC 5054](https://www.rfc-editor.org/rfc/rfc5054.txt), _Using the Secure Remote Password (SRP) Protocol for TLS Authentication_, the ClientHello needs to include a user name in a SRP extension. If the SRP extension is missing or if "the server doesn't have a verifier for the user name in the SRP extension," the server has to abort the handshake. If the SRP is provided with an unknown user name, the server _may_ choose to respond to the initial ClientHello as if the user name was recognized, and then later reject the client's finished message, but that's not a requirement. If the server did choose to do this, then `tls_sockets()` would be able to determine what SRP cipher suites the server supports (if we modified `socksend_tls_clienthello()` to include an SRP extension with a made-up user name). However, it seems that OpenSSL canot not be used to test for SRP cipher suites, since OpenSSL cannot not successfully connect to a server with an SRP cipher suite without a valid user name and password for the server.

According to [RFC 4279](https://www.rfc-editor.org/rfc/rfc4279.txt), _Pre-Shared Key Ciphersuites for Transport Layer Security (TLS)_, testing for PSK cipher suites with `tls_sockets()` should be okay, since the ClientHello just needs to advertise PSK cipher suites in order for the subsequent ServerHello to specify a PSK cipher suite, and then `tls_sockets()` can just ignore the contents of the ServerKeyExchange message. However, it seems that cannot be used to test for PSK cipher suites, since OpenSSL cannot successfully connect to a server with a PSK cipher suite without a valid PSK identity and key (which can be provided to `s_client` as command line arguments).

So, I thinks it's good that `openssl ciphers -V -s ALL:COMPLEMENTOFALL:@STRENGTH` doesn't list any PSK or SRP ciphers, since I don't think testssl.sh can use OpenSSL to test for these ciphers.

dcooper16:
> And I didn't realize the ADH + RC4 part though

Actually, it's the MD5 part. RC4-SHA is supported.

PR #665, opened 2017-03-18
merge 2.9dev_html into 2.9dev 

drwetter:


drwetter:
A big thx, @dcooper16

PR #669, opened 2017-03-22
Cleanup extraction of TLS extensions

dcooper16:
Currently there is code to extract TLS extensions in three places, in `get_server_certificate()` and two places in `determine_tls_extensions()`. This PR replaces them with one new function, `extract_new_tls_extensions()`.

In order for the new function to work correctly whether OpenSSL or `tls_sockets()` is being used, this PR also changes `parse_tls_serverhello()` so that extensions are formatted in the file it creates in the same way as they are formatted by OpenSSL.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -5075,6 +5075,30 @@ sclient_connect_successful() {
      return 1
 }
 
+extract_new_tls_extensions() {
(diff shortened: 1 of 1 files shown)


drwetter:
👍

PR #671, opened 2017-03-23
Use printf to print browser names

dcooper16:
Change the client_simulation branch to print browser names using printf rather than adding extra spaces to each name. Also change update_client_sim_data.pl to not add extra spaces to names.

Note: I don't know Perl, so the change I made to update_client_sim_data.pl is just a guess. However, it seems to work.


drwetter:
Travis failed/ There seems to be an offset in the client names in t/02_client_sims_old.t .

dcooper16:
Okay, I got Travis to work. The main problem was the offset issue that was introduced on July 20, 2016, when an extra entry was added to the JSON when client simulation was being run using the `--ssl-native` option.
```
fileout "client_simulation" "WARNING" "Depending on your openssl client and the server side this may yield to false values"
```
So, I changed 4 of the test scripts to deal with the offset issue.

Once those were fixed I discovered an error in testssl.sh that affected the testing of "BingPreview Jun 2014." In particular, it was a bug in `create_client_simulation_tls_clienthello()` that would affect the sockets-based simulation of any browser that included an SNI extension in the ClientHello, but no other extensions. I submitted PRs to fix this in the master and 2.9dev branches as well (#674 and #675).

drwetter:
thank you!

PR #674, opened 2017-03-24
Fix client simulation

dcooper16:
In `create_client_simulation_tls_clienthello()` the variable `sni_extension_found` should be set if the ClientHello includes an SNI extension. Instead it was being set if and only if the ClientHello included some extension other than SNI.

This bug wasn't detected before for two reasons:

* It is rare to have a ClientHello that includes an SNI extension, but no other extensions.

* The code still works correctly if `sni_extension_found` is set even if there is no SNI in the ClientHello.

So, the bug only creates a problem if the browser's ClientHello include an SNI extension and no other extensions (see "BingPreview Jun 2014" in the client_simulation branch).


drwetter:
thx

PR #676, opened 2017-03-24
Fix two minor bugs

dcooper16:
Two of the minor bugs that were fixed by #672 in the 2.9dev branch also appear in 2.8.


drwetter:
thx

PR #677, opened 2017-03-24
Just get non-SNI certificate once

dcooper16:
Currently the call to `$OPENSSL s_client` to obtain the certificate returned by the server when SNI is not provided is in `certificate_info()`, which means that it is called once for each certificate found the various called to `get_server_certificates()`.

This PR moves the call to `$OPENSSL s_client` to `run_server_defaults()` so that the call is made only once, even if more than one server certificate was found.

In addition, in most cases the certificate returned by the server when SNI is not provided will already have been retrieved by `run_server_defaults()` (in rounds 8-14), in which case `$HOSTCERT.nosni` can just be copied from there rather than making an additional call to `$OPENSSL s_client`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -5948,8 +5948,6 @@ certificate_info() {
      fi
 
      if [[ -n "$sni_used" ]]; then
-          # no cipher suites specified here. We just want the default vhost subject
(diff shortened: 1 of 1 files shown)


drwetter:
Thx!

PR #678, opened 2017-03-24
Remove extra call to s_client

dcooper16:
Hi Dirk,

On February 21 you added an extra call to `$OPENSSL s_client` to `get_server_certificate()`:
```
     # this all needs to be moved into determine_tls_extensions()
     >$TEMPDIR/tlsext.txt
     # first shot w/o any protocol, then in turn we collect all extensions
     $OPENSSL s_client $STARTTLS $BUGS $1 -showcerts -connect $NODEIP:$PORT $PROXY $addcmd -tlsextdebug -status </dev/null 2>$ERRFILE >$TMPFILE
     sclient_connect_successful $? $TMPFILE && grep -a 'TLS server extension' $TMPFILE >$TEMPDIR/tlsext.txt
```

This was in "[collect more TLS extensions](https://github.com/drwetter/testssl.sh/commit/be079acb5eb85617868880057e00250e51e90d1a)", and it was also done in [2.8](https://github.com/drwetter/testssl.sh/commit/2df798289002deeceeda8de8ea280d303fb3f3ae).

I just realized that my recent PR (#669) ignores the results of this extra call to `$OPENSSL s_client` since it passes `$TMPFILE` to `extract_new_tls_extensions()` rather than `$TEMPDIR/tlsext.txt`.

However, I don't understand the reason that the extra call was added. I've compared the output of this PR (which removes the extra call) to 2.9dev against over 200 servers, and removing the extra call doesn't result in any extensions being missed.

I have, however, discovered one server that includes the "status request" extension in its ServerHello if SNI is not provided, but does not include it if SNI is provided, even though the same certificate is provided in both cases. That seems an unusual case, but I changed `get_server_certificate()` to only collect new TLS extensions during the first 7 rounds of calls from `run_server_defaults()`. Presumably if the server returns different extensions depending on whether or not SNI is included, testssl.sh should only show those that are returned when SNI is included.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7111,7 +7111,12 @@ get_server_certificate() {
                fi
           fi
           cp $TEMPDIR/$NODEIP.parse_tls_serverhello.txt $TMPFILE
-          extract_new_tls_extensions $TMPFILE
+
+          # When "$2" is empty, get_server_certificate() is being called with SNI="".
+          # In case the extensions returned by the server differ depending on wheter
+          # SNI is provided or not, don't collect extensions when SNI="" (unless
+          # no DNS name was provided at the command line).
+          [[ -z "$2" ]] && extract_new_tls_extensions $TMPFILE
      else
           ciphers_to_test="$1"
           if [[ "$1" =~ aRSA ]] && [[ "$1" =~ eRSA ]]; then
@@ -7131,11 +7136,6 @@ get_server_certificate() {
           [[ "${ciphers_to_test:0:1}" == : ]] &&  ciphers_to_test="${ciphers_to_test:1}"
           [[ $(count_ciphers $(actually_supported_ciphers "$ciphers_to_test")) -ge 1 ]] || return 1
 
-          # this all needs to be moved into determine_tls_extensions()
-          >$TEMPDIR/tlsext.txt
-          # first shot w/o any protocol, then in turn we collect all extensions
-          $OPENSSL s_client $STARTTLS $BUGS -cipher $ciphers_to_test -showcerts -connect $NODEIP:$PORT $PROXY $SNI -tlsextdebug -status </dev/null 2>$ERRFILE >$TMPFILE
-          sclient_connect_successful $? $TMPFILE && grep -a 'TLS server extension' $TMPFILE >$TEMPDIR/tlsext.txt
           for proto in $protocols_to_try; do
                [[ 1 -eq $(has_server_protocol $proto) ]] && continue
                [[ "$proto" == ssl3 ]] && ! "$HAS_SSL3" && continue
@@ -7143,7 +7143,6 @@ get_server_certificate() {
                $OPENSSL s_client $(s_client_options "$STARTTLS $BUGS -cipher $ciphers_to_test -showcerts -connect $NODEIP:$PORT $PROXY $SNI -$proto -tlsextdebug $npn_params -status -msg") </dev/null 2>$ERRFILE >$TMPFILE
                if sclient_connect_successful $? $TMPFILE; then
                     success=0
-                    grep -a 'TLS server extension' $TMPFILE >>$TEMPDIR/tlsext.txt
                     break               # now we have the certificate
                fi
           done                          # this loop is needed for IIS6 and others which have a handshake size limitations
@@ -7158,7 +7157,6 @@ get_server_certificate() {
                     tmpfile_handle ${FUNCNAME[0]}.txt
                     return 7  # this is ugly, I know
                else
-                    grep -a 'TLS server extension' $TMPFILE >>$TEMPDIR/tlsext.txt
                     GOST_STATUS_PROBLEM=true
                fi
           fi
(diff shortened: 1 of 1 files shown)


drwetter:
> However, I don't understand the reason that the extra call was added.

I am pretty sure there was a reason but I can't recall exactly. Looking at it today it seems that the loop through the protocols seems to be the reason. Maybe if it stopped at a certain $proto it got only the protocols supported (``break`` is a thing I am also questioning).

Your and my section seems still like two parts and I would rather see it fit better.  Is there any reason you call ``extract_new_tls_extensions()`` with  ``$TMPFILE`` and not make use of the extensions already collected in ``$TEMPDIR/tlsext.txt``?

Then the order of calling the functions is something which can be done better. ``get_server_certificate`` gets via openssl the extensions, ``extract_new_tls_extensions()`` is being called which does its almost separate task and we have ``determine_tls_extensions()`` which contains ```extract_new_tls_extensions()`` too.

I would vote for ONE function which gets all extensions if needed for ``heartbleed()`` + ``run_server_defaults()`` and those are kept in $TLS_EXTENSIONS.

drwetter:
Regarding extension another open point is that an array ``("A" "B")`` would fit better as now the output format with quotes is determined by the functions which determines the extensions.

That's minor though and probably quick to do as probably only "()" would need to be added.

dcooper16:
> I am pretty sure there was a reason but I can't recall exactly. Looking at it today it seems that the loop through the protocols seems to be the reason. Maybe if it stopped at a certain $proto it got only the extensions supported by using this protocol (`break` is a thing I am also questioning).
>
> Your and my section seems still like two parts and I would rather see it fit better. Is there any reason you call `extract_new_tls_extensions()` with `$TMPFILE` and not make use of the extensions already collected in `$TEMPDIR/tlsext.txt`?

Using `$TMPFILE` instead of `$TEMPDIR/tlsext.txt` was a mistake on my part. However, `$TEMPDIR/tlsext.txt` is only needed if `get_server_certificate()` has more than one call to `$OPENSSL s_client` from which extensions need to be extracted. If `get_server_certificate()` makes at most one successful call to `$OPENSSL s_client` (as it does in this PR), then there's no reason to first create `$TEMPDIR/tlsext.txt` and then call `extract_new_tls_extensions()` rather than just calling 
`extract_new_tls_extensions()` with `$TMPFILE`.

> Then the order of calling the functions is something which can be done better. `get_server_certificate` gets via openssl the extensions, `extract_new_tls_extensions()` is being called which does its almost separate task and we have `determine_tls_extensions()` which contains `extract_new_tls_extensions()` too.
> 
> I would vote for ONE function which gets all extensions if needed for `heartbleed()` + `run_server_defaults()` and those are kept in $TLS_EXTENSIONS.

I think the primary purpose of `get_server_certificate()` should be to get the server's certificate, but I don't like the idea of having just one function that  gets all extensions.

If this PR is accepted, then `get_server_certificate()` will make not make any extra calls for the purpose of finding extensions. So, the function's primary purpose will be finding certificates.

For `heartbleed()`, a call to `determine_tls_extensions()` will be sufficient to determine whether the server supports the heartbeat extension.

The problem with having `get_server_certificate()` not collect into `$TLS_EXTENSIONS` the extensions that it has (incidentally) found, it that it would mean that `determine_tls_extensions()` would have to do more work than it currently does.

At the moment, `determine_tls_extensions()` makes only one successful call to either `tls_sockets()` or `$OPENSSL s_client`. However, testssl.sh cannot determine whether a server supports both the NPN and ALPN extensions without making two separate calls. At the moment, `get_server_certificate()` checks for NPN (since checking for NPN doesn't create any extra work for `get_server_certificate()`) and `determine_tls_extensions()` checks for ALPN. If `get_server_certificate()` didn't call  `extract_new_tls_extensions()`, then `determine_tls_extensions()` would have to connect with the server an extra time in order to test for the NPN extension.

In addition, `determine_tls_extensions()` would have to check for the "status request" extension. This is something `get_server_certificate()`, since it needs to find out for each certificate the server has whether OCSP staping is supported. Again, if `get_server_certificate()` didn't call  `extract_new_tls_extensions()`, then `determine_tls_extensions()` would have to make extra connections to the server in order to find out whether the server supports "status request." This would be even worse that NPN, since:

- The server might have more than one certificate and support OCSP stapling with some, but not all of the certificates.

- The server might have the `GOST_STATUS_PROBLEM`, and so including `-status` in the request may cause the connection to fail, and another call would be needed to test for the other extensions (e.g., ALPN).

In addition, the "EC point formats" extension is generally only returned if the certificate has an ECC public key. So, an extra connection might be needed to check for this extension, even though it would have already been found by `get_server_certificate()` if the server supported it.

So, I think `get_server_certificate()` should make no additional connections to the server just to check for extensions. However, whatever extensions are returned by the server when looking for certificates should be added to `$TLS_EXTENSIONS` so that `determine_tls_extensions()` doesn't have to redo work that `get_server_certificate()` has already done.

dcooper16:
> Regarding extension another open point is that an array ("A" "B") would fit better as now the output format with quotes is determined by the functions which determines the extensions.
> 
> That's minor though and probably quick to do as probably only "()" would need to be added.

It seems that using an array would make checking whether `$TLS_EXTENSIONS` includes an extension more complicated:
```
if [[ $line != "" ]] && [[ ! "$TLS_EXTENSIONS" =~ "$line" ]]; then
if [[ ! "$TLS_EXTENSIONS" =~ encrypt-then-mac ]]; then
if ! grep -q heartbeat <<< "$TLS_EXTENSIONS"; then
```

If the goal is to separate the collection of the extensions from the printing, perhaps (along the lines of what you suggested), `$TLS_EXTENSIONS` could just be a string of extension numbers, and the `run_server_defaults()` could use that string along with a lookup table based on http://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml to print the supported extensions.

drwetter:
Thanks David. For now we should go as you suggested and _atm_ there's no need to worry about the redesign how we get extensions.

Issue #653, opened 2017-03-02
--append: no footer / two footers


drwetter:
Two problems here:

1) ``./testssl.sh -S --json-pretty --file testfile`` / ``./testssl.sh -S --json --file testfile`` have no footers and thus produce incorrect JSON. (``-S`` just as an example).
``--append`` seems to be the culprit: ``testssl.sh -S --json-pretty --warnings=batch -q --append dev.testssl.sh`` is being called. W/o ``--append`` ist works fine.

2) if a ``--json-file`` is being specified (e.g. ``./testssl.sh -S --json-pretty --jsonfile=aa.json  --file testfile``), which effectively calls ``./testssl.sh -S --json-pretty --jsonfile=aa  --warnings=batch -q --append dev.testssl.sh`` the file ``aa.json`` has two JSON footers

AlGreed:
> 2) if a --json-file is being specified (e.g. ./testssl.sh -S --json-pretty --jsonfile=aa.json --file testfile), which effectively calls ./testssl.sh -S --json-pretty --jsonfile=aa --warnings=batch -q --append dev.testssl.sh the file aa.json has two JSON footers

Why would we use --json-pretty  and --jsonfile=aa.json together, if we have --jsonfile-pretty for this case? If we use them together, the same context will be saved twice, but in different formats, in one file.

dcooper16:
He really meant using `--jsonfile-pretty` along with `--file`. This is based on a comment that I sent to Dirk:

> When I use both the "--json" option and the "--file \<testfile\>" option, the JSON files that are created for each line in "testfile" are missing footers. When I use the "--jsonfile \<jsonfile\>" option with the "--file \<testfile\>" option, the resulting JSON file has two footers. The same happens with "--json-pretty" and "--jsonfile-pretty".
> 
> Another problem I discovered is that if I use the "--jsonfile \<jsonfile\>" option with the "--file \<testfile\>" option, and "jsonfile" already exists, then testssl.sh will append the JSON output to "jsonfile" rather than overwriting (and it won't add a JSON header), even though I didn't include the "--append" option in the command line.

See also PR #652.

drwetter:
done for both branches

PR #683, opened 2017-03-27
Missing HTML banner

dcooper16:
With the commit made on March 26, "partly (1/2) fixing #653," an HTML banner isn't added to the HTML files anymore.  A banner should be added to the top of the HTML file if mass testing is being performed and a separate HTML file is being created for each test.

The `$APPEND` flag being `true` is an indicator that mass testing is being performed and that this is one of the individual tests being run. Given that `$APPEND` is `true`, `$HTMLHEADER` being `true` indicates that testssl.sh is creating the file name for the HTML output. So, it is when both flags are `true` that the HTML banner should be created.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1075,7 +1075,7 @@ html_header() {
 }
 
 html_banner() {
-     if ! "$APPEND" && "$HTMLHEADER"; then
+     if "$APPEND" && "$HTMLHEADER"; then
           html_out "## Scan started as: \"$PROG_NAME $CMDLINE\"\n"
           html_out "## at $HNAME:$OPENSSL_LOCATION\n"
           html_out "## version testssl: $VERSION ${GIT_REL_SHORT:-$CVS_REL_SHORT} from $REL_DATE\n"


drwetter:
> With the commit made on March 26, "partly (1/2) fixing #653," an HTML banner isn't added to the HTML files anymore

IIRC there wasn't a banner at all before in the scenarios I tested but it seems I tested something I can't remember (and reproduce).  

General Q: Why should a banner depend on file append?

My impression as I closed the laptop that night was rather that ``--append`` needs to be reconsidered. ``run_mass_testing()`` has for historical reasons I forgot ``cmdline="$0 $global_cmdline --warnings=batch -q --append $cmdline"``. If there's a single file that makes sense but only then, right?

drwetter:
Shouldn't there be always an HTML banner? If so I would just recommend removing ``$APPEND`` completely, That's at least how I redid the JSON part, see ``fileout_banner()``

drwetter:
For the time being your PR indeed fixes it. With the release of this version that should be changed, see #684 .

dcooper16:
> Shouldn't there be always an HTML banner? If so I would just recommend removing $APPEND completely, That's at least how I redid the JSON part, see fileout_banner()

Actually, the JSON file doesn't always have a banner. The main body of testssl.sh is as follows:
```
if $do_mass_testing; then
     prepare_logging
     run_mass_testing
     exit $?
fi

html_banner
fileout_banner
```

So, in the case of mass testing, `fileout_banner()` is only called by the "children." If `--jsonfile-pretty` is used instead of `json-pretty`, then `$JSONHEADER` will be set to `false` in the "children," so none of them will call `fileout_pretty_json_banner()` (which they shouldn't). So, in the case of an invocation like
```
bash testssl.sh --jsonfile-pretty out.json --file serverstotest.txt
```
no banner will be included in out.json.

PR #685, opened 2017-03-27
Populate OPENSSL_LOCATION in find_openssl_binary

dcooper16:
`$OPENSSL_LOCATION` is used in `fileout_pretty_json_banner()`, `html_banner()`, `mybanner()`, and `prepare_logging()`, but the value of `$OPENSSL_LOCATION` is populated in `mybanner()`. This is usually okay, since `mybanner()` is always called before the other three functions are called. However, if `$QUIET` is `true`, then `mybanner()` returns immediately, without populating `$OPENSSL_LOCATION`, even though the value of `$OPENSSL_LOCATION` may be needed by one or more of the other functions.

This PR addresses this problem by populating `$OPENSSL_LOCATION` in `find_openssl_binary()` rather than `mybanner()`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -10544,6 +10544,7 @@ test_openssl_suffix() {
 find_openssl_binary() {
      local s_client_has=$TEMPDIR/s_client_has.txt
      local s_client_starttls_has=$TEMPDIR/s_client_starttls_has.txt
+     local openssl_location cwd=""
 
      # 0. check environment variable whether it's executable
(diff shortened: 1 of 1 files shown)


drwetter:
FYI: There was a problem with the number of ciphers displayed in the banner in early times under some circumstances. That's why a short search has been done in mybanner.

Let me check what the was...

drwetter:
Can't fond anymore what it was. But it looks to me that the order in (non-existing) main() is different.

Thanks!

PR #686, opened 2017-03-28
out_row_aligned_max_width()

dcooper16:
This PR improves `out_row_aligned_max_width()` in a few ways:

* It makes better use of bash's string manipulation capabilities in order to simplify the function.

* It improves the function's performance. One of the most costly parts of `out_row_aligned_max_width()` was the while loop to print each entry in the text. Since there is only one place in the code where the entries are not all printed the same way (the list of supported curves printed by `run_pfs()`), the PR changes `out_row_aligned_max_width()` to just return a plain text string, which the calling function prints in the appropriate way. For the curves printed by `run_pfs()`, a new function, `out_row_aligned_max_width_by_entry()` takes care of getting the output from `out_row_aligned_max_width()` and then printing each entry appropriately.

* The PR also introduces a trick so that when the TLS extensions are printed, the text for an extension won't get split across two rows. It does this by replacing the space charters within the text for an extension with "}", formatting the result with `out_row_aligned_max_width()`, and then converting the "}" back to space characters.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1222,61 +1222,71 @@ out_row_aligned() {
 }
 
 # prints text over multiple lines, trying to make no line longer than $max_width.
-# Each line is indented with $spaces and each word in $text is printed using $print_function.
+# Each line is indented with $spaces.
 out_row_aligned_max_width() {
      local text="$1"
      local spaces="$2"
      local -i max_width="$3"
-     local print_function="$4"
      local -i i len
      local cr=$'\n'
-     local line entry first=true last=false
+     local line first=true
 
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -1222,63 +1222,73 @@ out_row_aligned() {
...
+# Each line is indented with $spaces.
 out_row_aligned_max_width() {
      local text="$1"
      local spaces="$2"
      local -i max_width="$3"
-     local print_function="$4"
      local -i i len
      local cr=$'\n'
-     local line entry first=true last=false
+     local line first=true
```
drwetter: One minor request wrt to readability: For new code please use one line per variable type in the declaration. "line" has type string, "first" is boolean.

Issue #684, opened 2017-03-27
Make use of append option only when append to a file


drwetter:
Historically ``--append`` was existing IIRC for safety reasons, so that for mass testing the out is being sent to ONE file it is **not** overwritten by subsequent writes for checks against other hosts.

In that sense the ``--append`` option should be used -- i.e. only when one output file is used ``--append`` should be used.

drwetter:
``--append`` is currently being used _also_ for inheriting the signal to the child "we do mass testing".

dcooper16:
How would you feel about adding a command line flag that is only intended to be used to signal to testssl.sh that it is a "child" in a mass testing? A flag that would be recognized by the code, but that wouldn't be advertised in the `--help` output. Perhaps `--mass-testing-child`. Then, functions such as `json_header()`, `csv_header()`, `html_header()`, and `html_banner()` wouldn't have to use the `$APPEND` flag as a way of trying to determine if mass testing is being performed.

At the moment `html_header()` does the following (and `json_header()` and `csv_header()` do the same):
```
     # Don't create HTML headers and footers in the following scenarios:
     #  * HTML output is not being created.
     #  * mass testing is being performed and each test will have its own HTML file.
     #  * this is an individual test within a mass test and all HTML output is being placed in a single file.
     if ! "$do_html" || \
        ( "$do_mass_testing" && ( [[ -z "$HTMLFILE" ]] || [[ -d "$HTMLFILE" ]] ) ) || \
        ( "$APPEND" && [[ -n "$HTMLFILE" ]] && [[ ! -d "$HTMLFILE" ]] ); then
               HTMLHEADER=false
               return 0
     fi
```
If there were a specific flag indicating mass testing, then the third line of this test could use that flag rather than `$APPEND` to determine whether "this is an individual test within a mass test."

drwetter:
Am 27.03.2017 um 21:39 schrieb David Cooper:
> How would you feel about adding a command line flag that is only intended to be used to signal
> to testssl.sh that it is a "child" in a mass testing? A flag that would be recognized by the
> code, but that wouldn't be advertised in the |--help| output. Perhaps |--mass-testing-child|.

I vote in run_mass_testing() for CHILD=true ./testssl.sh ... or maybe CHILD_MASS_TESTING as
opposed
to the parent $do_mass_testing

> Then, functions such as |json_header()|, |csv_header()|, |html_header()|, and |html_banner()|
[...]

PR #687, opened 2017-03-29
Child mass testing env variable

dcooper16:
This PR introduces the environment variable `CHILD_MASS_TESTING` (see #684), and uses it as an indicator that testssl.sh is running as a child within mass testing rather than using the `$APPEND` flag. It also makes a number of other changes to make the handling, of HTML, CSV, JSON, and log files consistent, and it fixes a number of bugs related to the generation of these files when mass testing is being performed.

Please let me know if you disagree with any of the changes in this PR, or if you would prefer that it be broken up into multiple smaller PRs.

Some of the changes are as follows:
- When the `$APPEND` flag is true, all of these files are appended to and headers and footers are omitted. (Perhaps this should be changed. Appending to a log file isn't an issue, but appending to a JSON or HTML file without including headers or footers seems to just create an improperly formatted file).
- Following the code in `prepare_logging()`, an error is printed and the program stops if the `$APPEND` flag is false and one of the files to be written to already exists.

Some of the bugs fixed:

Creating log files did not work with mass testing:
- If `--logfile <logfile>` is used, then the parent and each child try to write to "logfile".
- If `--logging` is used, then a log file is created for each child, but an oddly-named log file is also created for the parent. The one created by the parent contains the entire output.

Plain JSON files:
- When `--jsonfile <jsonfile>` is run, there is no comma separating the final finding for one child and the first finding for the next child.

Pretty JSON files:
- When `--jsonfile-pretty <jsonfile>` is called without mass testing, the "target host" line is empty, since `$NODE` has not yet been set.
- When `--jsonfile <jsonfile>` is run with mass testing, there is no comma separating the final finding for one child and the first finding for the next child. In addition, `fileout_pretty_json_banner()` is never called, and the entries for individual tests have insufficient information to determine what is being tested (it lists "service" and "ip", but not port number).

For the final issue, when mass testing is being performed and all output is being placed in a single file, I have the parent call `fileout_pretty_json_banner()`, but tell `fileout_pretty_json_banner()` to not include a "target host" or "port", but then have each child include a "target host" or "port" (when the "service" and "ip" are being printed).

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -186,6 +186,7 @@ GIVE_HINTS=false                        # give an addtional info to findings
 HAS_IPv6=${HAS_IPv6:-false}             # if you have OpenSSL with IPv6 support AND IPv6 networking set it to yes
 UNBRACKTD_IPV6=${UNBRACKTD_IPV6:-false} # some versions of OpenSSL (like Gentoo) don't support [bracketed] IPv6 addresses
 SERVER_SIZE_LIMIT_BUG=false             # Some servers have either a ClientHello total size limit or a 128 cipher limit (e.g. old ASAs)
+CHILD_MASS_TESTING=${CHILD_MASS_TESTING:-false}
 
 # tuning vars, can not be set by a cmd line switch
 EXPERIMENTAL=${EXPERIMENTAL:-false}
@@ -866,6 +867,8 @@ fileout_json_print_parameter() {
 }
 
 fileout_json_finding() {
+     local target
+
      if "$do_json"; then
           "$FIRST_FINDING" || echo -n "," >> "$JSONFILE"
           echo -e "         {"  >> "$JSONFILE"
@@ -884,9 +887,19 @@ fileout_json_finding() {
             if [[ $SERVER_COUNTER -gt 1 ]]; then
                 echo "          ," >> "$JSONFILE"
             fi
-            echo -e "          {
+            if "$CHILD_MASS_TESTING" && ! "$JSONHEADER"; then
+                 target="$NODE"
+                 $do_mx_all_ips && target="$URI"
+                 echo -e "          {
+                    \"target host\"     : \"$target\",
+                    \"port\"            : \"$PORT\"
                     \"service\"         : \"$finding\",
                     \"ip\"              : \"$NODEIP\","  >> "$JSONFILE"
+            else
+                 echo -e "          {
+                    \"service\"         : \"$finding\",
+                    \"ip\"              : \"$NODEIP\","  >> "$JSONFILE"
+            fi
             $do_mx_all_ips && echo -e "                    \"hostname\"        : \"$NODE\","  >> "$JSONFILE"
          else
              ("$FIRST_FINDING" && echo -n "                            {" >> "$JSONFILE") || echo -n ",{" >> "$JSONFILE"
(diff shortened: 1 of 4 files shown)


testssl.sh
```diff
@@ -11790,22 +11832,27 @@ run_mass_testing_parallel() {
...
      while read cmdline; do
           cmdline=$(filter_input "$cmdline")
           [[ -z "$cmdline" ]] && continue
           [[ "$cmdline" == "EOF" ]] && break
-          cmdline="$0 $global_cmdline --warnings=batch -q --append $cmdline"
+          cmdline="$0 $global_cmdline --warnings=batch $cmdline"
           draw_line "=" $((TERM_WIDTH / 2)); outln;
           outln "$cmdline"
-          $cmdline
+          if ! "$first" && "$JSONHEADER"; then
```
drwetter: Thanks first!

I do not understand the sense of L11850-L11855, could you shed some light on this?

dcooper16:
Hi Dirk,

This is to deal with one of the issues I mentioned above:

- When `--jsonfile <jsonfile>` is run, there is no comma separating the final finding for one child and the first finding for the next child.

As an example if I run `bash testssl.sh --jsonfile out.json  --crime www.ietf.org` the resulting out.json is:
```
[
         {
              "id"           : "service",
              "ip"           : "www.ietf.org/104.20.0.85",
              "port"         : "443",
              "severity"     : "INFO",
              "finding"      : "Service detected: HTTP"
          }
,         {
              "id"           : "crime",
              "ip"           : "www.ietf.org/104.20.0.85",
              "port"         : "443",
              "severity"     : "OK",
              "cve"          : "CVE-2012-4929",
              "cwe"          : "CWE-310",
              "finding"      : "CRIME, TLS: Not vulnerable"
          }
,         {
              "id"           : "service",
              "ip"           : "www.ietf.org/104.20.1.85",
              "port"         : "443",
              "severity"     : "INFO",
              "finding"      : "Service detected: HTTP"
          }
,         {
              "id"           : "crime",
              "ip"           : "www.ietf.org/104.20.1.85",
              "port"         : "443",
              "severity"     : "OK",
              "cve"          : "CVE-2012-4929",
              "cwe"          : "CWE-310",
              "finding"      : "CRIME, TLS: Not vulnerable"
          }
]
```
If I create a file called ietf.txt that contains:
```
 --ip 104.20.0.85 www.ietf.org
 --ip 104.20.1.85 www.ietf.org
```
and then run `bash testssl.sh --jsonfile out.json --crime --file ietf.txt` I would expect the resulting out.json to be the same, but instead the output is:
```
[
         {
              "id"           : "service",
              "ip"           : "www.ietf.org/104.20.0.85",
              "port"         : "443",
              "severity"     : "INFO",
              "finding"      : "Service detected: HTTP"
          }
,         {
              "id"           : "crime",
              "ip"           : "www.ietf.org/104.20.0.85",
              "port"         : "443",
              "severity"     : "OK",
              "cve"          : "CVE-2012-4929",
              "cwe"          : "CWE-310",
              "finding"      : "CRIME, TLS: Not vulnerable"
          }
         {
              "id"           : "service",
              "ip"           : "www.ietf.org/104.20.1.85",
              "port"         : "443",
              "severity"     : "INFO",
              "finding"      : "Service detected: HTTP"
          }
,         {
              "id"           : "crime",
              "ip"           : "www.ietf.org/104.20.1.85",
              "port"         : "443",
              "severity"     : "OK",
              "cve"          : "CVE-2012-4929",
              "cwe"          : "CWE-310",
              "finding"      : "CRIME, TLS: Not vulnerable"
          }
]
```
It is nearly identical, except that there is no comma before the opening "{" for the second "service" finding.

I just ran the second out.json through http://jsonlint.com/, and it complains about the missing comma, but says that the first out.json (with the comma) is valid.

The same missing comma issue appears in the json-pretty output.

drwetter:
Thx for the explanation (that would be helpful as a comment in the code). Of course that's no valid json. 

It was kind of what I was expecting, except the fact that this is a host with two IP addresses. I thought that was only for a command line which rather should make use of ``--append``, read: continuously writing into one file.

What bugs me a bit is that the lowlevel stuff -- adding a comma -- looks a little bit oddly placed here. Let me sleep over it whether I come up with a better idea for this.

drwetter:
> Please let me know if you disagree with any of the changes in this PR, or if you would prefer that it be broken up into multiple smaller PRs.

No worries, that's fine. The context is the same here.

>    When the $APPEND flag is true, all of these files are appended to and headers and footers are omitted. (Perhaps this should be changed. Appending to a log file isn't an issue, but appending to a JSON or HTML file without including headers or footers seems to just create an improperly formatted file).

That is a matter which can be tackled later. Step by step approach is easier here

dcooper16:
Hi Dirk,

I made a slight change to `run_mass_testing()` by moving the lines that add a comma separator between the findings for each test into a separate function, `fileout_separator()`.

I also discovered, by using http://jsonlint.com/, that I was missing a comma in the JSON output in the "port" line in the case of mass testing in which all output goes to a single file.

drwetter:
sounds good, David ! Take care of that by tomorrow, cheers, Dirk

drwetter:
Excellent! Thanks

PR #688, opened 2017-03-30
Dockerfile for testssl.sh

dr4y:
Added a Dockerfile for using testssl.sh with docker.

Usage:
docker run -t testssl.sh example.com

Ability to pull image from docker hub:
docker run -t dr4y/testssl.sh example.com

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,6 @@ FROM debian:latest
 
 RUN apt-get update && apt-get install -y git bsdmainutils dnsutils ldnsutils host
 
-#ADD ./ /testssl/
-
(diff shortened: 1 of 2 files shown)


drwetter:
@dr4y : have you looked into what @jumanjiman did?

Another thing to mention is that Debian is used and not Alpine Linux which is way smaller.

jumanjiman:
@dr4y If you maintain your docker image as a separate repo similar to [1], it would be appropriate to suggest your repo within the readme [here](https://github.com/drwetter/testssl.sh#externalrelated-projects).

[1] https://github.com/jumanjihouse/docker-testssl

drwetter:
As discussed in person Alpine would be great

drwetter:
temporarily merge to get thing going

PR #689, opened 2017-03-30
Cleanup variable definitions in run_server_defaults()

dcooper16:
In `run_server_defaults()` the variable `success` is defined twice, once an an ordinary variable and once as an array. The PR removes the incorrect definition. It also removes the definitions of some variables that are no longer used and reorganizes the definitions so that each line has only one variable type.

I also noticed a typo later in `run_server_defaults()` and corrected it.

I also corrected the indentation in `run_server_defaults()`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6080,18 +6080,16 @@ certificate_info() {
 
 
 run_server_defaults() {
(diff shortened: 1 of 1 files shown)


drwetter:
Thx

PR #698, opened 2017-04-05
Mass testing in parallel

dcooper16:
This PR modifies `run_mass_testing_parallel()` so that it may be used in place of  `run_mass_testing()`. This PR does not change the behavior of testssl.sh. In order to actually test it out, line 12682 needs to be changed to call `run_mass_testing_parallel()` instead of `run_mass_testing()`.

In order for the parallel testing to work correctly, some global variables needed to be created. The reason for this is that if testing is interrupted, and `cleanup ()` is called in the parent, then the parent needs to kill all of the child processes before the parent deletes its `$TEMPDIR`. Otherwise, error messages will be created as the child processes try to write to files in the parent's `$TEMPDIR` after that directory has been deleted.

One issue that is unresolved. At the moment, if testing is interrupted, then the output of any child processes that haven't yet completed is just thrown away rather than adding the partially completed results. Similarly, if a child process is killed because it doesn't finish within a reasonable time (if parent has been waiting `$MAX_WAIT_TEST` seconds for it to complete), then any partial results from that child process are thrown away.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -296,6 +296,15 @@ HEX_CIPHER=""
 
 SERVER_COUNTER=0                        # Counter for multiple servers
 
+########### Global variables for parallel mass testing
+readonly PARALLEL_SLEEP=1                 # Time to sleep after starting each test
+readonly MAX_WAIT_TEST=600                # Maximum time to wait for a test to complete
+readonly MAX_PARALLEL=20                  # Maximum number of tests to run in parallel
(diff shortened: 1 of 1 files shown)


drwetter:
nice shot. 👍 

yes, that needs to mature. For serial testing I didn't care but for parallel tests I think there should be also a confirmation like "you are sure you want to stop this - children aren't finished yet?"

For the filenames you put more thoughts into it than I would have done. :-) Allow directories only for {csv,json*,html}file options I guess would have been also ok with the majority of users.

PR #699, opened 2017-04-06
Fix #695

### LINKED ISSUE #695: formatting long CSP lines
In seldom cases for hosts which have e.g.  long CSP headers the line wrapping by  ``out_row_aligned_max_width()`` is not yet optimal, see 
``testssl.sh -H securityheaders.io`` or ``testssl.sh -H github.com``.

dcooper16:
This PR fixes issue #695 by changing the call to `out_row_aligned_max_width()` so that the length of the flag is considered in determining the length of the first line of the output.


drwetter:
Thank you

PR #701, opened 2017-04-06
Handle word splitting of log file names.

dcooper16:
This PR fixes a problem that occurs if log file names include spaces.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -11025,13 +11025,13 @@ prepare_logging() {
      fi
 
      if ! "$APPEND"; then
-          [[ -e $LOGFILE ]] && outln && fatal "\"$LOGFILE\" exists. Either use \"--append\" or (re)move it" 1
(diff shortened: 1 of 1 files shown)


drwetter:
Those missing quotes also have the impact that it allows potentially command injection, to give you a picture e.g. ``./testssl.sh --XXXfile="foo ; cat /etc/passwd" host2test`` (non-working example) . That is not a problem for a local user as they have command line access anyway but for folks e.g. supplying a web interface it is a problem IF THEY RELY ON perfect input validation of testssl.sh .

As the banner says ... USE IT AT YOUR OWN RISK! :-)

My commit c75a2cd8382edbc79b6f46c4ba3a8e1cd2d9262b  improves matters. But there's always still room for improvements...

drwetter:
One evil example:

ops

which doesn't work:

notworking

(presented at a Bsides conference a while back...)

Issue #618, opened 2017-02-06
Argument separation in CSV/flat JSON 


drwetter:
In CSV output generate by ``./testssl.sh -x CBC --csv HOSTNAME`` I got lines like 

``"cipher_x35","HOSTNAME/AA.BB.CC.DD","443","INFO","x35 AES256-SHA RSA AES 256 TLS_RSA_WITH_AES_256_CBC_SHA available","","",""``

rather than 

``"cipher_x35","HOSTNAME/AA.BB.CC.DD","443","INFO","x35","AES256-SHA","RSA","AES","256" "TLS_RSA_WITH_AES_256_CBC_SHA"``

Flat JSON is similar. 

Which is from the user perspectiveunnecessary difficult to further process -- at least as far as my expectation as a user would be concerned.

Atm probably there's no easy solution, I am open for suggestions though.

AlGreed:
For a normal mode (./testssl --csv hostname:port) it will not work because of different numbers of columns for different findings. 

For --json it can be done if we use an array for "finding":
```
[
        {
              "id"           : "service",
              "ip"           : "ccc-in2.wulf.dc.dca.wm.in/10.100.61.21",
              "port"         : "443",
              "severity"     : "INFO",
              "finding"      : ["Service detected: HTTP"]
         }
,        {
              "id"           : "cipher_xc028",
              "ip"           : "ccc-in2.wulf.dc.dca.wm.in/10.100.61.21",
              "port"         : "443",
              "severity"     : "INFO",
              "finding"      : ["xc028", "ECDHE-RSA-AES256-SHA384", "ECDH 256", "AES", "256",    "TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384", "available"]
         }
```


**So do we want to have it only for CBC mode??? If - Yes, then it can be done easily (almost) in the following way:**

**CSV:**
```
"id","fqdn/ip","port","severity","finding","cve","cwe","hint"
"finding" ->>>> "hexcode","csn_openssl","key_exch","encryption","bits","csn_rfc","availability"
```
**JSON:**
```
   {
              "id"           : "cipher_xc028",
              "ip"           : "ccc-in2.wulf.dc.dca.wm.in/10.100.61.21",
              "port"         : "443",
              "severity"     : "INFO",
[...]

PR #703, opened 2017-04-06
Argument separation in CSV/flat JSON

AlGreed:
As i mentioned in the thread #618 , in my opinion it would be enough to have for this case a csv table.  Additionally JSON format for the user perspective is not enough representable to get an overview of comparison of encryption ciphers.

This scenario works only for testing of ciphers.
For example:
./testssl.sh -x CBC --csv host:port
./testssl.sh -x SSL_CK_RC2_128_CBC_EXPORT40_WITH_MD5  --csv host:port

So: 
cbc cyphers

**About code:**
Some fields were like "ECDH 570", other like " RSA(512)"

If a cipher was not available, so key exchange was printed colorless and in round brackets 
like "RSA(512)".
If a cipher was available, so bits in key exchange was printed separate and colorful according to cipher strength like "ECDH 570"

For now  i made it more consistent and bits are always printed in round brackets, but still colorful only for available ciphers.

image


testssl.sh
```diff
@@ -760,8 +760,17 @@ fileout() {
      if ( "$do_pretty_json" && [[ "$1" == "service" ]] ) || show_finding "$severity"; then
          local finding=$(strip_lf "$(newline_to_spaces "$(strip_quote "$3")")")
          [[ -f "$JSONFILE" ]] && (fileout_json_finding "$1" "$severity" "$finding" "$cve" "$cwe" "$hint")
-         "$do_csv" && \
-              echo -e \""$1\"",\"$NODE/$NODEIP\",\"$PORT"\",\""$severity"\",\""$finding"\",\""$cve"\",\""$cwe"\",\""$hint"\"" >> "$CSVFILE"
+
+         if "$do_csv"; then
+            if "$do_test_just_one"; then
```
AlGreed: a boolean variable "do_test_just_one" will be set to true only in the case of -x:

```
-x|-x=*|--single[-_]cipher|--single[-_]cipher=*)
     do_test_just_one=true
     single_cipher=$(parse_opt_equal_sign "$1" "$2")
     [[ $? -eq 0 ]] && shift
     ;;
```
 Is it a right name for it? How i understand if we type "-x CBC" -> all ciphers with cipher block chaining will be analysed. Maybe could we rename it to something like "only_specific_ciphers"? Otherwise "if "$do_test_just_one"; then" is abit confusing in my opinion by reading of the code.

testssl.sh
```diff
@@ -841,7 +850,11 @@ csv_header() {
           CSVHEADER=false
      else
           [[ -e "$CSVFILE" ]] && fatal "\"$CSVFILE\" exists. Either use \"--append\" or (re)move it" 1
-          echo "\"id\",\"fqdn/ip\",\"port\",\"severity\",\"finding\",\"cve\",\"cwe\",\"hint\"" > "$CSVFILE"
+          if "$do_test_just_one"; then
+             echo "\"ID\",\"FQDN/IP\",\"Port\",\"Severity\",\"Hexcode\",\"OpenSSL CSN\",\"Key Exchange\",\"Encryption\",\"Bits\",\"RFC CSN\",\"Availability\",\"CVE\",\"CWE\",\"Hint\"" > "$CSVFILE"
```
AlGreed: Not sure that we need CWE, CVE and maybe Hint for this case? to remove from table?

testssl.sh
drwetter: yes, we could change this to ~ ``do_cipher_matched``. ``test_just_one()``. That anyway wasn't consistent and then would be needed to be renamed then to ``run_cipher_matched()``.

drwetter: But lets keep this for a separate PR pls

testssl.sh
drwetter: For this case not, at least not yet. We would have to look at single ciphers in order to tell whether it represents something good or bad -- which we don't do.

testssl.sh
```diff
@@ -2417,7 +2430,7 @@ neat_list(){
           fi
           return 0
      fi
-     read what_dh bits <<< "$kx"
+     read what_dh bits <<< "$(echo "$kx" | awk -F"(" '{a=$1" ("$2; print a}')"
```
drwetter: Not sure about the reason of the change here but if there's one, one minor thing: I would rather avoid a pipe like  ``read what_dh bits <<< "$(awk -F"(" '{a=$1" ("$2; print a}') <<< "$kx" "

testssl.sh
```diff
@@ -2624,7 +2637,7 @@ test_just_one(){
                     ciphers_found[i]=true
                     if [[ ${kx[i]} == "Kx=ECDH" ]] || [[ ${kx[i]} == "Kx=DH" ]] || [[ ${kx[i]} == "Kx=EDH" ]]; then
                          dhlen=$(read_dhbits_from_file "$TMPFILE" quiet)
-                         kx[i]="${kx[i]} $dhlen"
+                         kx[i]="${kx[i]}($dhlen)"
```
drwetter: Nah, that looks odd for everything else than CSV output. Also the round brackets are colorized.

drwetter:
@AlGreed: The part except in neat_list() is fine, thank you! Could you please rebase and then do a separate PR for this part? For CWE. CVE and hints just omit them

PR #706, opened 2017-04-11
Fix #702

### LINKED ISSUE #702: File names with space characters
In developing PR #698 I discovered a problem with filenames that include space characters.

When mass testing is not being performed, then I have only verified the problem in two places, with log file names and with the "PATH" for the OpenSSL binary. PR #701 addresses the problem with log file names. I believe the issue could be addressed for the OpenSSL binary by placing every use of `$OPENSSL` in quotes, but I've haven't tried it yet.

The biggest problem I've encountered is if mass testing is being performed and a filename with space characters needs to be included in the command line for the child processes. Using the command line:
```
bash testss.sh --htmlfile my\ results testssl.sh
```
or
```
bash testssl.sh --htmlfile "my results" testssl.sh
```
results in an HTML file being created in the "my results" directory.

However, the command line:
```
bash testssl.sh --htmlfile "my results" --file testlist.txt
```
results in child processes being called with command lines like:
```
testssl.sh --htmlfile my results  --warnings=batch testssl.sh
```
and each child printing "Fatal error: URI comes last". A malformed file "my" is created that contains, for each row in "testlist.txt", the fatal error message and an HTML footer.

I tried to find a way to correct the command lines for the child processes, but nothing I tried worked.

dcooper16:
This PR addresses issue #702. Rather than create the command line for each child process in `run_mass_testing()` as a string, it creates it as an array, with each argument being a separate element in the array. This was done based on http://mywiki.wooledge.org/BashFAQ/050.

The printing of each child's command line done based on http://stackoverflow.com/questions/10835933/preserve-quotes-in-bash-arguments.

The `$CMDLINE` string remains unchanged, even though it isn't entirely "correct," since http://jsonlint.com/ complains if the "Invocation:" string contains backslashes.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -11763,7 +11763,12 @@ run_mass_testing() {
           draw_line "=" $((TERM_WIDTH / 2)); outln;
           outln "$(create_cmd_line_string "$0" "${MASS_TESTING_CMDLINE[@]}")"
           "$first" || fileout_separator                              # this is needed for appended output, see #687
-          CHILD_MASS_TESTING=true "$RUN_DIR/$PROG_NAME" "${MASS_TESTING_CMDLINE[@]}"  # we call ourselves here. $do_mass_testing is the parent, $CHILD_MASS_TESTING... you figured
+          # we call ourselves here. $do_mass_testing is the parent, $CHILD_MASS_TESTING... you figured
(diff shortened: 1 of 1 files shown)


drwetter:
great 👍 Thanks!

PR #707, opened 2017-04-12
More fixes for #702

dcooper16:
This PR just addresses some places where quotes need to be used to avoid word splitting in case the referenced file, or path to the file, contains space characters.


drwetter:
Oh yes, Thank you David.

PR #708, opened 2017-04-13
Use get_cipher() helper function

dcooper16:
The new `get_cipher()` helper function was not being used in every place where it could be used.


drwetter:
Thx!

PR #709, opened 2017-04-13
Fix #616

### LINKED ISSUE #616: --show-each is default for -x <pattern>
e.g. while testing ``./testssl.sh -x CBC <target>``

dcooper16:
This PR addresses issue #616, changing `run_cipher_match()` so that only those ciphers that are available are shown, unless the `--show-each` flag has been provided.

It also fixes a problem where the signature algorithm isn't being shown, even if `$SHOW_SIGALGO` is true.


drwetter:
very good

PR #712, opened 2017-04-14
Add Travis test for HTML output

dcooper16:
I've never programmed in perl before, but this script seems to work. It includes two checks:

* I runs testssl.sh without the `--debug` flags and checks that the HTML file is the same as what is sent to the terminal.

* It runs testssl.sh with `--debug 4` and checks that the HTML file created is the same as the one created without the `--debug` flag.

This PR is based on the request in #692.

### DIFF
--- a/t/02_http.t
+++ b/t/02_http.t
@@ -1,0 +1,47 @@
+#!/usr/bin/env perl
+
+use strict;
+use Test::More;
+use Data::Dumper;
+
+my $tests = 0;
+
+pass("Running testssl.sh against badssl.com to create HTML and terminal outputs (may take 2~3 minutes)"); $tests++;
+# specify a TERM_WIDTH so that the two calls to testssl.sh don't create HTML files with different values of TERM_WIDTH
+my $okout = `TERM_WIDTH=80 ./testssl.sh --color 0 --quiet --append --htmlfile tmp.html badssl.com`;
+my $okhtml = `cat tmp.html`;
+# $modedhtml will contain the HTML with formatting information removed in order to compare against terminal output
+my $modedhtml = `cat tmp.html`;
+unlink 'tmp.html';
+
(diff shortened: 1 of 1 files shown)


drwetter:
thanks!

drwetter:
Travis complained , thus I unfortunately needed to reverted that..

```
ok

t/02_http.t .................. 

ok 1 - Running testssl.sh against badssl.com to create HTML and terminal outputs (may take 2~3 minutes)

ok 2 - Comparing HTML and terminal outputs

ok 3 - HTML file matches terminal output

ok 4 - Running testssl.sh against badssl.com with --debug 4 to create HTML output (may take 2~3 minutes)

ok 5 - Checking that using the --debug option doesn't affect the HTML file

not ok 6 - HTML file created with --debug 4 matches HTML file created without --debug

#   Failed test 'HTML file created with --debug 4 matches HTML file created without --debug'

#   at t/02_http.t line 52.
```

drwetter:
Hi David,

I thought a git-reset brings your PR back but it did not. If the travis error has been fixed would you mind to resubmit?

Cheers, Dirk

dcooper16:
Hi Dirk,

I looked at the travis error and found out why the comparison between the non-debug and debug versions of the HTML differed:
```
> diff expected.txt got.txt                                                                                                                                                                                          
229c229                                                                                                                                                                                                                          
< #  HTTP clock skew              +1 sec from localtime                                                                                                                                                                          
---                                                                                                                                                                                                                              
> #  HTTP clock skew              X sec from localtime
```

I guess in all the previous times that I had run the script I had never encountered a case in which one had a positive offset and the other had a different offset. The problem was that I did not escape the plus sign in my regular expression, and so it did not replace "+1" with "X". That should be fixed now, so hopefully it won't happen again.

PR #713, opened 2017-04-18
Use $TESTSSL_INSTALL_DIR instead of $RUN_DIR in find_openssl_binary()

gniltaws:
find_openssl_binary() should use $TESTSSL_INSTALL_DIR instead of $RUN_DIR once get_install_dir() has discovered it, especially since $TESTSSL_INSTALL_DIR is set to $RUN_DIR if it is correct.  

I have a symlink in /usr/local/bin pointing to testssl.sh so I can run it from anywhere.  Before making this change, it was looking for openssl in /usr/local/bin, failing, using the default system openssl, and complaining about GOST support.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -957,17 +957,17 @@ hex2ascii() {
 count_lines() {
      #echo "${$(wc -l <<< "$1")// /}"
      # ^^ bad substitution under bash, zsh ok. For some reason this does the trick:
-     echo $(wc -l <<< "$1")
+     wc -l <<< "$1"
 }
 
 count_words() {
      #echo "${$(wc -w <<< "$1")// /}"
      # ^^ bad substitution under bash, zsh ok. For some reason this does the trick:
-     echo $(wc -w <<< "$1")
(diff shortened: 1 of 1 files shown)


drwetter:
Hi Todd,

thanks for your efforts but please follow the rules: One PR per context.

* egrep eplacement I'll likely don't want now.
* shellcheck isn't always right. Randomly I am running shellcheck myself. Sometimes there's a reason why it is as it is, sometimes not. , E.g.``wc -l <<< "$1"`` doesn't work for BSD.
* trailing spaces: the lower part screws up the banner 
* find_openssl_binary(): need to look into to. Please file a separate PR

Cheers, Dirk

gniltaws:
Sorry, I only meant for the find_openssl_binary() part to be in the pull request!  I'm fairly new to github and I didn't realize my commits were being added automatically.

Thanks,
-Todd

drwetter:
No prob, Todd. Happened to me too a while back. Thx!

PR #715, opened 2017-04-19
Add Travis test for HTML output

dcooper16:
It seems that I needed to escape the plus sign in the check for the HTTP clock skew.


drwetter:
Thanks for the fix and the new PR!

PR #716, opened 2017-04-19
Use $TESTSSL_INSTALL_DIR instead of $RUN_DIR in find_openssl_binary() - Second Try

gniltaws:
find_openssl_binary() should use $TESTSSL_INSTALL_DIR instead of $RUN_DIR once get_install_dir() has discovered it, especially since $TESTSSL_INSTALL_DIR is set to $RUN_DIR if it is correct.

I have a symlink in /usr/local/bin pointing to testssl.sh so I can run it from anywhere. Before making this change, it was looking for openssl in /usr/local/bin, failing, using the default system openssl, and complaining about GOST support.


drwetter:
Thx, Todd

PR #721, opened 2017-04-21
Add wide option for client simulations

dcooper16:
There is a comment in the `run_client_simulation()` function that says "FIXME: printf formatting would look better, especially if we want a wide option here."

This PR is an attempt at addressing that FIXME and adding a wide option. The proposed wide option prints the same information as the non-wide option, just with the columns aligned. I didn't add any of the additional information that is displayed by other functions in wide mode, since I thought that made the output too wide.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3515,8 +3515,8 @@ run_client_simulation() {
      local minEcdsaBits=()
      local requiresSha2=()
      local i=0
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, David!

I think it's better now, but yeah, you're right. As I saw now your suggestion for the wide option there's not so much space left for the wide option used elsewhere in testssl.sh. 

Despite that it feesl a bit more consistent now though.

PR #732, opened 2017-04-30
Add support for single character reads as fallback for STARTTLS

### LINKED ISSUE #731: starttls_full_read: Support SMTP tarpits
Checking tarpits, especially SMTP tarpits, is cumbersome. They are so slow that they force the tester to set STARTTLS_SLEEP to a very high value – long enough for fetching a whole line. Considering a server greeting with 80 characters, with characters sent one per second, the timeout has to be in the range of minutes instead of seconds.

Such a long timeout has two major disadvantages. First, if a connection really hangs, the whole starttls_full_read process will have a very slow death. Second, choosing the right timeout is hard as it depends on the per-character delay, times the maximum line length, which is variable.

A new implementation for reading will first try to read a full line with a timeout of STARTTLS_SLEEP. The timeout can be chosen in a range of single-digit seconds, e.g. `3` (seconds). If a full line could not be read, single-character reads with a different timeout, e.g. `1` (second) as STARTTLS_SLEEP_TARPIT are done, until either a full line was read, or a timeout was reached which returned not a single character.

Conventional servers will respond within the STARTTLS_SLEEP time, so there is no slowdown for them. Servers with a real problem will see STARTTLS_SLEEP+STARTTLS_SLEEP_TARPIT before the routine fails. Tarpits are correctly scanned if they supply a character at least every STARTTLS_SLEEP_TARPIT seconds.

typingArtist:
Reading text for STARTTLS has moved into a separate function. The function handles switching between the modes (line-based vs. character-based reading) autonomously.

Fixes #731 

This PR is based on #730.


drwetter:
Thx @typingArtist .

You have a test host for me (tarpit) (dm/pm/whatsover)?

drwetter:
0) It's a cool idea!

1) doesn't fly yet, seems to hang at SSLv3 'reading server hello... \nTLS message fragments:'
2) debugging info seems to be different if it is a tarpit host, i.e. if first line read in ``starttls_read_one_line()`` fails
3) testssl.sh uses double square brackets as single ones are an external command (try ``which ]`` in bash)

Cheers, Dirk

drwetter:
Hi @typingArtist ,

I would like to close this for housekeeping reasons: As said it's a good idea but would need a bit of work. last but not least the branch you tried to patch against (2.9dev), is dead.

typingArtist:
Ah, I have completely forgotten about that. Would you accept fixes rebased on current dev branch? 

Also, what shall we do about testing? I see two options here:

1. Create a permanent public server for the sole purpose of tarpit testing
2. Create a temporary server-like local thread that simulates a tarpit STARTTLS server

I’d like to go with option 2 but I first have to learn about Perl testing anyway.

drwetter:
> Ah, I have completely forgotten about that. Would you accept fixes rebased on current dev branch?

No prob. There's no current dev branch :-) After release (or shortly before) of 3.0 there will be a 3.1dev branch

>     Create a temporary server-like local thread that simulates a tarpit STARTTLS server
>
> I’d like to go with option 2 but I first have to learn about Perl testing anyway.

2) sounds fine to me. But I don't expect to write a unit test for that rare species. It's ok if we have something to test against. 

OT: In general (@ all) unit tests are documented [here](https://perldoc.perl.org/Test/More.html) and if one looks a existing scripts that should suffice.

Ok to close the PR and you resubmit when 3.1dev is there?

typingArtist:
Sure. I do.

PR #740, opened 2017-05-10
Improvements to mass testing in parallel

dcooper16:
This PR provides improvements to `run_mass_testing_parallel()`. Currently, `run_mass_testing_parallel()` treats `$MAX_PARALLEL` as the maximum difference between the number of the test whose results were last processed and the number of the most recently started test. This means that test 40 will not be started until the results of test 20 have been processed. I've encountered situations in which tests 21 though 39 have completed, but test 20 is still running, and so no new tests are started.

This PR fixes the problem by checking the status of all running child tests to see if any are complete, rather than just looking at `$NEXT_PARALLEL_TEST_TO_FINISH`. This prevents one slow child test (or a few slow child tests) from slowing up the entire mass testing process.

This PR also changes the basis for determining whether a slow child process should be killed. Rather than waiting `$MAX_WAIT_TEST` seconds from the time that the parent started waiting (which is rather arbitrary), it kills the process if `$MAX_WAIT_TEST` seconds have passed since the child test was started. Given this, and that the above change makes it less likely that a slow child test will slow up the overall testing, I increased `$MAX_WAIT_TEST` from 600 seconds to 1200 seconds.

I added some `debugme` statements that provide feedback on the status of testing, but in non-debug mode there may be a perception issue. If one test (e.g., test 20) is very slow, testssl.sh will not display any results from later tests until the slow test finishes, even though testssl.sh will continue running new tests in the background. The user, seeing no output from testssl.sh for an extended period of time, may think that testssl.sh has frozen, even though it is really just holding back on displaying the later results so that the results will be displayed in the order in which the tests were started.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -305,9 +305,9 @@ SERVER_COUNTER=0                        # Counter for multiple servers
 
 ########### Global variables for parallel mass testing
 readonly PARALLEL_SLEEP=1                 # Time to sleep after starting each test
-readonly MAX_WAIT_TEST=600                # Maximum time to wait for a test to complete
+readonly MAX_WAIT_TEST=1200               # Maximum time (in seconds) to wait for a test to complete
 readonly MAX_PARALLEL=20                  # Maximum number of tests to run in parallel
-declare -a -i PARALLEL_TESTING_PID=()     # process id for each child test
+declare -a -i PARALLEL_TESTING_PID=()     # process id for each child test (or 0 to indicate test has already completed)
 declare -a PARALLEL_TESTING_CMDLINE=()    # command line for each child test
 declare -i NR_PARALLEL_TESTS=0            # number of parallel tests run
 declare -i NEXT_PARALLEL_TEST_TO_FINISH=0 # number of parallel tests that have completed and have been processed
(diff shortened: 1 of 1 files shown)


drwetter:
Thx, David.

Currently in this branch ``run_mass_testing_parallel`` isn't called yet -- so I take this as a preparation for PRs to come :-)

What is the sense of ``curr_time=$(date +%s)`` and later ``curr_time+=1`` ?

Cheers, Dirk

dcooper16:
> Currently in this branch `run_mass_testing_parallel` isn't called yet -- so I take this as a preparation for PRs to come :-)

I hope that `run_mass_testing_parallel` can replace `run_mass_testing` at some point. I've been using it locally, and it really helps when I want to test something against a large number of sites (I also tend to increase `$MAX_PARALLEL` to 50). However, I understand that its too early to make `run_mass_testing_parallel` the default. In #698 you said:

> yes, that needs to mature. For serial testing I didn't care but for parallel tests I think there should be also a confirmation like "you are sure you want to stop this - children aren't finished yet?"

I'm going to try to work on developing something that will provide feedback as tests are being run in the background (so that testssl.sh doesn't appear to be frozen when it isn't), but I wasn't planning to develop a "are you sure you want to stop this?" confirmation dialog. I was hoping someone else would do that.

> What is the sense of `curr_time=$(date +%s)` and later `curr_time+=1` ?

I guess `curr_time+=1` could be replaced with `curr_time=$(date +%s)`. I just figured that `curr_time+=1` was more efficient than calling `date +%s`, and it was sufficiently accurate since I knew that slightly over 1 second had passed since the previous time that `curr_time` had been updated. But, I suppose the cost of calling `date +%s` is so small it doesn't make a difference.

PR #745, opened 2017-05-16
More improvements to mass testing in parallel

dcooper16:
This PR introduces the following changes/improvements to running mass testing in parallel:

* Continuous feedback is provided on the progress of testing, so that testssl.sh doesn't appeared to be frozen, even if it is waiting a long time for a child test to complete. [The feedback text is sent to `stderr` so that it doesn't appear in the log file if the `--logging` option is used.]

* Text sent to `stderr` by a child test is captured and then displayed in context with the appropriate test rather than being displayed at the time the error occurs. [Note that this means that if the `--logfile <logfile>` option is used (with `logfile` being the name of a file), then `logfile` will include both `stdout` and `stderr`.]

* If a test is started but it does not complete (because testssl.sh was stopped or because the test timed out), then a message is displayed indicating that the test didn't finish.

I believe that `run_mass_testing_parallel()` is either ready for use or nearly ready for use. So, I changed the program to use `run_mass_testing_parallel()` rather than `run_mass_testing()` if the `$EXPERIMENTAL` flag is set.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -307,6 +307,7 @@ SERVER_COUNTER=0                        # Counter for multiple servers
 readonly PARALLEL_SLEEP=1                 # Time to sleep after starting each test
 readonly MAX_WAIT_TEST=1200               # Maximum time (in seconds) to wait for a test to complete
 readonly MAX_PARALLEL=20                  # Maximum number of tests to run in parallel
+                                          # This value may be made larger on systems with faster processors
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, David. Much appreciated! 👍 

@ everybody listening: Please run ``EXPERIMENTAL=true testssl.sh --file <yourhostsortestshere>`` and let @dcooper16 know. I did some saber rattling on twitter, hoping for feedback.

In the end I probably would like to keep the serial function as a fallback .

Cheers, Dirk

PR #753, opened 2017-05-22
Massing testing with command line error

dcooper16:
There is a bug in testssl.sh that occurs if mass testing is being performed, there is an error in the command line for one of the child tests, and either a single HTML file or a single JSON file is being created.

If mass testing is being performed and `parse_cmd_line()` detects an error in the command line for one of the child tests, then it will call `help()`, which will exit the program, resulting in `cleanup ()` being called. `cleanup ()` will call `html_footer()` and `fileout_footer()`. Since `html_header()` and `json_header()` have not yet been called, `$HTMLHEADER` and `$JSONHEADER` will both be `true, and so `html_footer()` and `fileout_footer()` will output HTML and JSON footers, even though no headers have been output.

This PR fixes the problem by having `help()` set `$HTMLHEADER` and `$JSONHEADER` to `false` so that no HTML or JSON footers are created.

A related problem is that if a single JSON file is being created, the parent process will insert a separator (a comma) into the JSON file between the outputs of each child process. However, if there is an error in one of the child process's command lines, then this child process will not produce any JSON output and so the JSON file will have two consecutive separators (commas), which is invalid according to http://jsonlint.com.

This PR provides a partial fix for the problem for parallel mass testing by checking whether a child process has created a non-empty JSON output before adding a separator to the JSON file. It leaves two unresolved problems:

* It does not fix the problem at all for `run_mass_testing()`, where the separator is added before the test with the command line error is run.

* It does not fix the problem for parallel mass testing for the case in which the first child test has a command line error.


drwetter:
Thanks, @dcooper16 !

As a human I would rather expect testssl.sh to exit immediately if there's a problem parsing the file. Then the user is supposed to fix this. 

Admittedly this seems not really easy to implement: The master has currently no clue about the parsing status of the lines in the file done by the children and we don't want the master parsing the file upfront (probably, maybe to be considered later).

This would leave then e.g. signal(l?)ing or other intercommunication mechanisms from child to parent. Bash has ``$PPID`` as a variable. Wouldn't it be easier sending e.g. SIGUSR1 to PPID and write another trap handler to deal with this signal, e.g. cleanup everything and call ``fatal()``? This approach would be probably less complex and should address the unresolved problems you mentioned too.

dcooper16:
It seems that the problem of a child process not creating any JSON output is bigger than I initially thought. It can happen if there is a syntax error in the command line (of the type that would be caught by `parse_cmd_line()`), but there are other cases, such as if testssl.sh cannot connect to the target server.

So, I modified this PR so that, whether serial or parallel mass testing is performed, the JSON output of the child is first placed in a temporary file. Then, a separator is only added to the parent JSON file if the child produced a non-empty JSON output and it is not the first child to have produced a non-empty JSON output.

This doesn't address the issue of exiting testssl.sh if there is a problem parsing the mass testing file, but it seems that that should be a separate PR.

drwetter:
Oh, got you now, thx!

PR #754, opened 2017-05-23
Check mass testing file before testing

dcooper16:
In response to #753, this PR checks the mass testing file for syntax errors before starting any testing.

It seems that if all testing is to be stopped upon encountering a syntax error in the testing file, it would be best to find the error up front rather than potentially run a large number of tests and then stop testing in the middle.

Note that this PR declares that certain options are not valid for a child test:
 * --help
* -b, --banner, -v, --version,
* -V, -V=\*, --local, --local=\*
* --file|--file=\*

I included the `--file` option since I believe that testssl.sh will not work correctly if a single CSV, JSON, or HTML file is to be created and of of the child tests is itself a mass test.

The suggestion of having the child process create a signal may still be necessary, if there are other errors that would not be caught by this code that should cause testing to stop. For example, if the testing file includes `--openssl=..." and no OpenSSL binary can be found at the specified location. It is just a question of what errors in a child test should cause all testing to stop rather than allowing the other child tests to continue running.


dcooper16:
I decided to add options to the list of options that are invalid for a child test:
* --openssl=\*, --openssl
 * --jsonfile=\*, --jsonfile, --jsonfile-pretty=\*, --jsonfile-pretty
* --logfile=\*, --logfile
* --csvfile=\*, --csvfile
* --htmlfile=*, --htmlfile

I added these for two reasons. First, testssl.sh will get confused a create incorrect results if a CSV, HTML, JSON, or log file name is provided for a child test and the file is only supposed to hold the results of that one test (testssl.sh currently assumes that provided file will hold the combined results of all of the child tests).

The second reason is that testssl.sh's ability to handle space characters in file names that are specified in the mass testing file. For example, it cannot a line in the mass testing file being:
```
--htmlfile "html output.html" 127.0.0.
```
The code that creates the command line array from the line in the mass testing file doesn't recognize "html output.html" as a single argument.

Forbidding the mass testing file from including options that specify a file name avoids the problem.

drwetter:
Hi @dcooper16,

thx for the PR.

As indicated in #753 though first I would love to check the approach sending a signal. This is less effort and seems more elegant to me. It'll be not as user friendly though. 

I am not sure either whether all syntax checks can be run upfront without executing. If not using the existing parser, it boils down writing a second one.

Sending a signal from child to parent however let the client do all the parsing while executing. If one child stumbles over one line the parent will be killed -- under some circumstances like serial execution or writing into different files this could be handled differently.

Cheers, Dirk

drwetter:
Thx, David. But as indicated I think fixing user errors and the effort to do so seems not so good balanced to me.

PR #760, opened 2017-06-01
Fix #757

### LINKED ISSUE #757: BEAST and multiple IPs
By scanning a domain which resolves to more than one IP address, the first scanned IP is vulnerable to BEAST (TLSv1 and CBC ciphers) whereas the others (also TLSv1 and CBC ciphers) are marked with "no CBC ciphers found for any protocol (OK)". An example URL is github.com.

My request was: `./testssl.sh github.com`

dcooper16:
I believe I discovered the reason for issue #757: https://github.com/drwetter/testssl.sh/commit/f2303a0d79bc6f5c21181c0ed08ef5135abbeeeb.

This commit attempted to replace `$cbc_cipher_list_hex` (which was computed on the fly) with `$cbc_ciphers_hex` (which is static). However, the function was still using `$cbc_cipher_list_hex`, and since it wasn't being initialized to "" at the beginning of the function, the second call to `run_beast()` (to handle the second IP address) just appended to the value created by the first. Then, when the first two bytes were removed from the resulting string the result was a malformed cipher suite list, which caused `tls_sockets()` to fail.


drwetter:
I was just about to look into it. Thx for the fast reaction, David!

PR #775, opened 2017-07-03
run_hpkp() bug fix

dcooper16:
In `run_hpkp()` there is a call to `$OPENSSL s_client` that uses `${sni[i]}` as one of the command line options, but `sni` is not defined. My guess is that this was a copy/paste error from `run_client_simulation()`, which is the only function where an `sni` array is defined.

I am guessing that the intention was to use `$SNI` in `run_hpkp()`.


drwetter:
thx for spotting this

PR #782, opened 2017-07-11
yaSSL (used in MySQL) returns unsupported cipher instead of Alert

sdann:
yaSSL has a bug which will return a successful Server Hello message,
with a cipher that the server supports, even if none of the client
provided ciphers are supported.

This causes an infinite loop on many tests, which whittle down the
supported cipher list, by sending a full list in a loop, and
removing the successfully negotiated cipher from the list each time.

ex:
  run_pfs()
  cipher_pref_check()
  run_rc4()
  run_allciphers()

This is a proposed fix within run_rc4() that should apply to the other
situations as well.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1068,6 +1068,13 @@ is_ipv6addr() {
      return 0
 }
 
+contains() {
+     local array="$1[@]"
+     for item in "${!array}"; do
+          [[ $item == $2 ]] && return 0
+     done
+     return 1
+}
 
 ###### END helper function definitions ######
 
@@ -2401,6 +2408,17 @@ rfc2openssl() {
      return 0
 }
 
+# argv[1]: rfc cipher name
+# return: hexcode (ex: "00,33")
+rfc2hexcode() {
+     local hexcode=""
+     local -i i
+
+     for (( i=0; i < TLS_NR_CIPHERS; i++ )); do
+          [[ "$1" == "${TLS_CIPHER_RFC_NAME[i]}" ]] && hexcode="${TLS_CIPHER_HEXCODE[i]}" && break
+     done
+     echo "${hexcode:2:2},${hexcode:7:2}"
+}
 
 show_rfc_style(){
      local rfcname="" hexcode
@@ -10783,8 +10801,14 @@ run_rc4() {
 
      for (( success=0; success==0 ; 1 )); do
           ciphers_to_test=""
+          local -i nr_ciphers_to_test=0
+          local -a hex_ciphers_to_test
           for (( i=0; i < nr_nonossl_ciphers; i++ )); do
-               ! "${ciphers_found2[i]}" && ciphers_to_test+=", ${hexcode2[i]}"
+               if ! "${ciphers_found2[i]}"; then
+                    ciphers_to_test+=", ${hexcode2[i]}"
+                    hex_ciphers_to_test[nr_ciphers_to_test]=${hexcode2[i]}
+                    nr_ciphers_to_test+=1
+               fi
           done
           success=1
           if [[ -n "$ciphers_to_test" ]]; then
@@ -10801,6 +10825,13 @@ run_rc4() {
                     for (( i=0; i < nr_nonossl_ciphers; i++ )); do
                          [[ "$cipher" == "${rfc_ciph2[i]}" ]] && ciphers_found2[i]=true && break
                     done
+                    # Deal with yaSSL server bug
+                    cipher_hex=$(rfc2hexcode $cipher)
+                    if ! contains hex_ciphers_to_test $cipher_hex; then
+                         debugme echo "Server returned a cipher not in the client list: $cipher"
+                         rc4_offered=0
+                         break
+                    fi
                     i=${index[i]}
                     ciphers_found[i]=true
                     if "$WIDE" && ( [[ ${kx[i]} == "Kx=ECDH" ]] || [[ ${kx[i]} == "Kx=DH" ]] || [[ ${kx[i]} == "Kx=EDH" ]] ); then


sdann:
@drwetter This patch is my proposal for catching and dealing with TLS servers (yaSSL in this instance) that don't return an Alert when a cipher can't be negotiated. I have also opened a bug with yaSSL to fix their side, but testssl.sh will still have to deal with older server versions.

The most generic fix is to confirm that the cipher in the Server Hello matches one sent in the Client Hello. If we check in the run_* tests, as I've done here, we'll need to dup this in 3 - 4 other places. Or I could push this down directly into tls_sockets().

Thoughts?

This is part of my work to get the MySQL STARTTLS sockets implementation working.

drwetter:
Hello @sdann ,

thx. "Sounds weird" (Server Hello) was my first reaction. Is there any publicly available server out there you know of so that I can see myself what's happening (DM/PM/..)?

Cheers, Dirk

dcooper16:
Hi @sdann,

Thanks for pointing this out. I guess this is something we need to check for in each of the functions, but I believe there is an easier solution. Here is the relevant section of code from `run_cipher_match()`:
```
                    [[ $sclient_success -ne 0 ]] && [[ $sclient_success -ne 2 ]] && break
                    cipher=$(get_cipher "$TEMPDIR/$NODEIP.parse_tls_serverhello.txt")
                    for (( i=bundle*bundle_size; i < end_of_bundle; i++ )); do
                         [[ "$cipher" == "${rfc_ciph2[i]}" ]] && ciphers_found2[i]=true && break
                    done
                    [[ $i -eq $end_of_bundle ]] && break
                    i=${index[i]}
                    ciphers_found[i]=true
```
The `for` loop is already comparing the ServerHello cipher to all of the ciphers included in the ClientHello. So, if the loop gets to the end without finding a match, then we already know that your `contains hex_ciphers_to_test $cipher_hex` test would fail. Similarly, we can be sure the test would pass if a match is found in the loop. I included the ` [[ $i -eq $end_of_bundle ]] && break` check in `run_cipher_match()`, but didn't include it everywhere, since I thought it was an unnecessary check.

If desired, we could note the error in some way rather than just silently breaking out of the loop.

sdann:
@dcooper16 

I originally thought this too and my first fix did this check. However, it's not a complete fix, because $rfc_ciph2 is not a list of the ciphers sent in the Client Hello, but rather a list of ciphers the client supports. $ciphers_to_test is the list of ciphers sent in the Client Hello.

Thus, the server can (and in this case does) support some of the ciphers which the client supports, but not the ciphers which were sent in the Client Hello.

For instance, the client may send only RC4 base ciphers and the server will reply with TLS_DHE_RSA_WITH_AES_128_CBC_SHA, which is in $rfc_ciph2. That cipher will again be marked true and the loop will never terminate.

dcooper16:
Hi @sdann ,

You are partially correct. My original suggestion wouldn't work. `$rfc_ciph2` is not a list of the ciphers sent in the ClientHello, but it is not a list of the ciphers the client support either. It is a list of the ciphers to be tested; so, in the case of `run_rc4()` it only contains RC4 ciphers. The problem, however, is that `$rfc_ciph2` includes ciphers that were detected in a previous iteration of the loop and so aren't in the current ClientHello.

So, I believe the solution is to add two new checks: one to verify that the cipher selected by the server is in `$rfc_cipher2` and a second to verify that the selected cipher hadn't already been detected. So, for the `tls_sockets()` portion of `run_rc4()` the code would be changed to the following:
```
               if [[ $ret -eq 0 ]] || [[ $ret -eq 2 ]]; then
                    cipher=$(get_cipher "$TEMPDIR/$NODEIP.parse_tls_serverhello.txt")
                    for (( i=0; i < nr_nonossl_ciphers; i++ )); do
                         [[ "$cipher" == "${rfc_ciph2[i]}" ]] && ciphers_found2[i]=true && break
                    done
# Make sure that something was found in the loop:
                    [[ $i -eq $nr_nonossl_ciphers ]] && break
                    i=${index[i]}
# If the cipher in the ServerHello had already been found, then it wasn't in the ClientHello.
                    "${ciphers_found[i]}" && break
                    ciphers_found[i]=true
# Since some servers return a ServerHello even with a cipher not in the ClientHello,
# don't set "success" or "rc4_offered" until after verifying that the cipher selected by
# the server was in the ClientHello.
                    success=0
                    rc4_offered=1
                    if "$WIDE" && ( [[ ${kx[i]} == "Kx=ECDH" ]] || [[ ${kx[i]} == "Kx=DH" ]] || [[ ${kx[i]} == "Kx=EDH" ]] ); then
                         dhlen=$(read_dhbits_from_file "$TEMPDIR/$NODEIP.parse_tls_serverhello.txt" quiet)
                         kx[i]="${kx[i]} $dhlen"
                    fi
                    "$WIDE" && "$SHOW_SIGALGO" && [[ -r "$HOSTCERT" ]] && \
                         sigalg[i]="$($OPENSSL x509 -noout -text -in "$HOSTCERT" | awk -F':' '/Signature Algorithm/ { print $2 }' | head -1)"
               fi
```

sdann:
@dcooper16

Yeah, I went through that iteration as well and determined I'm adding 2 special case checks, for 1 odd server error, to 4 different loops in 4 different functions, each with slightly different variable names. That's a lot of special case code.

Instead, logically, I want to compare ciphers sent to ciphers returned. Diff the lists and raise an error, simple. So that's why I'm suggesting the above patch as the cleanest solution. Though, I'm open to pushing it into tls_sockets() directly, so there's only 1 implementation over 4.

dcooper16:
> Yeah, I went through that iteration as well and determined I'm adding 2 special case checks, for 1 odd server error, to 4 different loops in 4 different functions, each with slightly different variable names. That's a lot of special case code.

It would only be adding two simple lines of code to each loop. Your proposed adds many more lines of code to each loop, so I don't see your proposal as being a simpler approach.

In addition, your proposal adds two new loops to check iteration of the loop in question: one in the call to `rfc2hexcode()` and one in the call to `contains()`. That's a lot more work than just comparing two integers and checking the value of a boolean.

> Instead, logically, I want to compare ciphers sent to ciphers returned. Diff the lists and raise an error, simple. So that's why I'm suggesting the above patch as the cleanest solution. Though, I'm open to pushing it into tls_sockets() directly, so there's only 1 implementation over 4.

Adding a check to `tls_sockets()` might be the best way to go. If it were just functions such as `run_pfs()`, `cipher_pref_check()`, `run_rc4()`, and `run_allciphers()`, I think the solution should be the one I suggested above, since it is the most efficient. The problem is other places where `tls_sockets()` is used where there isn't such an efficient way to check that the cipher in the ServerHello was in the ClientHello: `std_cipherlists()`, `run_sweet32()`, `run_ssl_poodle()`, `run_freak()`, `run_logjam()`, and `run_lucky13()`.

I'll work on a PR to change `tls_sockets()` to check for this scenario. Unfortunately, since this will require `tls_sockets()` looping through all of the ciphers in the ClientHello it will add to the running time of `tls_sockets()`. It'll be up to @drwetter to decide whether the cost of this extra check is worth the benefit.

drwetter:
> It'll be up to @drwetter to decide whether the cost of this extra check is worth the benefit.

you know my general stance. :-) There's still the issue that sockets are slower and we haven't figured out yet why <sup>1)</sup>. There are some suspects though. 

To be honest at the moment I can't tell whether it is worth it. I see @sdann's point for supporting MySQL which is a good thing but I am wondering about the costs in terms of speed but also readability. The latter seems to better addressed by @dcooper16 .

Before jumping into this or that I need to understand how widespread this bug/<whateveryoucallit> is. Any insights would be appreciated.

 <sup>1)</sup> @dcooper16 If you feel like experimenting a bit: there's a ``sockread_fast()`` function hidden which should save 2x ``fopen()`` and write().

sdann:
My preference is readability over performance, especially for an error check. We're certainly adding another loop, but it's over a list of 10 - 200 items, so it should be inconsequential.

yaSSL is definitely deprecated and decreasing in use, but it's the default in the Community Edition of MySQL since at least version 5.5 (2009) I believe. So millions of MySQL installs. The WolfSSL website also claims their products (WolfSSL, Cyassl, yassl all have a shared history) ared used in OpenWRT, cURL, and some versions of Ubuntu.

https://www.wolfssl.com/wolfSSL/Products-yassl.html
https://www.wolfssl.com/wolfSSL/About.html

dcooper16:
Hi @drwetter ,

While I was in the middle of responding to @sdann's earlier message, it occurred to me that this bug wouldn't affect just the four functions that he mentioned, where the result would be an infinite loop, but that it would also result in false positives in the six functions that I listed. That's when I came to the conclusion that the only way to properly address this server bug would be to do so in `tls_sockets()`.

I wasn't happy about the idea of adding extra cost to `tls_sockets()`, but, on my computer at least, the cost of the check in PR #784 is fairly small: less than 0.006 seconds per call to `tls_sockets()`. I understand there's still the question of whether any changes should be made to deal with this one implementation's bug, but I figured it would best to have a PR to look at in making that determination. Much easier to determine the costs in terms of speed and readability if there is code available.

sdann:
I've tested @dcooper16 proposed fix in PR #784 and it works well in more cases. I'm closing this PR in favor of accepting #784 as the best fix.

drwetter:
What I still do not understand how clients are dealing with this. If a client sends cipher set A, server hello contains a cipher which the client doesn't offer, the TLS connection should fail some time later.

I cannot believe this is a general wolfSSL/yassl bug but more a bug which happens for conditions like MySQL and STARTTLS.

What do you think?

drwetter:
> I've tested @dcooper16 proposed fix in PR #784 and it works well in more cases. I'm closing this PR in favor of accepting #784 as the best fix.

ACK.

sdann:
@drwetter 

I only have a MySQL implementation of yaSSL at the moment, so I can't say whether it's generic yet. I *may* be able to test an OpenWRT server I have this evening.

I have some pcaps of **openssl s_client** hitting the same issue, when testing against my MySQL instance.

And, since yaSSL supports the most commonly used ciphers by default:

DHE-RSA-AES256-SHA
DHE-RSA-AES128-SHA
AES128-RMD
DES-CBC3-RMD
DHE-RSA-AES256-RMD
DHE-RSA-AES128-RMD
DHE-RSA-DES-CBC3-RMD
AES256-SHA
RC4-SHA
RC4-MD5
DES-CBC3-SHA
DES-CBC-SHA
EDH-RSA-DES-CBC3-SHA
EDH-RSA-DES-CBC-SHA
AES128-SHA:AES256-RMD

It's likely that real clients usually always successfully negotiate a valid cipher like AES256-SHA, and never run into issues of having incompatible cipher lists.

PR #774, opened 2017-06-30
Add mysql (openssl) starttls support

sdann:
openssl/master branch now supports mysql STARTTLS in s_client

This patch adds support to call and use that s_client support to run
most, but not all (pfs, client simulation) tests.

The socket implementation is stubbed, but not yet functional.

----

@petermosmans: you'll need to pull the following 2 patches:

https://github.com/openssl/openssl/commit/a2d9cfbac5d87b03496d62079aef01c601193b58
https://github.com/openssl/openssl/commit/8530039a307f7aa8acb0516fdd38191baa91d434

----
@drwetter I've got a follow-up patch stack adding socket support, but since MySQL Community Edition uses yaSSL instead of OpenSSL, I'm finding a lot of small client/server compatibility issues. I'm working through them one by one, but wanted to get the s_client version to you first, as it's useful by itself.

PR #783, opened 2017-07-11
Add mysql (sockets) starttls support

sdann:
Following up on PR #774, here's the simple socket implementation of MySQL STARTTLS.

With this addition, several new tests will run against MySQL. However, there are still some bugs being worked out against the Community Edition (yaSSL), which are being addressed in PR #782.


drwetter:
Thanks, @sdann

PR #784, opened 2017-07-11
Handle server returning unsupported cipher

dcooper16:
As reported in #782, some servers will return a ServerHello with a cipher not listed in the ClientHello rather than than return an Alert, if the server does not support any of the ciphers listed in the ClientHello.

This PR modifies `tls_sockets()` to check whether the cipher in the ServerHello was one included in the ClientHello and to fail if it wasn't.

@sdann : Could you try this out on your yaSSL to verify that it addresses the problem.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7607,9 +7607,13 @@ check_tls_serverhellodone() {
 # arg1: ASCII-HEX encoded reply
 # arg2: (optional): "all" -  process full response (including Certificate and certificate_status handshake messages)
 #                   "ephemeralkey" - extract the server's ephemeral key (if any)
+# arg3: (optional): CIPHER_SUITES string (lowercase, and in the format output by code2network())
+#       If present, parse_tls_serverhello() will check that the cipher in the ServerHello appears in
+#       the CIPHER_SUITES string.
 parse_tls_serverhello() {
      local tls_hello_ascii="$1"
      local process_full="$2"
+     local cipherlist="$3"
      local tls_handshake_ascii="" tls_alert_ascii=""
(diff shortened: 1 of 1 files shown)


drwetter:
Thx, @dcooper16 . As said in #782 : _Before jumping into this or that I need to understand how widespread this bug/<whateveryoucallit> is. Any insights would be appreciated._

sdann:
@dcooper16 Tested and works well!

I was hitting a few other issues in tests like "NULL ciphers" that are now gone with this patch. I think this is definitely the right approach.

drwetter:
Thanks @dcooper16 for the fix and @sdann for detection.

In the end I think it'll make testssl.sh also more robust. This seems to be an edge case though -- I did a test with yassl and luci (web server used for OpenWRT) and there I spotted nothing suspicious. 

I will add later a comment in the code here why you did this, David. For me it would make sense also  in non-debug mode to tell the users that there's a problem with the server side TLS implementation. Question is where. Repeatedly in ``parse_tls_serverhello()`` mangles the output too much probably. 

OTOH I have no idea though/yet how often the message would appear if the ``debugme`` would be removed as I still need to check this with yaSSL + MySQL myself. I guess for Steven that would be easy to tell.

sdann:
With a "secure" configuration of MySQL (only 4 ciphers supported) the error is caught 9 times on a full run. I'm hitting the error in the following checks (sometimes twice per check):

NULL ciphers (no encryption)
Weak 128 Bit ciphers (SEED, IDEA, RC[2,4])
Strong encryption (AEAD ciphers)
PFS is offered
Cipher order
RC4 (CVE-2013-2566, CVE-2015-2808)
Testing 359 ciphers

We're likely to hit the error more often on the default MySQL config, where ~20 ciphers are supported.

PR #786, opened 2017-07-12
Make debug levels in parse_tls_serverhello() consistent

dcooper16:
A [commit](https://github.com/drwetter/testssl.sh/commit/e8b5a82c7e9c082396e2d016b870d4beae352427) on May 15 changed `parse_tls_serverhello()` so that summary information about the contents of the server's response is only displayed if `$DEBUG` is at least 3. However, some lines were not changed so that some output is still created if `$DEBUG` is 2.

If `$DEBUG` is 2, then the header "TLS message fragments:" is displayed, but the messages fragments are not. In addition, if `$DEBUG` is 2, alert messages are displayed, but handshake messages are not.

This PR makes `parse_tls_serverhello()` consistent by having it only print information about message fragments, alert messages, and handshake messages if `$DEBUG` is at least 3.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7662,7 +7662,7 @@ parse_tls_serverhello() {
      # byte 3+4:    fragment length
      # bytes 5...:  message fragment
      tls_hello_ascii_len=${#tls_hello_ascii}
-     if [[ $DEBUG -ge 2 ]] && [[ $tls_hello_ascii_len -gt 0 ]]; then
+     if [[ $DEBUG -ge 3 ]] && [[ $tls_hello_ascii_len -gt 0 ]]; then
           echo "TLS message fragments:"
      fi
      for (( i=0; i<tls_hello_ascii_len; i=i+msg_len )); do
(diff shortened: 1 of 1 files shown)


drwetter:
I am fine with the fragment message (it's only ``echo "TLS message fragments:"`` anyway)  to show at debug level >=3 -- the first but for the alerts it looks better to me that they stay @ level 2.

Now e.g.

``` LOW: 64 Bit + DES encryption (w/o export)     sending client hello...
reading server hello...
TLS message fragments:
TLS alert messages:
     tls_err_descr:          0x28 / = 40 (handshake failure)
     tls_err_level:          02 (fatal)

  (returned 2 lines)  
not offered (OK) -- LOW:DES:!ADH:!EXP:!NULL
```
where ``TLS message fragments:`` is redundant. Your suggestion:

```
reading server hello...
TLS message fragments:
     protocol (rec. layer):  0x0303
     tls_content_type:       0x15 (alert)
     msg_len:                2

TLS alert messages:
     tls_err_descr:          0x28 / = 40 (handshake failure)
     tls_err_level:          02 (fatal)

  (returned 2 lines)  
not offered (OK) -- LOW:DES:!ADH:!EXP:!NULL
```

As indicated my suggestion would be to tell about the handshake failure.


Or message fragments (all of them) go to >=4 and alert >=3.

drwetter:
Thx David. I reviewed the output for the whole program and debug level 2 so that the consistency for this is better now.

PR #787, opened 2017-07-13
Fix logic in has_server_protocol()

sdann:
If $PROTOS_OFFERED was empty, and thus the protocols unknown, this
function would return true for any protocol you passed it. This caused
most callers to assume TLS1.0, even if the server didn't offer it.

Instead return false and make the caller do an extra lookup.


drwetter:
yes, sure. Thx for catching

PR #788, opened 2017-07-13
Fix CCS Injection detection for MySQL (yaSSL)

sdann:
Another yaSSL server incompatibility. MySQL protects against CCS
Injection by erroring if it sees two CCS requests before the handshake
completes. But instead of returning a TLS alert, it seems to error up
the stack to MySQL which returns its own error message.

Debug output looks like:

```
sending client hello,
reading server hello

1st reply:

sending payload #2 with TLS version x03, x02:
tls_content_type: 16 | tls_protocol: 0000 | byte6: 04

2nd reply:
00000000  16 00 00 02 ff 13 04 23  30 38 53 30 31 42 61 64  |.......#08S01Bad|
00000010  20 68 61 6e 64 73 68 61  6b 65                    | handshake|
0000001a

test failed, probably read buffer too small (16000002FF1304)
```

This patch adds a custom check for this MySQL specific error, as far
down in the error check path as I can.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -9162,7 +9162,7 @@ run_ccs_injection(){
           fi
           ret=0
      elif [[ "$byte6" == "15" ]] && [[ "${tls_hello_ascii:0:4}" == "1503" ]]; then
-          # decyption failed received
+          # decryption failed received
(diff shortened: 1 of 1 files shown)


drwetter:
thank you, @sdann

PR #792, opened 2017-07-14
Make sure last ClientHello extension is not empty

dcooper16:
According to a discussion thread on the IETF TLS WG mail list (see https://www.ietf.org/mail-archive/web/tls/current/msg19720.html), there is at least one TLS server that will fail if the last extension in the ClientHello contains extension_data of length 0.

Currently, `tls_sockets()` will create such a ClientHello if:
* The padding extension is included, and the length of the ClientHello without the padding data would be between 508 and 511 bytes.
* No padding extension is included, and the caller provides `$extra_extensions` in which the last extension in `$extra_extensions` is empty.
* No padding extension is included, `$extra_extensions` is empty, no ECC cipher suites are offered, and the ClientHello is for TLSv1.1 or below (in this case the next protocol extension would be that last one).

This PR avoids the server bug by ensuring that the padding extension (when present) always contains at least one byte, and by ensuring that when the padding extension is not present that the (non-empty) heartbeat extension is the last extension.

This PR does leave one possible scenario in which the last extension would be empty. If the caller provides an `$extra_extensions` in which the last extension in `$extra_extensions` is empty, `tls_sockets()` does not add a padding extension (or a padding extension is included in `$extra_extensions`), and `$extra_extensions` includes a heartbeat extension, then the last extension in the ClientHello would be empty. This, however, is a highly unlikely scenario, and certainly there are currently no such calls to `tls_sockets()` in testssl.sh. (Currently, whenever a call to `$tls_sockets()` includes extra extensions, the final extension is non-empty, and no call to `$tls_sockets()` includes the heartbeat extension in the extra extensions).

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -8625,10 +8625,7 @@ socksend_tls_clienthello() {
                ,00, $len_servername_hex # server_name length. We assume len(hostname) < FF - 9
                ,$servername_hexstr"     # server_name target
           fi
-          if [[ ! "$extra_extensions_list" =~ " 000f " ]]; then
-               [[ -n "$all_extensions" ]] && all_extensions+=","
-               all_extensions+="$extension_heartbeat"
-          fi
+
           if [[ ! "$extra_extensions_list" =~ " 0023 " ]]; then
                [[ -n "$all_extensions" ]] && all_extensions+=","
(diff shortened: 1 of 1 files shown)


drwetter:
Thx for tracking this and thanks again for the workaround!

PR #795, opened 2017-07-19
Revert "small performance improvement for prepare_arrays() by replaci…

sdann:
...ng grep + awk by awk only"

This reverts commit 3b1638f60309b1c4e9636c6f5b62c71b22ff4cd1.

I'm not sure what went wrong here, but these commands are not equivalent.

```
$ export hexc=0x00,0x05 && openssl ciphers -tls2 -V 'ALL:COMPLEMENTOFALL:@STRENGTH' | grep -w "$hexc" -
          0x00,0x05 - RC4-SHA                 SSLv3 Kx=RSA      Au=RSA Enc=RC4(128)  Mac=SHA1
$ export hexc=0x00,0x05 && openssl ciphers -tls2 -V 'ALL:COMPLEMENTOFALL:@STRENGTH' | grep -w "$hexc" - | awk '{ print $3 }'
RC4-SHA
$ export hexc=0x00,0x05 && openssl ciphers -tls2 -V 'ALL:COMPLEMENTOFALL:@STRENGTH' | awk '/\<'"$hexc"'\>/ { print $3 }'
$
```

The old command finds my RC4-SHA cipher, the new one does not.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -11417,7 +11417,7 @@ prepare_arrays() {
                     if [[ $OSSL_VER_MAJOR -lt 1 ]]; then
                          [[ ":${ossl_supported_tls}:" =~ ":${TLS_CIPHER_OSSL_NAME[i]}:" ]] && TLS_CIPHER_OSSL_SUPPORTED[i]=true
                     else
-                         ossl_ciph="$(awk '/\<'"$hexc"'\>/ { print $3 }' <<< "$ossl_supported_tls")"
+                         ossl_ciph="$(grep -w "$hexc" <<< "$ossl_supported_tls" | awk '{ print $3 }')"
                          if [[ -n "$ossl_ciph" ]]; then
                               TLS_CIPHER_OSSL_SUPPORTED[i]=true
                               [[ "$ossl_ciph" != "${TLS_CIPHER_OSSL_NAME[i]}" ]] && TLS_CIPHER_OSSL_NAME[i]="$ossl_ciph"


drwetter:
Can reproduce it -- under FreeBSD. That doesn't have the 'word grep' in awk as assumed. 

What version of awk do you have?

sdann:
$ awk -W version
mawk 1.3.3 Nov 1996, Copyright (C) Michael D. Brennan

compiled limits:
max NF             32767
sprintf buffer      2040
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.2 LTS"
$

drwetter:
Thx, @sdann -- also for catching this.

I found a better solution. It looked to me like the "word grep" is not needed at all. @dcooper16 : any concerns? 

> $ awk -W version
mawk 1.3.3 Nov 1996, Copyright (C) Michael D. Brennan

mawk... never heard of it :-)

The awk I have here under all Ubuntu versions is still gawk but mawk is installed (and not used because of ``/etc/alternatives/awk`` points to the gnu edition).

sdann:
Yeah, I don't know how mawk got configured. I have a fairly standard Ubuntu desktop setup. My favorite clip is from the man page:

`$ man mawk`
```
BUGS
...
       Implementors of the AWK language have shown a consistent lack of imagination when naming their programs.
```

dcooper16:
> @dcooper16 : any concerns?

My knowledge of awk is extremely limited, so I don't understand any of the possible solutions.

However, I tried the code on a couple of platforms and it seemed to work.

drwetter:
> My knowledge of awk is extremely limited, so I don't understand any of the possible solutions.

Sorry, my question was not very precise. I was wondering why you picked ``grep -w`` instead of ``grep ``. I couldn't find a situation where ``grep`` shouldn't suffice.

drwetter:
> My favorite clip is from the man page: [..]

LOL . You know K+R  C? ;-)  I think that was a sprit of the time

dcooper16:
> I was wondering why you picked `grep -w` instead of `grep`.

I believe you are correct that the `-w` wasn't needed. It would have been necessary if the string being searched contained both SSLv2 and non-SSLv2 ciphers (e.g., since "0x00,0xC0" would have found both "0x00,0xC0" and "0x07,0x00,0xC0"), but since the string only contains non-SSLv2 ciphers, it seems that it wasn't needed.

drwetter:
"0x00,0xC0" is still a word in "0x07,0x00,0xC0". Seems I've missed that too while swapping ``grep -w`` with the awk expression.

Issue #793, opened 2017-07-14
MySQL (yaSSL) returns unsupported cipher instead of Alert


sdann:
yaSSL has a bug which will return a successful Server Hello message, with a cipher that the server supports, even if none of the client provided ciphers are supported.

This was first discussed in PR #782 and partially fixed by @dcooper16 in PR #784.

Opening a new Issue to track additional problems this server side bug causes.

I've also opened a bug with MySQL to fix the issue in their server at: https://bugs.mysql.com/87078

sdann:
With #784 testssl.sh now detects if the server has returned an incorrect cipher and raises an error. Unfortunately, this short circuit behavior is causing some false successes to be reported. There may be several, but I've found the first in run_rc4().

With the default MySQL (yaSSL) configuration 2 RC4 ciphers are supported:

RC4-SHA
RC4-MD5

However, the -4 test currently returns "no RC4 ciphers detected (OK)" against this configuration.

This is because, before those 2 ciphers can be negotiated, the server bug is hit, with the client sending:

Cipher Suite: TLS_ECDHE_PSK_WITH_RC4_128_SHA (0xc033)

and the server responding with:

Cipher Suite: TLS_DHE_RSA_WITH_AES_128_CBC_SHA (0x0033)

Causing the loop to terminate with success.

Two possible fixes:
1) Iterate over every cipher 1 at a time.
    - much more expensive in # of connections
    - 1 off fix for this server issue
2) Detect tls_sockets() returns 1 and mark the test output as "UNKNOWN" or "SERVER ERROR".

dcooper16:
Hi @sdann,

I think we'd need more information to understand what is going on here, as `run_rc4()` doesn't work as you describe. The client doesn't send TLS_ECDHE_PSK_WITH_RC4_128_SHA (0xc033), it sends a list of every RC4 cipher. So, if the client is sending a list of every RC4 cipher and the server responds with TLS_DHE_RSA_WITH_AES_128_CBC_SHA (0x0033), that either means:

- The server doesn't support any RC4 ciphers; or 

- The server is responding with a cipher not in the ClientHello even though it supports one of the ciphers in the ClientHello.

[...]

PR #796, opened 2017-07-20
Catch MySQL (yaSSL) server bug when testing standard cipher categories

sdann:
Unfortunately, most categories were passing incorrectly, since a
`tls_socket()` error was interpreted as a success for many of the cases.
Now we explicitly check for the known yaSSL failure and raise a warning
that the test is inconclusive.

New output against --starttls=mysql looks like:

```
Testing ~standard cipher categories
-----------------------------------

 NULL ciphers (no encryption)                  SERVER_ERROR: test inconclusive due to MySQL Community Edition (yaSSL) bug.
 Anonymous NULL Ciphers (no authentication)    SERVER_ERROR: test inconclusive due to MySQL Community Edition (yaSSL) bug.
 Export ciphers (w/o ADH+NULL)                 SERVER_ERROR: test inconclusive due to MySQL Community Edition (yaSSL) bug.
 LOW: 64 Bit + DES encryption (w/o export)     offered (NOT ok)
 Weak 128 Bit ciphers (SEED, IDEA, RC[2,4])    SERVER_ERROR: test inconclusive due to MySQL Community Edition (yaSSL) bug.
 Triple DES Ciphers (Medium)                   offered
 High encryption (AES+Camellia, no AEAD)       offered (OK)
 Strong encryption (AEAD ciphers)              SERVER_ERROR: test inconclusive due to MySQL Community Edition (yaSSL) bug.
```

Partial workaround for #793

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2294,59 +2294,72 @@ std_cipherlists() {
                     fi
                fi
           fi
-          case $3 in
-               2)  if [[ $sclient_success -eq 0 ]]; then
-                         # Strong is excellent to offer
-                         pr_done_best "offered (OK)"
-                         fileout "std_$4" "OK" "$2 offered"
-                    else
-                         pr_svrty_medium "not offered"
-                         fileout "std_$4" "MEDIUM" "$2 not offered"
-                    fi
-                    ;;
+          if [[ $sclient_success -ne 0 ]] && $BAD_SERVER_HELLO_CIPHER; then
+               # If server failed with a known error, raise it to the user.
+               if [[ $STARTTLS_PROTOCOL == "mysql" ]]; then
+                    pr_warning "SERVER_ERROR: test inconclusive due to MySQL Community Edition (yaSSL) bug."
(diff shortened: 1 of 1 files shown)


drwetter:
Hi @sdann,

basically good but I am wondering what we do with the other places where this inconclusiveness occurs. 

Cheers, Dirk

sdann:
@drwetter 
I put a status of what's fixed and what's still inconclusive at the end of #793. Basically, with this patch in, the remaining 3 areas that still have problems are:

* PFS is offered
* Cipher order
* Testing 359 ciphers

In each of these cases, a partial list of ciphers is still printed, that is correct. It's just the full list, is not printed, because the handler functions hit the YaSSL bug part way through.

I think that's sufficient for now, as I'd rather see a partial (but correct) list, than an error for those 3 tests.

sdann:
@drwetter, will you accept this patch as a reasonable user notification for the main issues?

drwetter:
yes, but give me a few days.

sdann:
Sounds good. Thanks.

drwetter:
Thank you, @sdann

PR #800, opened 2017-07-25
Fix client simulations with SSLv2 ClientHello

dcooper16:
This PR fixes the extraction of the cipher suites in SSLv2 ClientHellos in `client_simulation_sockets()`. Since `client_simulation_sockets()` can only handle SSLv3 and above ServerHellos, it removes any SSLv2 ciphers and converts the other cipher from 3-byte format to 2-byte format.


drwetter:
Thank you, it works as promised on my testcases

Issue #759, opened 2017-06-01
Indicator whether $PROTOS_OFFERED ...


drwetter:
.. is complete otherwise the usage of ``has_server_protocol()`` is limited, see #758

PR #801, opened 2017-07-25
std_cipherlists() and has_server_protocol()

dcooper16:
`std_cipherlists()` uses `has_server_protocol()` to determine whether to test for ciphers using SSLv2. However, this was resulting in false negatives due to #759. This PR removes the `has_server_protocol()` check so that SSLv2 is checked whenever the connection attempt with the TLSv1.2 ClientHello failed.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2276,7 +2276,7 @@ std_cipherlists() {
                sclient_success=$?
                [[ $sclient_success -eq 2 ]] && sclient_success=0
           fi
-          if [[ $sclient_success -ne 0 ]] && has_server_protocol "ssl2"; then
+          if [[ $sclient_success -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


drwetter:
Hi David,

I can follow you here partly:

> This PR removes the ``has_server_protocol()`` check so that SSLv2 is checked whenever the connection attempt with the TLSv1.2 ClientHello failed.

Admittedly ``has_server_protocol()`` is somewhat dubious. It currently only works the protocol check has been done. That's need to be fixed, that's why #759 is there.

All non-SSLv2 handshakes are done before your fixed line. Next come SSLv2 handshakes. Your PR would fix situations where PROTOS_OFFERED is not complete, SSLv2 is offered but not in PROTOS_OFFERED yet.

Is it that what you mean?

dcooper16:
> Your PR would fix situations where PROTOS_OFFERED is not complete, SSLv2 is offered but not in PROTOS_OFFERED yet.

Yes.

PR #807, opened 2017-07-28
Fix #806

### LINKED ISSUE #806: TLS12_CIPHER and ClientHello cipher limit bug
I just encountered a server that is failing the `run_protocols()` test. For TLS 1.2 it reports:
```
 TLS 1.2    not offered -- connection failed rather than downgrading to TLSv1.0
```
After some further testing, it seems that this server fails with an "illegal parameter" alert if the ClientHello contains more than 128 cipher suites. Since `$TLS12_CIPHER` contains 137 ciphers suites, this server is returning an error rather than a TLSv1.0 ServerHello for the TLS 1.2 test in `run_protocols()`. This error is presumably also affecting other tests that use  `$TLS12_CIPHER`, such as `determine_tls_extensions()`.

Could a few ciphers be removed from `$TLS12_CIPHER` so that it has no more than 128 ciphers? The OpenSSL that comes with testssl.sh only includes 112 ciphers in its ClientHello and other versions of OpenSSL include far fewer.

dcooper16:
This PR provides a partial fix of #806 by modifying `$TLS12_CIPHER` so that it includes only 128 ciphers. It also updates `$TLS12_CIPHER` by adding in some newer ciphers.

This PR also makes the test for TLS 1.2 in `run_protocols()` a little more robust by trying to connect with a second set of 127 ciphers (`$TLS12_CIPHER_2ND_TRY`) if no connection was established using `$TLS12_CIPHER`.

This is only a partial fix for #806, since `$TLS12_CIPHER` is used in places other than `run_protocols()` and it does not address for the other places in which `$TLS12_CIPHER` is used the rare cases in which the server does not support any of the ciphers offered in `$TLS12_CIPHER`.


drwetter:
Works without problems and brings it another step fwd. Thank you, David.

drwetter:
Hi David,

while working on #1211 : 

> This PR provides a partial fix of #806 by modifying $TLS12_CIPHER so that it includes only 128 ciphers. It also updates $TLS12_CIPHER by adding in some newer ciphers.

Shouldn't that be rather 127 ciphers (also for TLS12_CIPHER_2ND_TRY)?

Dirk

dcooper16:
> Shouldn't that be rather 127 ciphers (also for TLS12_CIPHER_2ND_TRY)?

Hi Dirk,

I don't have a server to test against, but I believe that $TLS12_CIPHER and $TLS12_CIPHER_2ND_TRY are okay. Both of these include 128 ciphers if the "00,ff" cipher is included in the count.

@PeterMosmans offered a tool for detecting servers that can't handle ClientHello messages with 128 or more ciphers at https://github.com/PeterMosmans/openssl/issues/28. In looking at the code and using Wireshark when running it, I see that when the code says that it trying _n_ ciphers it is really sending _n_+1 ciphers (_n_ real ciphers plus "00,ff"). Similarly, it will report that the 128 cipher limit bug was detected only if the connection failed when sending a ClientHello with 129 ciphers (128 real ciphers plus "00,ff"). The test code is counting the number of ciphers provided to the "-cipher" option of OpenSSL s_client rather than the number of ciphers in the actual ClientHello, and OpenSSL is adding "00,ff" to the list of ciphers specified in the "-cipher" option.

So, it seems that it is safe to send a ClientHello that includes 127 real ciphers plus "00,ff", which it what happens when $TLS12_CIPHER or $TLS12_CIPHER_2ND_TRY is used.

drwetter:
Hi David,

after some discussion we had I was believing 00FF is included in the 128 ciphers when the limit kicks in. Reading it again I think misinterpreted that.

The determination of SERVER_SIZE_LIMIT_BUG actually starts @ 127 ciphers (including 00FF) and does two steps. So that's safe.

@PeterMosmans tool is for sure an indicator but the icing on the cake would be if we would have a server to test against -- to be sure. I also can't tell whether all OpenSSL versions appended 00FF at that time. (this though I could test with an ancient version as you did).

I'll file a separate issue for clarification.

drwetter:
Thx btw!

PR #809, opened 2017-08-01
parse_tls_serverhello(), dh_bits, debug level 2

dcooper16:
Currently, when `$DEBUG` is 2 and the connection is successful, `parse_tls_serverhello()` prints out information about the server's ephermal (EC)DH key, but nothing else. For example:
```
sending client hello... reading server hello...
dh_bits:                ECDH, P-256, 256 bits
sending close_notify...
  (95 lines returned)
```
This PR changes `parse_tls_serverhello()` so that information about dh_bits is only displayed if `$DEBUG` is at least 3, making it the same as for other information about the server's response.

In addition, it indents the printing of the information about dh_bits in order to better align with other information displayed at  this debug level.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -8388,10 +8388,10 @@ parse_tls_serverhello() {
                     esac
                fi
                if [[ $dh_bits -ne 0 ]] && [[ $named_curve -ne 29 ]] && [[ $named_curve -ne 30 ]]; then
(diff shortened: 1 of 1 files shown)


drwetter:
Yeah, you're right.Thank you.

I am generally wondering though whether the logic here and in general costs time. Parameters, sometimes up to three, are checked, then only a debug statement is within the if-condition.

The other way round seems more effective (first check debugging mode, then the parameters). Whether it saves practically time -- not sure.

Issue #791, opened 2017-07-14
GREASE


dcooper16:
This is a follow up to #494 (which is about testing whether server's properly support version negotiation):
> Actually I thought it would be a good idea to reimplement a tolerance test for 2.9dev.

The IETF TLS WG adopted [draft-ietf-tls-grease](https://datatracker.ietf.org/doc/draft-ietf-tls-grease/) (Generate Random Extensions And Sustain Extensibility), which in some ways is an extension of the version tolerance test. The idea is to introduce unknown values into various places where the standard calls for servers to ignore unknown values in order to test that servers correctly ignore unknown values.

According to one [email](https://mailarchive.ietf.org/arch/msg/tls/k1KfZGwAGnZYgAsLN2sztRw5APc) on the TLS WG mail list, this technique has already been used to discover an implementation that would have failed a version tolerance test (based on TLSv1.3's new method of version negotiation).

I may be too early to implement [draft-ietf-tls-grease](https://datatracker.ietf.org/doc/draft-ietf-tls-grease/) (code points haven't been assigned yet), but we should consider whether testing of this sort should be added to testssl.sh at some point in the future.

drwetter:
To me it sounds certainly an interesting idea. Chrome/Chromium make use of it since last fall IIRC.

The test would be separate then and checking whether the server responds in a not compliant way (aborts, hangs etc) ?

dcooper16:
Yes, I think this should be a separate test. testssl.sh currently has tests to check for certain vulnerabilities (e.g., heartbleed), but otherwise it doesn't tend to check for implementation bugs that don't lead to security vulnerabilities. In fact, it does just the opposite; it silently tries to work around implementation bugs in order to correctly determine information about a server's configuration. For example, tests to determine what cipher suites a server supports are written to work around servers that have problems with ClientHello messages with more than 128 ciphers.

I think we should continue that idea; have other tests work around implementation bugs and have a separate test that looks for them.

drwetter:
Assigned this one to me too. I am just scratching my head whether one or the other test should make it into the default run. Everything would be too much as the test consumes too much time. 

Last but not least SSLlabs has also ~3 checks.

PR #814, opened 2017-08-04
GREASE

dcooper16:
This is an initial implementation of the idea I proposed in #791. It includes checks based on [draft-ietf-tls-grease](https://datatracker.ietf.org/doc/draft-ietf-tls-grease) as well as checks for specific implementation bugs that have previously been encountered.

This code needs testing. While I know of one server that will fail if the ClientHello contains more than 128 ciphers, I haven't been able to test any of the other code against any servers that have the tested-for bugs.

In addition, there is a need for polishing. The tests are referred to as `--grease`, which may not be a very user-friendly name. In addition, there is no output providing information about what tests are being performed. If a server fails a test, then a warning message is printed and is sent to `fileout()`. If the server passes all of the tests, then there is just a single output of "No bugs found."

At the moment, the code is not run by default. Unless `-g` or `--grease` is explicitly included in the command line, the code in this PR is not executed.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -638,6 +638,7 @@ fileout_json_section() {
            8) echo -e ",\n                    \"vulnerabilities\"   : [" ;;
            9) echo -e ",\n                    \"cipherTests\"       : [" ;;
           10) echo -e ",\n                    \"browserSimulations\": [" ;;
+          11) echo -e ",\n                    \"grease\"            : [" ;;
            *) echo "invalid section" ;;
      esac
 }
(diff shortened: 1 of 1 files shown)


drwetter:
Hi David,

thank you! This brings the project a good step forward. I merged it for now into the tls_robustness branch and not 2.9dev for two reasons: I currently have limited possibilities to check for non-https protocols and want to see first how it is doing against imap, smtp, pop, xmpp and the like. Secondly I plan to do a 2.9.5 snapshot release and want to keep this one for after the release.

For 2.9.5 snapshot I need to get 1-2 things off the table. Bear with me pls. I kept this open as a reminder.

Thx, Dirk

drwetter:
Good work!

PR #817, opened 2017-08-29
Use of fileout_insert_warning()

dcooper16:
This PR addresses the same issue as https://github.com/drwetter/testssl.sh/commit/6bb3494d9807991fd07b8ca0e120b218a0318dfd.  In its current form, `fileout_insert_warning()` cannot be used after `fileout_section_header()` has been called for the first time.


drwetter:
Oh, leads to invalid JSON too?

PR #818, opened 2017-08-30
Add OpenSSL names for ARIA ciphers

dcooper16:
A PR was just accepted into the master branch of https://github.com/openssl/openssl that specifies OpenSSL names for the ARIA GCM cipher suites: https://github.com/openssl/openssl/commit/bc32673869842c7f00ae7016040a612f516ead7e. This PR adds these OpenSSL names to the cipher-mapping.txt file. It also changes the description of the encryption algorithm for these ciphers from "ARIA" to "ARIAGCM" to be consistent with OpenSSL and with the other GCM ciphers in the cipher-mapping.txt file.

In addition, OpenSSL names for some of the ARIA CBC ciphers are provided in https://github.com/openssl/openssl/blob/master/doc/man1/ciphers.pod, and this PR adds those OpenSSL names to the cipher-mapping.txt file as well.


drwetter:
thx

PR #820, opened 2017-09-11
Fixed file output error in case of insecure redirect

seccubus:
In case of an insecure redirect a duplicate finding would be logged.

First finding would be HTTP_STATUS_CODE with severity HIGH and content ','.
Second finding would be HTTP_STATUS_CODE with severity INFO with the correct status code.

Now two findings will be created. 
1) insecure_redirect severity high with redirect URL
2) http_status_code with severity info and the redirect message


drwetter:
Thx!

Issue #829, opened 2017-09-22
OpenBSD


drwetter:
... is currently not very well supported.

1. seems that every s_client  call with LibreSSL makes problems because of the missing GOST engine
2. in the beginning there are some tput oddities:
```
tput: not enough arguments (3) for capability `setaf'
tput: not enough arguments (3) for capability `setaf'
tput: not enough arguments (3) for capability `setaf'
tput: not enough arguments (3) for capability `setaf'
tput: not enough arguments (3) for capability `setaf'
tput: not enough arguments (3) for capability `setaf'
tput: not enough arguments (3) for capability `setaf'
tput: not enough arguments (3) for capability `setaf'
tput: Unknown terminfo capability `0'
tput: Unknown terminfo capability `1'
```
3. a somewhat funny thing:
```

 Using "" [~0 ciphers]
 on _<redacted>_:/usr/bin/openssl
 (built: "", platform: "information not available")

```

PR #831, opened 2017-09-22
LibreSSL doesn't support "engine"

dcooper16:
LibreSSL prints an "invalid command" error for `openssl engine ...` regardless of what `...` is, but always returns 0. The result is that when using LibreSSL, the `initialize_engine()` function concludes that there is engine support even though there isn't. This results in error messages later on in testing.

This PR fixes the problem by checking whether `$OPENSSL engine -v` results in an "invalid command" error being send to `stderr` before trying any other tests with engine.

On the system in which I was testing, fixing the above problem still lead to a few places in which $OPENSSL printed a "can't open config file" warning, where the warning was not being redirected to `/dev/null` or to `$ERRFILE`. This PR also fixes that.

Hopefully this will address is problem noted in #829 that "every s_client call with LibreSSL makes problems because of the missing GOST engine."

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6356,7 +6356,7 @@ run_server_defaults() {
      # ciphers_to_test[7]: cipher suites using certificates with GOST R 34.10 (either 2001 or 94) public keys
      ciphers_to_test[1]=""
      ciphers_to_test[2]=""
-     for ciph in $(colon_to_spaces $($OPENSSL ciphers "aRSA")); do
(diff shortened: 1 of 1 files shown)


drwetter:
Hi David,

was just working on it, too. 

The thing with the engine test I borrowed and say thanks a lot :-) 
I didn't have that part yet.

The other parts weren't needed then. The openssl.conf was the culprit. Automagically also the cipher # was correct then. 

Thx! Dirk

PR #833, opened 2017-09-22
Suppress config file warning

dcooper16:
This may not be specific to LibreSSL, but just my local setup. However, when I test using LibreSSL testssl.sh still prints the following message several times:

     WARNING: can't open config file: /usr/local/etc/ssl/openssl.cnf

This PR suppresses the error message for several calls to $OPENSSL and so fixes the problem.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6509,7 +6509,7 @@ run_server_defaults() {
      # ciphers_to_test[7]: cipher suites using certificates with GOST R 34.10 (either 2001 or 94) public keys
      ciphers_to_test[1]=""
      ciphers_to_test[2]=""
-     for ciph in $(colon_to_spaces $($OPENSSL ciphers "aRSA")); do
+     for ciph in $(colon_to_spaces $($OPENSSL ciphers "aRSA" 2>>$ERRFILE)); do
           if grep -q "\-RSA\-" <<<$ciph; then
                ciphers_to_test[1]="${ciphers_to_test[1]}:$ciph"
           else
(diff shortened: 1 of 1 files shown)


drwetter:
Hi David,

I got the errer message too - using LibreSSL under Linux if I run it off the source directory. Isn't that because l and maybe you didn't do ``make install``? 

I remember seeing it also from folks compiled openssl under Mac OSX and probably run it off the source directory.

It boils down to the question whether we want to fix other people problems and beautify the output so that we are not to blame -- or rather not. It doesn't appear in HTML or JSON output (stderr).

*If* we decide to suppress this message I'd rather then put an additional warning upfront.

Opinions?

Thx, Dirk

dcooper16:
> I got the error message too - using LibreSSL under Linux if I run it off the source directory. Isn't that because l and maybe you didn't do make install?

Probably. I haven't tried running "make install," but it probably would create a config file in/usr/local/etc/ssl.

> It boils down to the question whether we want to fix other people problems and beautify the output so that we are not to blame -- or rather not. It doesn't appear in HTML or JSON output (stderr on screen).
> 
> If we decide to suppress this message I'd rather then put 1x additional warning upfront.
> 
> Opinions?

In this case I don't know what "other people's problems" refers to. I looked at the source code for LibreSSL, and the first thing the openssl program does is try to load the config file and print a warning if it can't. So, is the problem we'd be trying to fix that people are trying to use LibreSSL without providing it with a config file or that LibreSSL complains about not being able to open a config file even though it doesn't need one?

It seems that the general rule is that testssl.sh suppresses error messages from programs that it calls and then prints it own error/warning message, if appropriate. Error messages from `$OPENSSL pkey` are  suppressed since OpenSSL 0.9.8 doesn't support the `pkey` command, but there is an upfront warning against using such an old version of OpenSSL.

In this case, however, I don't see the reason for an upfront warning. If the warning from LibreSSL is unnecessary, why would we need to warn users about anything?

drwetter:
> In this case, however, I don't see the reason for an upfront warning. If the warning from LibreSSL is unnecessary, why would we need to warn users about anything?

I was about to say As a friendly reminder because the binary used isn't properly installed. In the context of testssl.sh it doesn't matter at all. We supply a config file overriding parts of that file and use options overriding configuration.

Thx!

PR #838, opened 2017-09-27
Remove extra space in BEAST output

dcooper16:
testssl.sh was inserting two spaces between the CBC ciphers detected by OpenSSL and those detected to `tls_sockets()`. This PR fixes the problem.


drwetter:
thx

PR #839, opened 2017-09-27
Make more use of has_server_protocol()

dcooper16:
This PR makes a few changes related to `add_tls_offered()`/`has_server_protocol()`.

* It changes `run_beast()` to check `has_server_protocol()` before running a test against the server when trying to determine whether the server supports SSLv3, TLSv1, TLSv1.1, and TLSv1.2.

* It has `run_server_defaults()` and `run_beast()` call `add_tls_offered()` whenever it detects that a protocol is offered (This makes `$PROTOS_OFFERED` more complete in the case that `run_protocols()` isn't run.) Note that in order for the results of the calls to `add_tls_offered()` in `cipher_pref_check()` not to be lost, the method of providing the input to the while loop needed to be changed.

* It changes `add_tls_offered()` and `has_server_protocol()` to use Bash string capabilities rather than grep in order to improve performance.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3968,13 +3968,12 @@ run_prototest_openssl() {
 # idempotent function to add SSL/TLS protocols. It should ease testing
 # PROTOS_OFFERED's content is in openssl terminology
 add_tls_offered() {
-     grep -qw "$1" <<< "$PROTOS_OFFERED" || PROTOS_OFFERED+="$1 "
+     [[ "$PROTOS_OFFERED" =~ "$1 " ]] || PROTOS_OFFERED+="$1 "
 }
 
 # function which checks whether SSLv2 - TLS 1.2 is being offereed
 has_server_protocol() {
-     [[ -z "$PROTOS_OFFERED" ]] && return 1  # if empty return 1, hinting to the caller to check at additional cost/connect
-     if grep -qw "$1" <<< "$PROTOS_OFFERED"; then
+     if [[ "$PROTOS_OFFERED" =~ "$1 " ]]; then
           return 0
      fi
      return 1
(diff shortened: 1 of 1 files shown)


drwetter:
Hi David,

thought I answered that but obviously not...

Changes are good, thx for the PR. However the ``has_server_protocol()`` isn't quite working yet. For BEAST probably a single test will or at least could fail. I haven't tried though.

PROTOS_OFFERED needs either a flag that all protocols have been tested or it needs a prefix/postfix for every protocol tested like "SSLv2,NO ... TLS_1.2,YES" <-- whatever is easier to parse.

While now writing it the second option makes more sense. 

Actually on several other occasions the completion of this mechanism would be quite useful and could save a few cycles.

Cheers, Dirk

dcooper16:
> Changes are good, thx for the PR. However the has_server_protocol() isn't quite working yet. For BEAST probably a single test will or at least could fail. I haven't tried though.

Yes, I understand that at the the moment if `has_server_protocol()` returns 0, then the server definitely supports the protocol, but if `has_server_protocol()` return 1, then it could either be that the server doesn't support the protocol or that support for the protocol has not yet been tested. That's why in this PR, if `has_server_protocol()` returns 1, a test is performed against the server to determine whether the protocol is supported.

> PROTOS_OFFERED needs either a flag that all protocols have been tested or it needs a prefix/postfix for every protocol tested like "SSLv2,NO ... TLS_1.2,YES" <-- whatever is easier to parse.

That seems like a good idea: have three possible return values for `has_server_protocol()`: yes, no, unknown.

drwetter:
@dcooper16 : I am working on ``has_server_protocol()``. Once it's done I'll happily merge this.

drwetter:
merged & thx, David!

PR #843, opened 2017-10-03
Fix incorrect client simulation data

dcooper16:
In the data provided by https://api.dev.ssllabs.com/api/v3/getClients, Chrome 57 Win 7 and Firefox 53 Win 7 send ClientHello messages that indicate support for TLSv1.3 draft 18, but the highest_protocol for each of these is specified as 0x0303. The result is that if the server being tested supports TLSV1.3 draft 18, `run_client_simulation()` will incorrectly report "No connection" for these servers since the DETECTED_TLS_VERSION (0x0304) will be higher than the specified highest_protocol.

This PR fixes the problem by changing the highest_protocol to 0x0304. Note that another solution to this problem would be to change the ClientHello messages for these two browsers. It is my understanding that TLSv1.3 is disabled by default for these browsers, so presumably the ClientHello messages would not specify TLSv1.3 support if they were configured with TLSv1.3 support disabled.


drwetter:
Thx, David!

The data comes from SSLlabs. Of course if not correct we can change this (or we can add something). For the former: we even should.

As far as I understood Ivan Ristic, there might be a change coming up by the end of this year. Whatever that will be, we should watch this.

PR #844, opened 2017-10-06
has_server_protocol() fixes

dcooper16:
This PR fixes the use of `has_server_protocol()` in two places.

Currently `std_ciphersuites()` only tries SSLv2 if the server is known to support SSLv2. This changes it to try SSLv2 unless the server is known to not support SSLv2.

In `run_beast()`, tests against the server are run to determine support for TLSv1.2, TLSv1.1, TLSv1, and SSLv3 unless the server is known to support that protocol (i.e., even if `has_server_protocol()` reports that the server does not support the protocol). This changes it so that a test is only performed against the server if `has_server_protocol()` reports that it doesn't know whether the protocol is supported.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2418,7 +2418,7 @@ std_cipherlists() {
                sclient_success=$?
                [[ $sclient_success -eq 2 ]] && sclient_success=0
           fi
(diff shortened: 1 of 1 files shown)


drwetter:
Thx @dcooper16

Issue #660, opened 2017-03-13
3DES and RC4 *only* on TLSv1.0 leads to false negatives 


fishouttah20:
Refer to twitter DM for domain. When using latest testssl.sh 2.9dev to scan a site that supports TLSv1.2 and TLSv1.0, offers 3DES on both and RC4 only on TLSv1.0, the standard ciphers section says there is no 3DES and the Vulnerabilities section says there is no RC4.

drwetter:
Thx for reporting.

> the standard ciphers section says there is no 3DES 

That I cannot confirm. 

> and the vulnerabilities section says there is no RC4

confirmed. ``testssl.sh -E``:

``` Testing ciphers per protocol via OpenSSL plus sockets against the server, ordered by encryption strength 

Hexcode  Cipher Suite Name (OpenSSL)       KeyExch.   Encryption  Bits     Cipher Suite Name (RFC)
-----------------------------------------------------------------------------------------------------------------------------
SSLv2 
SSLv3 
TLS 1 
 x35     AES256-SHA                        RSA        AES         256      TLS_RSA_WITH_AES_256_CBC_SHA                       
 x2f     AES128-SHA                        RSA        AES         128      TLS_RSA_WITH_AES_128_CBC_SHA                       
 x05     RC4-SHA                           RSA        RC4         128      TLS_RSA_WITH_RC4_128_SHA                           
 x0a     DES-CBC3-SHA                      RSA        3DES        168      TLS_RSA_WITH_3DES_EDE_CBC_SHA                      
TLS 1.1 
TLS 1.2 
 x3d     AES256-SHA256                     RSA        AES         256      TLS_RSA_WITH_AES_256_CBC_SHA256                    
 x35     AES256-SHA                        RSA        AES         256      TLS_RSA_WITH_AES_256_CBC_SHA                       
 x3c     AES128-SHA256                     RSA        AES         128      TLS_RSA_WITH_AES_128_CBC_SHA256                    
 x2f     AES128-SHA                        RSA        AES         128      TLS_RSA_WITH_AES_128_CBC_SHA                       
 x0a     DES-CBC3-SHA                      RSA        3DES        168      TLS_RSA_WITH_3DES_EDE_CBC_SHA        
```

``testssl.sh -e``

```
 Testing 359 ciphers via OpenSSL plus sockets against the server, ordered by encryption strength 

[...]

PR #854, opened 2017-10-10
Add TLSv1.3 support for run_std_cipherlists()

dcooper16:
This PR adds support for TLSv1.3 to run_std_cipherlists().

This PR also provides a partial fix for #660 - addressing the issue only for run_std_cipherlists(). Rather than testing the server once for each cipher list, it tries the server once for each protocol supported by the server. This makes the testing more robust, but adds significantly to the time it takes to perform the tests.


drwetter:
thx

PR #858, opened 2017-10-11
Fix #660 for run_sweet32()

### LINKED ISSUE #660: 3DES and RC4 *only* on TLSv1.0 leads to false negatives

dcooper16:
This PR addresses issue #660 for `run_sweet32()`, detecting if 3DES is used, even if it isn't used with the highest protocol version supported by the server. As with PR #854, this increased robustness comes at the expense of taking additional time to run the test.


drwetter:
Thx

PR #863, opened 2017-10-17
run_server_preference() bug fix

dcooper16:
If `run_server_preference()` is performed:
* against a server that supports SSLv3 and that does not have a cipher order; and
* using a version of OpenSSL that does not support SSLv3; and
* with the `--mapping` option set to "rfc" or "no-openssl"

then the "Negotiated cipher per proto" will not show the SSLv3 cipher since `cipher[i]` will be empty.


drwetter:
thx

PR #864, opened 2017-10-18
Make CAA record lookups resolve the entire DNS tree (Fixes #862)

### LINKED ISSUE #862: CAA record check does not consider whole DNS tree
I originally thought that the issue was something else.
---

Old dig output (e.g. OSx):
```
$ dig atom86.net type257 +short
\# 38 0005696F6465666D61696C746F3A6162757365407363687562657267 7068696C69732E636F6D
\# 19 00056973737565636F6D6F646F63612E636F6D
```

Newer dig (e.g. latest Redhat 6.9)
```
> dig atom86.net type257 +short
0 iodef "mailto:abuse@schubergphilis.com"
0 issue "comodoca.com"
```
The latter appears not to be recognized by testssl.sh

seccubus:
CAA record lookups should be performed on the entire DNS tree as per rfc https://tools.ietf.org/html/rfc6844#page-7

```
  For example, if a certificate is requested for X.Y.Z the issuer will
   search for the relevant CAA record set in the following order:

      X.Y.Z

      Alias (X.Y.Z)

      Y.Z

      Alias (Y.Z)

      Z

      Alias (Z)

      Return Empty
```

This is implemented with a simple loop

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -5984,6 +5984,7 @@ certificate_info() {
      local days2warn2=$DAYS2WARN2
      local days2warn1=$DAYS2WARN1
      local provides_stapling=false
+     local caa_node=""
 
      if [[ $number_of_certificates -gt 1 ]]; then
           [[ $certificate_number -eq 1 ]] && outln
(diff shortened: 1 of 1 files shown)


drwetter:
Yes, thx

PR #869, opened 2017-10-19
Please sort caa records to avoid flapping findings in tools like Seccubus

seccubus:
;)


drwetter:
Wrong spot. Tonight or tommorow I take care...

⁣-- 

PR #879, opened 2017-10-27
Use read_sigalg_from_file() helper function

dcooper16:
This PR fixes several places where the `read_sigalg_from_file()` helper function isn't being used.


drwetter:
cool!

PR #881, opened 2017-10-27
Fix problem with LibreSSL encryption names

dcooper16:
For ciphers that use the ChaCha20-Poly1305 cipher, LibreSSL shows "Enc=ChaCha20-Poly1305" in the `openssl ciphers -V` command rather than "Enc=ChaCha20(256)" and for some GOST ciphers it shows "Enc=GOST-28178-89-CNT" rather than "Enc=GOST(256)". This causes a problem for `neat_list()` if information is being obtained from `$OPENSSL ciphers -V` rather than from the cipher-mapping.txt file.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2686,6 +2686,11 @@ neat_list(){
 
      kx="${3//Kx=/}"
      enc="${4//Enc=/}"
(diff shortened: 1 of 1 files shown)


drwetter:
thx

PR #882, opened 2017-10-27
Suppress more config file warnings

dcooper16:
This PR is a continuation of #833.

With additional testing with different options I encountered more places where `$OPENSSL` was printing "WARNING: can't open config file: /usr/local/etc/ssl/openssl.cnf" where testssl.sh was not suppressing the error message.

This PR redirects `stderr` to `/dev/null` or to `$ERRFILE` for several more calls to `$OPENSSL` in order to suppress these warning messages.


drwetter:
good

PR #883, opened 2017-10-27
Add TLSv1.3 support for run_cipher_match()

dcooper16:
This PR adds TLSv1.3 support for `run_cipher_match()`. It also addresses issue #660 for `run_cipher_match()`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2757,6 +2757,7 @@ run_cipher_match(){
      local -i nr_ciphers=0 nr_ossl_ciphers=0 nr_nonossl_ciphers=0
      local -i num_bundles mod_check bundle_size bundle end_of_bundle
      local dhlen has_dh_bits="$HAS_DH_BITS"
+     local cipher proto protos_to_try
      local available
      local -i sclient_success
      local re='^[0-9A-Fa-f]+$'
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -2909,37 +2910,51 @@ run_cipher_match(){
...
-                         kx[i]="${kx[i]} $dhlen"
-                    fi
-                    "$SHOW_SIGALGO" && grep -q "\-\-\-\-\-BEGIN CERTIFICATE\-\-\-\-\-" $TMPFILE && \
-                         sigalg[i]="$(read_sigalg_from_file "$TMPFILE")"
                done
           done
 
           if "$using_sockets"; then
                for (( i=0; i < nr_ciphers; i++ )); do
-                    if ! "${ciphers_found[i]}" && [[ "${sslvers[i]}" != "SSLv2" ]] && [[ "${hexcode[i]}" != "13"* ]]; then
```
drwetter: Quick q: why was the ``${hexcode[i]}" != "13"* `` in before?

drwetter:
Thank you!

testssl.sh
dcooper16: To exclude any TLSv1.3 ciphers, since previously TLSv1.3 ciphers weren't being tested. I think the reason this was included here and not in other functions (e.g., `run_all_ciphers()`) is that at the time this line of code was added (back when `run_cipher_match()` was called `test_just_one()`) the output was always showing each tested cipher, not just those that were available, even if the `--show-each` flag was not set.

PR #895, opened 2017-11-07
Use tls_sockets() for run_heartbleed()

dcooper16:
This PR changes `run_heartbleed()` to use `tls_sockets()` to send the ClientHello and to read the ServerHello.


dcooper16:
I simplified the PR a bit. Previously I had `run_heartbleed()` pass a cipher list to `tls_sockets()`, either `$TLS12_CIPHER` or `$TLS_CIPHER` depending on whether a TLSv1.2 ClientHello or an earlier version was being sent. However, this was unnecessary since sending a empty cipher list has the same result.

drwetter:
Cool, good improvements. Thx!

PR #897, opened 2017-11-15
Fix socksend() debugging output

dcooper16:
For debug levels 4 and above, `socksend()` is supposed to print the data that is being sent to the terminal. It currently uses `echo -e` for this so that a newline character can be written before the data. However, the `-e` option causes every byte of `$data`, written as "\xHH" to be interpreted as an escape sequence rather than just being printed as plain text.

This PR fixes the problem by removing the `-e` option and using a separate call to `echo` to create the preceding newline character.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2604,7 +2604,7 @@ socksend() {
      else
           data=$(sed -e 's/# .*$//g' -e 's/ //g' <<< "$1" | sed -r 's/^[[:space:]]+//; s/[[:space:]]+$//; /^$/d' | sed 's/,/\\/g' | tr -d '\n')
      fi
-     [[ $DEBUG -ge 4 ]] && echo -e "\n\"$data\""
+     [[ $DEBUG -ge 4 ]] && echo "" && echo "\"$data\""
      printf -- "$data" >&5 2>/dev/null &
      sleep $2
 }


drwetter:
This should be better. ``safe_echo()`` was meant for places where user determinable output is processed. It is not the case here probably but it does the job as well

drwetter:
Thx for putting this on the table though

dcooper16:
Hi Dirk,

I tried your commit from yesterday and it is not working for me. Below is an example of what I am seeing:
```
> declare data="\x31\x32\x33\x34\x35\x36\x37"
> echo "\"$data\""
"\x31\x32\x33\x34\x35\x36\x37"
> safe_echo()  { printf -- "%b" "${1//%/%%}"; }
> safe_echo "\"$data\""
"1234567">
```
Using `echo` without the `-e` option displays the contents of `$data`, but `safe_echo()` prints out the binary data that `$data` represents. In the example above I choose the hex value for ASCII characters, but when called from `socksend()` what is printed just looks like garbage.

Also note that unlike the code it replaced, the current code does not insert a newline character either before or after the text that is printed.

How about this as an alternative to what was committed yesterday:
```
     [[ $DEBUG -ge 4 ]] && safe_echo "\n\"${data//\\/\\\\}\"\n"
```
That command seems to work for me:
```
> declare data="\x31\x32\x33\x34\x35\x36\x37"
> safe_echo()  { printf -- "%b" "${1//%/%%}"; }
> safe_echo "\n\"${data//\\/\\\\}\"\n"

"\x31\x32\x33\x34\x35\x36\x37"
>
```

PR #898, opened 2017-11-16
Fix socksend() debugging output

dcooper16:
In case you are okay with my idea of escaping the `\` characters before calling `safe_echo()` to address #897, this PR implements that approach. It also changes three other places that print similarly formatted data to use `safe_echo()` rather than `echo`.


drwetter:
Thanks for the PR and your correction again.

The data comes from the code and thus we don't need to use ``safe_echo()``. Honestly if we need backslashes here (readability) I would rather prefer to pick up your previous suggestion.

Issue #333, opened 2016-03-30
Test for conformance to NIST SP 800-52


dcooper16:
The reason that I started contributing to testssl.sh is that I was tasked with developing a test suite that tests a web server's configuration to the[ National Institute of Standards and Technology](http://csrc.nist.gov)'s ([NIST](http://csrc.nist.gov)) [Special Publication 800-52 Revision 1, _Guidelines for the Selection, Configuration, and Use of Transport Layer Security (TLS) Implementations_](http://dx.doi.org/10.6028/NIST.SP.800-52r1). I'm using testssl.sh since it already does most of what is needed, and the code that is there will make the implementation of the rest much easier.

If necessary, the version of the code that tests for conformance to SP 800-52 will be maintained as a separate fork, but if there's interest in incorporating the additional tests into the baseline code that would be my preference. My thought would be to add a "--SP800-52" flag and then only test for things that are specific to that document if the flag is set.

Would this be of interest, or should I not bother to submit PRs for code that would test conformance to SP 800-52, but that wouldn't be of interest to anyone who isn't trying to test conformance to SP 800-52?

Thanks,

David

drwetter:
Am 30. März 2016 17:10:49 MESZ, schrieb David Cooper <notifications@github.com>:
>The reason that I started contributing to testssl.sh is that I was
>tasked with developing a test suite that tests a web server's
>configuration to the[ National Institute of Standards and
>Technology](http://csrc.nist.gov)'s ([NIST](http://csrc.nist.gov))
>[Special Publication 800-52 Revision 1, _Guidelines for the Selection,
>Configuration, and Use of Transport Layer Security (TLS)
>Implementations_](http://dx.doi.org/10.6028/NIST.SP.800-52r1). I'm
>using testssl.sh since it already does most of what is needed, and the
>code that is there will make the implementation of the rest much
>easier.

Good to hear!

>If necessary, the version of the code that tests for conformance to SP
>800-52 will be maintained as a separate fork, but if there's interest
>in incorporating the additional tests into the baseline code that would
>be my preference. My thought would be to add a "--SP800-52" flag and
>then only test for things that are specific to that document if the
>flag is set.

I like both ideas, a) having one common code b) a NIST check.

Besides it's kind of an honor for the project.

>Would this be of interest, or should I not bother to submit PRs for
>code that would test conformance to SP 800-52, but that wouldn't be of
>interest to anyone who isn't trying to test conformance to SP 800-52?

[...]

PR #905, opened 2017-11-21
Color all ciphers

dcooper16:
This PR adds an additional COLOR level (3). If color is set to 3 then all ciphers are printed according to `pr_cipher_quality()` rather than just the "Negotiated cipher" in `run_server_preference()`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -168,7 +168,7 @@ TERM_CURRPOS=0                                              # custom line wrappi
 # 0 means (normally) true here. Some of the variables are also accessible with a command line switch, see --help
 declare -x OPENSSL OPENSSL_TIMEOUT
 FAST_SOCKET=${FAST_SOCKET:-false}       # EXPERIMENTAL feature to accelerate sockets -- DO NOT USE it for production
-COLOR=${COLOR:-2}                       # 2: Full color, 1: b/w+positioning, 0: no ESC at all
+COLOR=${COLOR:-2}                       # 3: extra color, 2: Full color, 1: b/w+positioning, 0: no ESC at all
 COLORBLIND=${COLORBLIND:-false}         # if true, swap blue and green in the output
 SHOW_EACH_C=${SHOW_EACH_C:-false}       # where individual ciphers are tested show just the positively ones tested
 SHOW_SIGALGO=${SHOW_SIGALGO:-false}     # "secret" switch whether testssl.sh shows the signature algorithm for -E / -e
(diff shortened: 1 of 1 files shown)


drwetter:
Yeah, now I remember we discussed that a while back. And IIRC I personally wasn't so in favor of having this as a default. :) I realize though that it is a matter of taste. ``--color=3`` seems absolutely fine as less obtrusive and it doesn't collide with a white background. Thanks for the PR!

The thing though I would like to have explained is the color rating, see picture.
screenshot_20171201_214220

``anon`` seems like an accident to me. That needs to be in light red. AES256-SHA256/TLS_RSA_WITH_AES_256_CBC_SHA256 is the the very best but I wouldn't pick the color for medium, same with CAMELLIA. Yellow seems better to me.

dcooper16:
Hi Dirk,

I've been maintaining the code to color all the ciphers for a while now in the branch that I've been working on for #333, and decided to create a PR with that code as a result of #891. I just changed it to only color the ciphers if `--color=3` based on our earlier discussion.

This PR doesn't change the color ratings at all. It just uses `pr_cipher_quality()` for the cipher ratings, and `pr_cipher_quality()` was developed using the cipher ratings that were previously embedded in `run_server_preference()` (PR #637). So, it seems that changing the color ratings should be a separate PR.

TLS_RSA_WITH_AES_256_CBC_SHA256 isn't rated as the best since it is a CBC cipher. The comment in `pr_cipher_quality()` implies that this is because of BEAST. However, if TLSv1.2 isn't vulnerable to BEAST, then perhaps TLSv1.2-only CBC ciphers should be upgraded.

The color ratings don't really matter to me, though. In the branch for #333, `pr_cipher_quality()` uses a completely different set of rules for rating ciphers if the `--SP800-52` flag is used:

- If the cipher is permitted by SP 800-52 Revision 2, it is rated a 5, 6, or 7 (ok, good, best).
- If the cipher is not permitted by SP 800-52 Revision 2, but was permitted by SP 800-52 Revision 1, it is rated a 3 (medium severity), unless the cipher uses 3DES. (3DES is rated worse since the Department of Homeland Security issued a Binding Operational Directive mandating that U.S. Federal agencies disable 3DES: https://cyber.dhs.gov/assets/report/bod-18-01.pdf).
- If the cipher is not permitted by either SP 800-52 Revision 2 or Revision 1, then it is rated a 1 (critical).

drwetter:
Thank you David.

``pr_cipher_quality()`` was kind of a forgotten function, at least for my weak memory. :-)

As stated in the commit there's some work ahead to get the color rating consistent throughout testssl.sh. That is some work which needs to be done for a release. I need to look at your branch.

For now I guess I am fine to leave ``--color=3`` as a hidden tweak (i.e. not appearing in ``help()`` or in the man page). 

Thx again, Dirk

PR #912, opened 2017-11-29
Support TLSv1.3 draft 22

dcooper16:
This PR adds support for TLSv1.3 draft 22. This PR has testssl.sh operate in "middlebox compatibility mode" as described in Appendix D.4 of draft-ietf-tls-tls13-22 to maximize the chances of being able to perform a successful test even if there is a misbehaving middlebox between testssl.sh and the server being tested. Support for drafts 18 through 21 is still maintained.

This PR has been tested against a few different implementations of draft 22 that were made available shortly before draft 22 was posted.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4587,6 +4587,11 @@ run_protocols() {
                          [[ -n "$drafts_offered" ]] && drafts_offered+=", "
                          drafts_offered+="draft 21"
                     fi
+                    tls_sockets "04" "$TLS13_CIPHER" "" "00, 2b, 00, 03, 02, 7f, 16"
+                    if [[ $? -eq 0 ]]; then
+                         [[ -n "$drafts_offered" ]] && drafts_offered+=", "
+                         drafts_offered+="draft 22"
+                    fi
                     tls_sockets "04" "$TLS13_CIPHER" "" "00, 2b, 00, 03, 02, 03, 04"
                     if [[ $? -eq 0 ]]; then
                          [[ -n "$drafts_offered" ]] && drafts_offered+=", "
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -8445,8 +8451,8 @@ check_tls_serverhellodone() {
           [[ $remaining -lt 10 ]] && return 1
 
           tls_content_type="${tls_hello_ascii:i:2}"
-          [[ "$tls_content_type" != "15" ]] && [[ "$tls_content_type" != "16" ]] && \
-               [[ "$tls_content_type" != "17" ]] && return 2
+          [[ "$tls_content_type" != "14" ]] && [[ "$tls_content_type" != "15" ]] && \
+               [[ "$tls_content_type" != "16" ]] && [[ "$tls_content_type" != "17" ]] && return 2
```
drwetter: a case statement (which matches 14 to 17 and do nothing, for all other a ``return 2``) would be more readable

drwetter:
I know  of one only: https://tls13.crypto.mozilla.org/  (which behaves using testssl.sh as you said a while back strangely for all other protocols).

Any more publicly available guineapigs?

drwetter:
Sorry, I trusted the docu ;-) @ https://github.com/tlswg/tls13-spec/wiki/implementations

```
enabled.tls13.com_p443:         draft     18"
h2o.examp1e.net_p443:           "TLSv1.3  is   not     offered  and  downgraded  to  a  weaker  protocol"
rustls.jbp.io_p443:             draft     18"
tls13.baishancloud.com_p443:    draft     18,  final"
tls13.baishancloud.com_p44344:  draft     21,  final"
tls13.crypto.mozilla.org_p443:  draft     22"
tls.ctf.network_p443:           draft     18,  draft   22"
www.henrock.net_p443:           draft     18,  final"
www.mew.org_p13443:             draft     22,  final"
www.mew.org_p443:               draft     18,  final"
```

drwetter:
good work!

dcooper16:
The only publicly available server I tested it against was https://tls13.crypto.mozilla.org, which I assume is using NSS. However, I also tested against https://github.com/openssl/openssl/pull/4701 and against https://github.com/h2o/picotls.

drwetter:
> However, I also tested against openssl/openssl#4701 

ah, just pulled compiled openssl without that patch :-(

PR #913, opened 2017-11-29
Reducing processing of alert messages

dcooper16:
This PR reduces the amount of work `parse_tls_serverhello()` does when processing alert messages when not in debug mode. It delays writing anything to `$TMPFILE` unless `$DEGUG -ge 1` until it has reason to believe that the response was successful. If `$DEBUG` is 0 and alert messages are sent, then no file operations are performed processing the alert messages.

In almost every case, there is no attempt to look at the contents of `$TEMPDIR/$NODEIP.parse_tls_serverhello.txt` unless the connection was successful. So, in most cases, it is okay to not call `tmpfile_handle()` in `parse_tls_serverhello()` unless the connection was successful. There is, however, one place in `run_grease()` where the code reads the contents of `$TEMPDIR/$NODEIP.parse_tls_serverhello.txt` even if the connection was not successful. In order to address this, the DEBUG level is temporarily set to 1 when performing this test if its value is 0. Also in order to address this, changes were made in `parse_tls_serverhello()` to ensure that `tmpfile_handle $FUNCNAME.txt` is always called before returning if `$DEBUG -ge 1`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -8629,7 +8629,7 @@ parse_tls_serverhello() {
      local -i dh_p_len dh_param_len
 
      DETECTED_TLS_VERSION=""
-     [[ -n "$tls_hello_ascii" ]] && echo "CONNECTED(00000003)" > $TMPFILE
+     [[ $DEBUG -ge 1 ]] && echo > $TMPFILE
 
      [[ "$DEBUG" -ge 5 ]] && echo $tls_hello_ascii      # one line without any blanks
 
@@ -8649,6 +8649,7 @@ parse_tls_serverhello() {
                if [[ "$process_full" == "all" ]]; then
                     # The entire server response should have been retrieved.
                     debugme tmln_warning "Malformed message."
+                    [[ $DEBUG -ge 1 ]] && tmpfile_handle $FUNCNAME.txt
                     return 1
                else
                     # This could just be a result of the server's response being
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -12436,6 +12471,12 @@ run_grease() {
 
      # Send a list of non-existent ciphers where the second byte does not match
      # any existing cipher.
+
+     # Need to ensure that $TEMPDIR/$NODEIP.parse_tls_serverhello.txt contains the results of the
+     # most recent calls to tls_sockets even if tls_sockets is not successful. Setting $DEBUG to
+     # a non-zero value ensures this. Setting it to 1 prevents any extra information from being
+     # displayed.
+     [[ $DEBUG -eq 0 ]] && DEBUG=1
```
drwetter: That seems not correct to me. DEBUG is a global

dcooper16: This is just a temporary change to the value of $DEBUG. It gets set back to its original value about 20 lines later.

As noted above and in the comments in the code, there is one set of tests in run_grease() that need to be able to read `$TEMPDIR/$NODEIP.parse_tls_serverhello.txt` even if the connection was unsuccessful, and with this PR the only way to ensure that `$TEMPDIR/$NODEIP.parse_tls_serverhello.txt` contains information from the most recent call to `parse_tls_serverhello()` is to ensure `$DEBUG -ge 1`.

drwetter:
Thanks! I shouldn't work so late anymore ;/

PR #917, opened 2017-11-30
Compare selected version against supported_versions

dcooper16:
If a supported_versions extension was included in the ClientHello, then check that the version returned by the server was included in the ClientHello's supported_versions extension.

OpenSSL will respond to a TLSv1.3 ClientHello that only specifies 0304 in its supported_versions extension with a ServerHello that specifies whatever draft of TLSv1.3 it currently supports (e.g., 7F16). The result is that `run_protocols()` incorrectly reports that OpenSSL supports TLSv1.3 "final" in addition to whatever draft version it supports.

This PR fixes that problem by treating it as a failed connection when the ClientHello offers only 0304 and the ServerHello specifies something else (e.g., 7F16).

Performing this check is actually a requirement for clients in Section 4.2.1 of draft-ietf-tls-tls13-22. So, including this check will also help make client simulations more accurate when clients that support TLSv1.3 are added to client-simulation.txt.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -9992,14 +9992,25 @@ socksend_tls_clienthello() {
      printf -- "$data" >&5 2>/dev/null &
      sleep $USLEEP_SND
 
-     if [[ "$tls_low_byte" -gt 0x03 ]]; then
-          TLS_CLIENT_HELLO="$(echo -n "$NW_STR" | tr 'A-F' 'a-f' | \
-               sed -e 's/\\x0\\/\\x00\\/g' -e 's/\\x1\\/\\x01\\/g' \
-               -e 's/\\x2\\/\\x02\\/g' -e 's/\\x3\\/\\x03\\/g' -e 's/\\x4\\/\\x04\\/g' \
-               -e 's/\\x5\\/\\x05\\/g' -e 's/\\x6\\/\\x06\\/g' -e 's/\\x7\\/\\x07\\/g' \
-               -e 's/\\x8\\/\\x08\\/g' -e 's/\\x9\\/\\x09\\/g' -e 's/\\xa\\/\\x0a\\/g' \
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -9961,6 +9992,17 @@ socksend_tls_clienthello() {
...
 
+     if [[ "$tls_low_byte" -gt 0x03 ]]; then
+          TLS_CLIENT_HELLO="$(echo -n "$NW_STR" | tr 'A-F' 'a-f' | \
+               sed -e 's/\\x0\\/\\x00\\/g' -e 's/\\x1\\/\\x01\\/g' \
+               -e 's/\\x2\\/\\x02\\/g' -e 's/\\x3\\/\\x03\\/g' -e 's/\\x4\\/\\x04\\/g' \
+               -e 's/\\x5\\/\\x05\\/g' -e 's/\\x6\\/\\x06\\/g' -e 's/\\x7\\/\\x07\\/g' \
+               -e 's/\\x8\\/\\x08\\/g' -e 's/\\x9\\/\\x09\\/g' -e 's/\\xa\\/\\x0a\\/g' \
+               -e 's/\\xb\\/\\x0b\\/g' -e 's/\\xc\\/\\x0c\\/g' -e 's/\\xd\\/\\x0d\\/g' \
+               -e 's/\\xe\\/\\x0e\\/g' -e 's/\\xf\\/\\x0f\\/g' -e 's/\\x//g')"
+          TLS_CLIENT_HELLO="${TLS_CLIENT_HELLO:10}"
```
drwetter: can't this be better done in bash?

dcooper16: Yes. Sorry about that. I noticed that when I was submitting the PR. I've changed it now to use internal Bash string manipulation functions.

drwetter:
thank you!

PR #920, opened 2017-12-01
HTTP-related checks and certificate based client authentication

dcooper16:
If certificate-based client authentication is required by the server, then most HTTP-related checks are skipped, even if the "--assume-http" flag is used. If `$CLIENT_AUTH` is true, then `$ASSUME_HTTP` is ignored.

In some cases the checks are appropriately skipped, since the tests cannot be performed. In other places, the value of `$CLIENT_AUTH` is used as a hint as to whether HTTP is being used. For example, in `run_tickbleed()`:
```
     if [[ "$SERVICE" != HTTP ]] && ! "$CLIENT_AUTH"; then
          outln "--   (applicable only for HTTPS)"
          fileout "ticketbleed" "INFO" "Ticketbleed: not applicable, not HTTP" "$cve" "$cwe"
          return 0
     fi
```
There are some places, however, where tests are just skipped if both `$CLIENT_AUTH` and `$ASSUME_HTTP` are true, even though the test could be performed. For example, `run_client_simulation()` only simulates generic clients in this case.

This PR attempts to address this:
* In `run_client_simulation()` it runs all of the tests if `$ASSUME_HTTP` is true.
* In `certificate_transparency()` it only says that the lack of CT information is "N/A" it can verify that HTTP is not being used (if `$SERVICE` is not HTTP and `$CLIENT_AUTH` is false). Otherwise it just says "no" without flagging it as an issue.
* In `certificate_info()` it displays additional warnings (about use of SHA-1 or subjectAltName matching) only if it can verify that HTTP is being used (`$SERVICE` is HTTP or `$ASSUME_HTTP` is true).
* In `run_crime()`, if compression is used, it only says " but not using HTTP" only if it can verify that HTTP is not being used (if `$SERVICE` is not HTTP and `$CLIENT_AUTH` is false).

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3945,14 +3945,14 @@ run_client_simulation() {
 
      if [[ $SERVICE != "" ]];  then
           client_service="$SERVICE"
-     else
+     elif [[ -n "$STARTTLS_PROTOCOL" ]]; then
           # Can we take the service from STARTTLS?
-           if [[ -n "$STARTTLS_PROTOCOL" ]]; then
-               client_service=$(toupper "${STARTTLS_PROTOCOL%s}")    # strip trailing 's' in ftp(s), smtp(s), pop3(s), etc
-          else
-               outln "Could not determine the protocol, only simulating generic clients."
-               client_service="undetermined"
-          fi
+          client_service=$(toupper "${STARTTLS_PROTOCOL%s}")    # strip trailing 's' in ftp(s), smtp(s), pop3(s), etc
+     elif "$ASSUME_HTTP"; then
+          client_service="HTTP"
+     else
+          outln "Could not determine the protocol, only simulating generic clients."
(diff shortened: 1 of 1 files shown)


drwetter:
thank you, David.

We're a bit behind WRT unit tests. It would be great to have one also for a server requiring X509/client based auth.

Any idea here?

Cheers, Dirk

dcooper16:
Hi Dirk,

No, I don't have any thoughts. Are you thinking about a test of whether a server requires certificate based client authentication beyond what is already done?

drwetter:
Hi David,

yes. But honestly I think I shot to fast here. :-) The delta / difference to other servers is too meager, so probably not worth a separate test. I also do not expect to catch much which wouldn't be caught otherwise.

Cheers, Dirk

PR #932, opened 2017-12-12
Reduce use of temporary files

dcooper16:
This PR reduces testssl.sh's use of temporary files by making greater use of pipes. This results in a very slight performance improvement.


drwetter:
Looks good, thank you.

PR #936, opened 2017-12-13
Introduce property global HAS_PEY

dcooper16:
Similar to the recently added HAS_PKUTIL (https://github.com/drwetter/testssl.sh/commit/f829878a43098ac3bb146f4905c2fa78d7b90e9e), this PR adds HAS_PKEY, which indicates whether OpenSSL has the `pkey` utility. HAS_PKEY is then checked before attempting to do anything that requires the `pkey` utility.


drwetter:
Thank you

PR #938, opened 2017-12-14
Derive handshake traffic key for TLSv1.3

dcooper16:
This is the first in a series of PRs to add support for processing the encrypted portions of the server's response in a TLSv1.3 handshake.

This PR adds the code to derive the handshake traffic key needed to decrypt the response (the next PR will add the code to perform the symmetric-key decryption of the encrypted portions of the response).

Since this PR does not make use of the traffic key that it derives, it doesn't yet add any new functionality.

Note that testssl.sh will not always be able to derive the session keys. If the version of OpenSSL that is bundled with testssl.sh is used and the server chooses to use an X25519 ephemeral key, OpenSSL will be unable to perform the shared secret in `derive-handshake-traffic-secret()`. (OpenSSL 1.1.0 supports X25519.) Since X25519 use a different encoding than ECDH keys, the lack of X25519 support will be discovered in `parse_tls_serverhello()` when `$OPENSSL pkey` is unable to convert the key from DER to PEM. So, in debugging mode, `parse_tls_serverhello()` now displays a warning if it receives a key share that `$OPENSSL pkey` cannot handle.

testssl.sh tries to avoid the above problem by discouraging the server from using X25519 if testssl.sh needs to decrypt the response and OpenSSL 1.1.x isn't being used. It does this by listing X25519 as its least preferred group in the supported_groups extension and by not including an X25519 key share in the key_share extension.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -8491,6 +8491,257 @@ parse_sslv2_serverhello() {
      return $ret
 }
 
+# arg1: hash function
+# arg2: key
+# arg3: text 
+hmac() {
+     local hash_fn="$1"
+     local key="$2" text="$3" output
+     local -i ret
+
+     output="$(asciihex_to_binary_file "$text" "/dev/stdout" | $OPENSSL dgst "$hash_fn" -mac HMAC -macopt hexkey:"$key" 2>/dev/null)"
+     ret=$?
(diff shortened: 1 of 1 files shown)


drwetter:
good!

PR #940, opened 2017-12-15
Decrypt server's TLSv1.3 response

dcooper16:
This PR adds code to decrypt the encrypted portion of the server's response for TLSv1.3 and to then process any certificates and encrypted extensions. This code supports all 5 TLSv1.3 cipher suites, and so any response can be decrypted as long as the session key can be derived (which requires OpenSSL to support the ephemeral key that was used - see #938).

For the symmetric decryption, the `sym-decrypt()` function uses OpenSSL when possible and internal Bash functions when needed.

For AES-GCM and AES-CCM ciphers `sym-decrypt()` normally uses internal Bash functions, which rely on using `$OPENSSL enc` in AES-ECB mode to generate the key stream and then Bash functionality to XOR the key stream with the ciphertext. With some versions of OpenSSL the AES-GCM ciphers are decrypted using `$OPENSSL enc` in AES-GCM mode directly. On my system, however, both methods seem to work about equally fast.

For ChaCha20 ciphers `$OPENSSL enc -chacha20` is used if supported (OpenSSL 1.1.x only), and Bash internal functions (without any OpenSSL support) are used otherwise. In this case, if the Bash internal functions need to be used, decryption is _very, very, very_ slow. Fortunately, in a typical run of testssl.sh there won't be many cases in which the connection will be TLSv1.3 with ChaCha20 and the entire response needs to be processed (requiring decryption). In most cases, even if the connection is TLSv1.3 with ChaCha20, testssl.sh will at most need the ephemeral key, which is available in plain text.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -286,6 +286,9 @@ HAS_PROXY=false
 HAS_XMPP=false
 HAS_POSTGRES=false
 HAS_MYSQL=false
+HAS_CHACHA20=false
+HAS_AES128_GCM=false
+HAS_AES256_GCM=false
 PORT=443                                # unless otherwise auto-determined, see below
 NODE=""
 NODEIP=""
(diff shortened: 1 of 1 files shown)


drwetter:
wow.

drwetter:
@PeterMosmans: this reminds me on a OpenSSL 1.1.0 support with backports of all the dirty stuff. Did you get anywhere or is the time before Xmas as busy as on the northern hemisphere ;-)

PeterMosmans:
Hi @drwetter - I haven't prioritized this (enough) yet, so I'm afraid you still need to hang on some more...

PR #944, opened 2017-12-20
run_renego() and OpenSSL 1.1.1

dcooper16:
`run_renego()` appears to produce a false positive if OpenSSL 1.1.1 is used and the server being tested supports TLSv1.3 (i.e., the server supports the same draft version of TLSv1.3 as the version of OpenSSL 1.1.1 being used does). This PR fixes the problem by telling calls to `$OPENSSL s_client` in `run_renego()` to not use TLSv1.3.

When testing a server that supports TLSv1.3 using OpenSSL 1.1.1 the server is reported as being VULNERABLE to both "Secure Renegotiation" and "Secure Client-Initiated Renegotiation." It is my guess, however, that these vulnerabilities do not apply to TLSv1.3 connections, and so testing should only be performed using TLSv1.2 or below connections.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -11815,18 +11815,20 @@ run_ticketbleed() {
 
 run_renego() {
 # no SNI here. Not needed as there won't be two different SSL stacks for one IP
-     local legacycmd=""
+     local legacycmd="" proto="$OPTIMAL_PROTO"
      local insecure_renogo_str="Secure Renegotiation IS NOT"
      local sec_renego sec_client_renego
      local cve="CVE-2009-3555"
      local cwe="CWE-310"
      local hint=""
 
+     "$HAS_TLS13" && [[ -z "$proto" ]] && proto="-no_tls1_3"
+
      [[ $VULN_COUNT -le $VULN_THRESHLD ]] && outln && pr_headlineln " Testing for Renegotiation vulnerabilities " && outln
 
      pr_bold " Secure Renegotiation "; out "($cve)      "    # and RFC 5746, OSVDB 59968-59974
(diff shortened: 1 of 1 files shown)


drwetter:
Good point, yes. Thank you.

I haven't check myself but openssl support of TLS 1.3 implies also the (negating) switch ``-no_tls1_3`` / it is safe to use that, right?

dcooper16:
> I haven't check myself but openssl support of TLS 1.3 implies also the (negating) switch ` -no_tls1_3` / it is safe to use that, right?

It seems to be safe. It seems that OpenSSL always has a flag to enable or disable each protocol version it supports.

We could probably do the same with the `-no_ssl2` switch (use it only if `$HAS_SSL2` is true). However, the `-no_ssl2` switch continued to be supported even after support for SSLv2 was dropped, so it is safe to include it even in some cases in which $OPENSSL doesn't support SSLv2.

drwetter:
thanks. I would suggest we leave it like it is. I think the window between appearing TLS 1.3 in openssl 1.1 and the ``-no_tls1_3 `` is which matters.

dcooper16:
It just occurred to me that there is a possible (but extremely unlikely) problem with the fix that I created for this issue:
```
proto="$OPTIMAL_PROTO"
...
"$HAS_TLS13" && [[ -z "$proto" ]] && proto="-no_tls1_3"
```
If `$OPTIMAL_PROTO` is `-tls1_3`, then the tests will be run using a TLSv1.3 connection, which will presumably result in a false positive.

Of course, in order for this to happen, `$OPENSSL s_client` in `determine_optimal_proto()` would have to fail if no protocol were specified (and if the `-tls1_2` or `-tls1` flag was used), but then succeed if the `-tls1_3` flag was used.

PR #950, opened 2017-12-21
run_tls_fallback_scsv() improvements

dcooper16:
This PR is intended to improve the functionality of `run_tls_fallback_scsv()`.

The original goal of this PR was to address servers that support TLSv1.3 when using OpenSSL 1.1.1. That does not seem to be an issue, as using `$OPENSSL s_client` with the `-no_tls1_2` flag results in a TLSv1.1 ClientHello, even if `$OPENSSL` supports TLSv1.3. However, if the server supports TLSv1.3, then a message that says "No fallback possible, TLS 1.2 is the only protocol (OK)" isn't entirely correct.

The main issue this PR fixes is some false positives in servers that do not support TLSv1.2. On a few servers that I tested, the current code incorrectly reports "Downgrade attack prevention NOT supported." Some of the servers only support TLSv1, so it should report that fallback is not possible. Another server supports TLSv1.1 and TLSv1, and it supports fallback protection, but the current code says that it doesn't. In both cases, the current code produces a false positive since it assumes that TLSv1.2 is supported.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -12281,6 +12281,7 @@ run_tls_poodle() {
 
 run_tls_fallback_scsv() {
      local -i ret=0
+     local p high_proto="" high_proto_str low_proto="" protos_to_try
 
      [[ $VULN_COUNT -le $VULN_THRESHLD ]] && outln && pr_headlineln " Testing for TLS_FALLBACK_SCSV Protection " && outln
      pr_bold " TLS_FALLBACK_SCSV"; out " (RFC 7507)              "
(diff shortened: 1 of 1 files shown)


drwetter:
Thank you, David.

For those of things I wish (sigh) we would have a good unit test. If it weren't you submitting this PR I would have headaches with those changes without a proof (other then the ones I run, with limited value)

But admittedly those things don't change often and for other scenarios unit tests would have a higher priority.

PR #951, opened 2017-12-22
Define extract_certificates() function

dcooper16:
This PR defines an `extract_certificates()` function in order to remove some redundant code from `get_server_certificate()`. Currently, nearly identical code appears in two places in `get_server_certificate()` to extract certificates from the output of `$OPENSSL sclient`, in one place for SSLv2 responses and in another for SSLv3 through TLSv1.2. The code to get the certificates used with TLSv1.3 (see https://github.com/dcooper16/testssl.sh/tree/extended_tls_sockets) would have added a third place where this same code would be needed. This PR allows the code to be written once and used in all three places.


drwetter:
Thank you!

PR #958, opened 2018-01-12
Add support for TLSv1.3 draft 23

dcooper16:
This PR adds support for draft 23, which contains 2 changes that are relevant for testssl.sh. It adds a few new values for the signature_algorithms extension and it changes to extension number for the key_share extension from 40 to 51.

With the change in the extension number, it is no longer possible to send a single ClientHello that works for all supported drafts of TLSv1.3. (I tried sending a ClientHello with two key share extensions, 40 and 51, but that didn't work.) So, this commit adds a test to `determine_optimal_proto()` to determine whether TLSv1.3 is supported and if so whether draft 23 is supported or only some earlier draft (18-22). In subsequent tests, the ClientHello uses the appropriate number for the key share (40 or 51) and specifies the appropriate draft version(s) in the supported_versions extension (either 23 or 18-22). In the case of `run_protocols()` the test for each draft version uses the appropriate key share extension number so that servers that support both draft 23 and an earlier draft can be detected.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -255,6 +255,9 @@ HEADERFILE=""
 HEADERVALUE=""
 HTTP_STATUS_CODE=""
 PROTOS_OFFERED=""                       # this is a global to keep the info which protocol is being offered. See has_server_protocol()
+KEY_SHARE_EXTN_NR="33"                  # The extension number for key_share was changed from 40 to 51 in TLSv1.3 draft 23. In order to
+                                        # support draft 23 in additional to earlier drafts, need to know which extension number to use.
+                                        # Note that it appears that a single ClientHello cannot advertise both draft 23 and earlier drafts.
 TLS_EXTENSIONS=""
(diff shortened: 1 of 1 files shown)


drwetter:
cool, thx!

drwetter:
One question, David: I didn't quite get why you preset KEY_SHARE_EXTN_NR as it needs to be determined anyway?

dcooper16:
I guess in the end there wasn't any need to preset KEY_SHARE_EXTN_NR. Maybe I initialized it to "33" before writing the rest of the code and then left it that way.

One thing I just noticed, though. This code probably won't work as expected with a server that supports TLSv1.3 and that requires STARTTLS. I can't test against any servers that require STARTTLS, let alone one that supports TLSv1.3. Perhaps initializing KEY_SHARE_EXTN_NR to "33" would (by chance) allow the code to work with a STARTTLS server that supports TLSv1.3 draft 23, but not one that supports an earlier draft.

drwetter:
Thx for explaining it.

PR #960, opened 2018-01-17
TLSv1.3 related bugfix

dcooper16:
In three different places there is a line that is supposed to check whether the list of ciphers to be tested contains any TLSv1.3 ciphers. This check currently fails if there is only one TLSv1.3 cipher in the list and it is the first cipher in the list. This PR fixes the problem.


drwetter:
thx!

PR #961, opened 2018-01-18
Certificate-based auth host scanning produces invalid JSON output

tissieres:
When we scan a service with certificate-based authentication enabled with the `--json-pretty` output, the file generated is invalid:
```
"scanResult"  : [
                           {
                               "id"           : "client_auth",
                               "severity"     : "INFO",
                               "finding"      : "certificate-based authentication => skipping all HTTP checks"
                          }                    "protocols"         : [
                           {
                               "id"           : "sslv2",
                               "severity"     : "OK",
                               "finding"      : "SSLv2 is not offered"
```

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1472,7 +1472,7 @@ service_detection() {
           *)   if "$CLIENT_AUTH"; then
                     out " certificate-based authentication => skipping all HTTP checks"
                     echo "certificate-based authentication => skipping all HTTP checks" >$TMPFILE
-                    fileout "client_auth" "INFO" "certificate-based authentication => skipping all HTTP checks"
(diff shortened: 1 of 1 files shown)


drwetter:
Thx, @tissieres . I can reproduce the bug. But looking at the PR suggested I am not quite sure it'll help. Did you miss a commit?

tissieres:
Thanks to you for testssl.sh @drwetter !

No, that is the only change I needed to make it work correctly, but I am not sure of the impact of the change on the rest of the code.

drwetter:
Oops :-) That looked so wrong if one don't have a look @ the context.

Thx for the fix!

PR #965, opened 2018-01-19
Check keyUsage and extended key usage extensions

dcooper16:
This PR adds code to print the contents of the keyUsage and extended key usage extensions in certificates and to check that the public keys in the certificates are not being used in a manner that is inconsistent with these extensions (see #964).

The code for this PR was written a long time ago, so it is possible that the two lines for extracting the contents of the extensions (6573 and 6592) could be improved. They work, but use several  non-Bash functions.

Also, if this PR is accepted, then some `fileout()` calls should be added to at least include any error messages in the JSON/CSV outputs.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6346,11 +6346,12 @@ certificate_info() {
      local -i number_of_certificates=$2
      local cipher=$3
      local cert_keysize=$4
-     local ocsp_response=$5
-     local ocsp_response_status=$6
-     local sni_used=$7
-     local ct="$8"
-     local cert_sig_algo cert_sig_hash_algo cert_key_algo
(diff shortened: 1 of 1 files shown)


drwetter:
> The code for this PR was written a long time ago, so it is possible that the two lines for extracting the contents of the extensions (6573 and 6592) could be improved. They work, but use several non-Bash functions.

Yes, but there's not only one or two places where this needs to be improved. Especially ``certificate_info()``  and similar x509 calls were done a longer while back and leave space for improvements. It's also not relevant for execution time. So I'd opted rather to leave it like it is until this whole function will be improved , need to check BSDish ``sed`` though.

PR #968, opened 2018-01-22
Fix typos in HSTS tests: days -> seconds

jpstevens:
Noticed a typo whilst using the library earlier:

`HSTS timeout 180 days (=15552000 seconds) > 15465600 days`

Should be:

`HSTS timeout 180 days (=15552000 seconds) > 15465600 seconds`

There were some other typos/inconsistencies so I fixed them.


drwetter:
thanks!

Issue #970, opened 2018-01-23
Informational announcement: Changes for JSON ahead


drwetter:
Hi,

I've been working the last 2 days to get an improvement done in 2.9dev which will affect all folks using JSON (and CSV) as the output format for further processing. As the screen output won't be affected I thought a heads up would do good:

Following changes will be implemented shortly in 2.9dev:

1) The "id" will change in 90% of the cases. It'll be more verbose, acronyms with be written in capitals.

2) The repetition of the check name in "finding" will vanish. So instead of 

```
"finding" "Server banner identified: Apache" 
```

there will be only ``"finding": "Apache"``. The "id" will tell you where this finding belongs to: ``"id"           : "server_banner"``

This increases readability because of the redundant information in finding will disappear and for humans "id" will better explain what's this object is about. 

If you want to have a peak you should have a look @ the branch where this action is taking place: https://github.com/drwetter/testssl.sh/tree/new_json_for_serverdefaults. Some ids weren't converted yet but will be done soon -- either in that branch or after merge into dev.

Following pictures should give you an idea what I am talking about:

**OLD
===**
![screenshot_20180123_154809](https://user-images.githubusercontent.com/8036727/35282074-e3c72776-0054-11e8-8a2c-8e9cbd79d507.png)

**NEW
===**
![screenshot_20180123_154833](https://user-images.githubusercontent.com/8036727/35282449-d3c03d6c-0055-11e8-84e9-eb4218505d73.png)

drwetter:
in vim diff mode:
![screenshot_20180123_160250](https://user-images.githubusercontent.com/8036727/35282852-eba30ca6-0056-11e8-8da9-c6d4b5afb0c2.png)

bao7uo:
These are helpful changes, thanks :-)

drwetter:
closed via e446c80

drwetter:
I just committed some adjustments in d2d62a9.

As much as possible the ``jsonID``s were changed to OpenSSL / IETF names so that testssl.sh is more compliant to the rest of the world -- mainly in ``certificate_info()``. 

If you happen to spot improvements let me know or submit a PR.

bao7uo:
Just had a quick read through - looks ideal, thanks.

fogs:
Danke!

PR #971, opened 2018-01-25
New JSON id and findings (breaks things!)

drwetter:
This PR adds **breaking changes** to the JSON output parameters ``id`` and ``findings``.

``findings`` in JSON are now more crunchy and don't repeat parts of ``id``. Also ``id`` has changed so that it is more reflecting what has been tested.

**Further changes:**
The change from @dcooper16 wrt TLS_FALLBACK_SCSV still has to be changed in the same manner.

It also finally merges @dcooper16 PR about key usage and extended key usage extensions. ``fileout()`` handing has been added, it also works on BSDish sed flavors.

As the cipherlist checks in the beginning have been less and less to do with the OpenSSL standard lists a change to remove the word "standard" was long overdue. That has been addressed now also in the code and in the  JSON/CSV output.

``$HOSTCERT`` has now an ``.pem`` file extension. ``$HOSTCERT_TXT`` on the other hand will contain the text output of ``openssl x509 -in $HOSTCERT -text -noout`` operation which enables testssl.sh to remove some of the redundant calls.

``fileout <somestrings>`` has not been consistently replaced by ``fileout $jsonID`` yet.

### DIFF
--- a/t/31_badssl.com.t
+++ b/t/31_badssl.com.t
@@ -14,7 +14,7 @@ my (
 );
 # OK
 pass("Running testssl.sh against badssl.com to create a baseline (may take 2~3 minutes)"); $tests++;
-my $okout = `./testssl.sh -S -e -U --jsonfile tmp.json --color 0 badssl.com`;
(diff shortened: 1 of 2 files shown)


drwetter:
see #970

PR #973, opened 2018-01-26
Add TLSv1.3 support for run_server_defaults()

dcooper16:
This commit adds TLSv1.3 support for run_server_defaults(). It uses `get_server_certificate()` to run separate checks for RSA and ECDSA certificates by sending TLSv1.3 ClientHello messages with different signature_algorithms values. A similar change is made to `certificate_transparency()` in order to check for certificate transparency support for certificates returned over TLSv1.3. It also modifies the check for certificates offered without SNI by including an option to use `tls_sockets()` with servers that support TLSv1.3 if OpenSSL does not support TLSv1.3.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6037,6 +6037,38 @@ get_server_certificate() {
      local success
      local npn_params="" line
 
+     if [[ "$1" =~ "-cipher tls1_3" ]]; then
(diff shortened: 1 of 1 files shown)


drwetter:
Thank you for this and the other PRs.

PR #976, opened 2018-02-02
Add CVE commas for consistency

AresS31:
This is the only instance of `local cve` that I found without formatted on a non-comma-separated format.


drwetter:
Hi @AresS31 ,

thank you for the PR and pointing to an inconsistency.

If I look at other findings I am not sure yet whether I want to add commas here or remove the commas in other places with multiple CVEs are mentioned, probably the latter. That would be more consistent with e.g. ``PFS_ciphers`` and ``TLS_extensions``.

If you happen to have time to submit a PR where the commas between multiple CVEs are removed I would rather merge that one.

Thx, Dirk

PR #978, opened 2018-02-02
Fix HOSTCERT_TXT

dcooper16:
testssl.sh was recently changed to store the text printout of the host's certificate in a file, `$HOSTCERT_TXT`, and then use this file in some places rather than calling `$OPENSSL x509 -in $HOSTCERT -text -noout`. There was a problem, however, in cases in which the server had more than one certificate (including cases in which the server returned an unrelated certificate when sent a ClientHello w/o SNI), since the contents of `$HOSTCERT_TXT` was not always being updated whenever `$HOSTCERT` changed.

This PR fixes that problem by replacing the previous solution with a less ambitious one. In this version, the global variable is eliminated and instead `run_server_defaults()` stores a text version of each certificate it finds in an array. This value is then passed to `certificate_transparency()` and `certificate_info()` for use. It is also passed from `certificate_info()` to `must_staple()`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7280,6 +7280,10 @@ run_server_defaults() {
                          [[ $n -ge 10 ]] && sni_used[certs_found]="" || sni_used[certs_found]="$SNI"
                          tls_version[certs_found]="$DETECTED_TLS_VERSION"
                          previous_hostcert_type[certs_found]=" ${certificate_type[n]}"
+                         if [[ $DEBUG -ge 1 ]]; then
+                              echo "${previous_hostcert[certs_found]}" > $TEMPDIR/host_certificate_$certs_found.pem
+                              echo "${previous_hostcert_txt[certs_found]}" > $TEMPDIR/host_certificate_$certs_found.txt
+                         fi
                     else
                          previous_hostcert_type[i]+=" ${certificate_type[n]}"
                     fi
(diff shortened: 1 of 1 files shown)


dcooper16:
Based on a suggestion from Dirk, I just added a second commit to this PR that saves each certificate found by `run_server_defaults()` in a separate file in $TMPDIR, if testssl.sh is being run in debug mode. It saves both the PEM-encoded certificate and the pretty-print text version of the certificate.

testssl.sh
```diff
@@ -6292,13 +6290,13 @@ must_staple() {
...
      # previous versions of OpenSSL, OpenSSL can only show if the extension OID is present.
-     if $OPENSSL x509 -in "$HOSTCERT" -noout -text 2>>$ERRFILE | grep -A 1 "TLS Feature:" | grep -q "status_request"; then
+     if grep -A 1 "TLS Feature:" <<< "$hostcert_txt" | grep -q "status_request"; then
           # FIXME: This will indicate that must staple is supported if the
           # certificate indicates status_request or status_request_v2. This is
           # probably okay, since it seems likely that any TLS Feature extension
           # that includes status_request_v2 will also include status_request.
           supported=true
-     elif $OPENSSL x509 -in "$HOSTCERT" -noout -text 2>>$ERRFILE | grep -q "1.3.6.1.5.5.7.1.24:"; then
+     elif [[ "$hostcert_txt" =~ "1.3.6.1.5.5.7.1.24:" ]]; then
```
drwetter: cool!

drwetter:
great job! Thank you.

Nice that some matches are way easier to do

PR #982, opened 2018-02-06
Fix #981

### LINKED ISSUE #981: Wrong wildcard parsing in SAN
Hello,

I'm running testssl.sh on some domain (cannot disclose it due to NDA, sorry) and the following is output:

```
 Common Name (CN)             *.placeholder.ext
 subjectAltName (SAN)         *.placeholder.ext placeholder.ext
 Issuer                       Gandi Standard SSL CA 2 (Gandi from FR)
 Trust (hostname)             via CN wildcard, but not SAN -- Browsers are complaining (same w/o SNI)
```

The issue is the message `via CN wildcard, but not SAN -- Browsers are complaining`, since the wildcard is also in SAN.

Seems like there is some error in parsing the wildcard domain from the SAN.

If you want me to perform some tests on that domain for you, feel free to ask.

***

Output from `certtool -i` on the certificate:

```
X.509 Certificate Information:
	Version: 3
--<snip>--
	Subject: CN=*.placeholder.ext,OU=Gandi Standard Wildcard SSL,OU=Domain Control Validated
--<snip>--
		Subject Alternative Name (not critical):
			DNSname: *.placeholder.ext
			DNSname: placeholder.ext
```

***

Versions (testssl.sh cloned from github, commit 364011b6ca870f0fe6139246ad943d193d073286):

```
$ testssl.sh -b 2>/dev/null | head -4 | tail -2
testssl.sh       2.9dev from https://testssl.sh/dev/
$ testssl.sh -b 2>/dev/null | head -16 | tail -3
 on placeholder:/placeholder/testssl.sh/bin/openssl.Linux.x86_64
 (built: "Jun 22 19:32:29 2016", platform: "linux-x86_64")
$ uname -a
Linux placeholder 4.14.15-1-ARCH #1 SMP PREEMPT Tue Jan 23 21:49:25 UTC 2018 x86_64 GNU/Linux
```

dcooper16:
This PR fixes #981. It changes `get_san_dns_from_cert()` to place each DNS name in single quotes (') so that when the resulting string is used in a for loop, Bash does not try to perform wildcard completion. `compare_server_name_to_cert()` removes the single quotes (') from each name before comparing it against the presented name. `run_server_defaults()` does not need to remove the  single quotes ('), since all checks in that function continue to work with the quotes added.

An on-line search suggested that the best way to deal with Bash attempting to perform wildcard expansion in cases like this is to create an array, but placing the entries in quotes seems to work, and it is a much easier fix.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6235,26 +6235,27 @@ compare_server_name_to_cert()
 
      # Check whether any of the DNS names in the certificate match the servername
      dns_sans="$(get_san_dns_from_cert "$cert")"
-     for san in $dns_sans; do
-          [[ $(toupper "$san") == "$servername" ]] && ret=1 && break
-     done
+     while read san; do
+          [[ -n "$san" ]] && [[ $(toupper "$san") == "$servername" ]] && ret=1 && break
+     done <<< "$dns_sans"
 
      if [[ $ret -eq 0 ]]; then
           # Check whether any of the IP addresses in the certificate match the servername
           ip_sans=$($OPENSSL x509 -in "$cert" -noout -text 2>>$ERRFILE | grep -A2 "Subject Alternative Name" | \
                   tr ',' '\n' | grep "IP Address:" | sed -e 's/IP Address://g' -e 's/ //g')
-          for san in $ip_sans; do
-               [[ "$san" == "$servername" ]] && ret=1 && break
-          done
+          while read san; do
+               [[ -n "$san" ]] && [[ "$san" == "$servername" ]] && ret=1 && break
+          done <<< "$ip_sans"
      fi
 
      # Check whether any of the DNS names in the certificate are wildcard names
      # that match the servername
      if [[ $ret -eq 0 ]]; then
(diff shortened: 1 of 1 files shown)


Rogdham:
Instead of adding quotes, an alternative way could be to remove passname expension with `set -f`, like mentionned here: https://stackoverflow.com/a/13271009

```bash
A="a* b*"

( set -f
  for ex in $A ; do
      echo "$ex"
  done
)
```

dcooper16:
I came up with a simpler way to fix #981. It occurred to me that the problem with pathname expansion was not occurring in the part of `certificate_info()` that prints all of the names in the subjectAltName extension. So, I just copied that approach, using a `while` loop instead of a `for` loop. So, I replaced the initial fix with this simpler fix.

drwetter:
> I came up with a simpler way to fix #981. 

Indeed, thank you! Before this suggestion I was about to suggestion another solution (quoting by using backslashes). But this one is the best of all. :-)

PR #984, opened 2018-02-08
Fix #983

### LINKED ISSUE #983: Line jump missing after TLS_FALLBACK_SCSV test result (before SWEET32 results)
In the development testssl version (2.9dev from https://testssl.sh/dev/ (7bf070b 2018-02-08 14:02:24 -- ),  after the TLS_FALLBACK_SCSV result there seems to be a missing line jump before the SWEET32 result (which appears at the end of that line)

See attached screenshot.
image

dcooper16:
This PR fixes #983 by ensuring that the line printed just before calling `return` prints a newline character. It also fixes the problem that no output is sent to the JSON/CSV file in some cases in which no fallback is possible since the server does not support two different protocols below TLSv1.3.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -12535,7 +12535,7 @@ run_tls_fallback_scsv() {
                fileout "$jsonID" "HIGH" "only SSLv3 supported"
                return 7
                ;;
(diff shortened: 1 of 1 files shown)


drwetter:
thx

PR #993, opened 2018-02-15
Bugfix for extra extensions

dcooper16:
If extra extensions are provided to `socksend_tls_clienthello()`, then `socksend_tls_clienthello()` needs to determine what extensions were provided so that it doesn't add any of these extensions a second time. The code that was looping through the extra extensions to get the extension IDs was using the wrong value for the length of the string. This PR fixes the error.


drwetter:
thx!

PR #994, opened 2018-02-15
TLSv1.3 draft 24

dcooper16:
This commit adds support for [TLSv1.3 draft 24](https://tools.ietf.org/html/draft-ietf-tls-tls13-24). The only change between draft 23 and draft 24 is that draft 24 requires the version number of the record layer to be "0303" for a second ClientHello, whereas draft 23 allowed the version number for both the initial and second ClientHello to be "0301".

The current text reads:
```
   legacy_record_version  This value MUST be set to 0x0303 for all
      records generated by a TLS 1.3 implementation other than an
      initial ClientHello (i.e., one not generated after a
      HelloRetryRequest), where it MAY also be 0x0301 for compatibility
      purposes.
```

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4687,6 +4687,11 @@ run_protocols() {
                          [[ -n "$drafts_offered" ]] && drafts_offered+=", "
                          drafts_offered+="draft 23"
                     fi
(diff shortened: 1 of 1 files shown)


drwetter:
cool, thank you, David!

PR #1002, opened 2018-02-22
Using generic HAProxy name

danielchodusov:
This prime appears to be not only in HAProxy 1.5 but as well in the newer versions. The test result will return incorrect response message, when testing on the newer HAProxy versions (ie. 1.5 is detected but 1.8 is installed).


drwetter:
thx!

FYI: It was imported from  https://svn.nmap.org/nmap/scripts/ssl-dh-params.nse

danielchodusov:
Thanx, I will notify them.

Issue #988, opened 2018-02-12
Empty finding causes corrupt json files


seccubus:
#Please make sure that you provide enough information so that we understand what your issue is about.

   
1. testssl version from the banner (testssl.sh -b 2>/dev/null | head -4 | tail -2)

    testssl.sh       2.9dev fromhttps://testssl.sh/dev/
    (9b37932 2018-02-12 15:44:39 -- )

2. what exactly was happening, output is needed

We start seccubus with the following command line:
```
testssl.sh  --color 0 -p -S -P -h -U --fast --jsonfile=/tmp/seccubus.testssl.82454.json --file /tmp/seccubus.hosts.82452
```

/tmp/seccubus.hosts.82452:
```
www.seccubus.com
```

Output at offset 1490 is:

```
,         {
              "id"           : "ALPN",
              "ip"           : "www.seccubus.com/95.101.39.93",
              "port"         : "443",
              "severity"     : "OK",

          }
,         {
              "id"           : "ALPN",
              "ip"           : "www.seccubus.com/95.101.39.93",
              "port"         : "443",
              "severity"     : "INFO",
              "finding"      : "http/1.1"
          }
```

3. what did you expect instead?

Output should be something like:
```
,         {
              "id"           : "ALPN",
              "ip"           : "www.seccubus.com/95.101.39.93",
              "port"         : "443",
              "severity"     : "OK",
              "finding"      : "(empty)"
          }
,         {
              "id"           : "ALPN",
              "ip"           : "www.seccubus.com/95.101.39.93",
              "port"         : "443",
              "severity"     : "INFO",
              "finding"      : "http/1.1"
          }
```

Actually I would expect only one ALPN finding, I suspect this is because of the fileout statement on line 8027

4. steps to reproduce

   1. testssl.sh command line

testssl.sh  --color 0 -p -S -P -h -U --fast --jsonfile=/tmp/seccubus.testssl.82454.json --file /tmp/seccubus.hosts.82452

    1. if possible: target IP

seccubus.com on 95.101.39.95 (Subject to change as this is behind Akamai)
 
    1. openssl version used (testssl.sh -b 2>/dev/null | head -16 | tail -3)
 
 Using "OpenSSL 1.0.2-chacha (1.0.2i-dev)" [~183 ciphers]
 on sbplt90dhhv2j:./bin/openssl.Darwin.x86_64
 (built: "Sep  7 19:34:54 2016", platform: "darwin64-x86_64-cc")

[...]

PR #1009, opened 2018-03-12
Fixing issue #988

danielchodusov:
This will fix empty trailing comma in JSON output #988

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -849,6 +849,7 @@ fileout() {
 
      if ( "$do_pretty_json" && [[ "$1" == "service" ]] ) || show_finding "$severity"; then
          local finding=$(strip_lf "$(newline_to_spaces "$(strip_quote "$3")")")
+         [[ -z "$finding" ]] && finding="(empty)"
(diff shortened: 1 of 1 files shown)


drwetter:
Hi Daniel,

thx for the PR but that should be fixed, see commit in the referenced, closed issue.

Or do still see this? If so please provide details.

Thx, Dirk
-- 

danielchodusov:
Hello Dirk,

yes it's not fixed yet - still problem with ALPN. Try this website:

testssl.sh --fast --parallel --sneaky --quiet --ip=one -o out.json hrad.cz and see line 62 by - 63 the line is ending with comma.

This fix will insert "(empty)" value as @seccubus suggested in his fix.
screen shot 2018-03-13 at 11 39 11

drwetter:
My improvement in #988 wasn't complete, i.e. when HTTP/2 is the only protocol you ran into the same problem. Like you did.

But as said in seccubus' PR: The fix should be better at the root cause, rater than patching an output function if a string is empty. So, thanks for the PR, Daniel. My fix follows

drwetter:
PS: ``-parallel`` currently only makes sense if you test more than 1 host.

PR #1010, opened 2018-03-13
Replace "--" with colorful descriptions to match JSON severity

krissi:
I noticed this issue for HSTS, which was uncolorized on stdout, but HIGH (=red for me) in JSON. I also got rid of the `--` output values to make them (hopefully) more meaningful and create a bigger canvas for the ansi colors.

I used `pr_svrty_good` for `INFO`. Please verify I didn't mess up the meaning.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1808,7 +1808,7 @@ run_hsts() {
                #      https://chromium.googlesource.com/chromium/src/+/master/net/http/transport_security_state_static.json
           fi
      else
-          out "--"
+          pr_svrty_high "not offered"
           fileout "HSTS" "HIGH" "not offered"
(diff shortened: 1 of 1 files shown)


drwetter:
> I used ``pr_svrty_good`` for ``INFO``. 

That is not correct, see ``testssl.sh(1)``, section ``COLOR RATINGS``.

> I also got rid of the -- output values to make them (hopefully) more meaningful and create a bigger canvas for the ansi colors.

That could make sense, need to align it with the rest.

Labeling for HSTS is not correct as file output is HIGH and screen is not.

drwetter:
Thx, @krissi . See commit log

PR #1012, opened 2018-03-14
TLS 1.3 drafts 25 and 26

dcooper16:
Drafts 25 and 26 of TLS 1.3 were issued in early March. Neither made any change that is relevant to testssl.sh.

This PR adds checks for these two drafts.


drwetter:
Thanks for the update!

PR #1014, opened 2018-03-15
Check for TLSv1.3 in OpenSSL

dcooper16:
This PR adds a missing check for local TLSv1.3 support in cipher_pref_check(). If the "--ssl-native" mode is being used and $OPENSSL does not support TLSv1.3, then a warning should be displayed indicating that TLSv1.3 cannot be checked rather than displaying the same results as if the server did not support TLSv1.3.


drwetter:
thank you!

PR #1017, opened 2018-03-16
Fix #1013

### LINKED ISSUE #1013: OpenSSL 1.1.1 and cipher suite lists
A PR was just merged into the OpenSSL 1.1.1 development branch that will require significant changes to testssl.sh in order for it to support use with OpenSSL 1.1.1: see https://github.com/openssl/openssl/pull/5392.

The relatively simple change in https://github.com/openssl/openssl/pull/5392 is that it changes the OpenSSL names for the TLS 1.3 cipher suites. For example, TLS13-AES-128-GCM-SHA256 was changed to TLS_AES_128_GCM_SHA256. When using OpenSSL names, testssl.sh currently checks whether the name begins with "TLS13" to determine whether it is a TLS 1.3 cipher suite.

The more difficult change to address is that OpenSSL 1.1.1 now uses different parameters to configure TLS 1.3 cipher suites from the one used to configure TLS 1.2 and below cipher suites. This affects both the `cipher` command and the `-cipher` option of the `s_client` command.

An example of the change is the following:
```
> ./old_openssl/openssl ciphers -V RC2
Error in cipher list
140629856110400:error:1410D0B9:SSL routines:SSL_CTX_set_cipher_list:no cipher match:ssl/ssl_lib.c:2535:

> ./PR5392_openssl/openssl ciphers -V RC2
          0x13,0x02 - TLS_AES_256_GCM_SHA384  TLSv1.3 Kx=any      Au=any  Enc=AESGCM(256) Mac=AEAD
          0x13,0x03 - TLS_CHACHA20_POLY1305_SHA256 TLSv1.3 Kx=any      Au=any  Enc=CHACHA20/POLY1305(256) Mac=AEAD
          0x13,0x01 - TLS_AES_128_GCM_SHA256  TLSv1.3 Kx=any      Au=any  Enc=AESGCM(128) Mac=AEAD

> ./PR5392_openssl/openssl ciphers -V -ciphersuites TLS_CHACHA20_POLY1305_SHA256 RC2
          0x13,0x03 - TLS_CHACHA20_POLY1305_SHA256 TLSv1.3 Kx=any      Au=any  Enc=CHACHA20/POLY1305(256) Mac=AEAD
```

Starting with https://github.com/openssl/openssl/pull/5392, if `-ciphersuites` is not present, then it is the same as if it were present with a default list of TLS 1.3 cipher suites (TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256).

Similarly with the `s_client` command, if `-ciphersuites` is not present and `-no_tls1_3` is not present, the result is the same as if `-ciphersuites` were present and contained a list of the three defaults TLS 1.3 cipher suites.

I'll try to work on a PR (or set of PRs) to address this.

dcooper16:
This PR fixes the issue raised in #1013. It primarily does this in two ways:

* In calls to `$OPENSSL s_client` that specify ciphers, the TLSv1.3 ciphers are provided separately using the `-ciphersuites` option. Then, the `s_client_options()` function manipulates the command-line options as necessary based on the version of OpenSSL being used.

* Calls to `$OPENSSL ciphers` were replaced with calls to `actually_supported_ciphers()`, which calls `$OPENSSL ciphers`. `actually_supported_ciphers()` modifies the parameters for the call to `$OPENSSL ciphers` as necessary, based on the version of OpenSSL being used.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -299,6 +299,7 @@ HAS_PKUTIL=false
 HAS_PKEY=false
 HAS_NO_SSL2=false
 HAS_NOSERVERNAME=false
+HAS_CIPHERSUITES=false
 HAS_ALPN=false
 HAS_SPDY=false
 HAS_FALLBACK_SCSV=false
@@ -412,6 +413,7 @@ declare TLS_CIPHER_AUTH=()
 declare TLS_CIPHER_ENC=()
 declare TLS_CIPHER_EXPORT=()
 declare TLS_CIPHER_OSSL_SUPPORTED=()
+declare TLS13_OSSL_CIPHERS="TLS_AES_128_GCM_SHA256:TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_CCM_SHA256:TLS_AES_128_CCM_8_SHA256"
 
 ###### output functions ######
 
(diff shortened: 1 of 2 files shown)


drwetter:
Thx, good work!

testssl.sh
```diff
@@ -1067,8 +1069,20 @@ count_ciphers() {
      echo $(wc -w <<< "${1//:/ }")
 }
 
+#arg1: TLS 1.2 and below ciphers
+#arg2: TLS 1.3 ciphers
+#arg3: options (e.g., -V)
 actually_supported_ciphers() {
```
drwetter: Actually your PR made clear somehow that $OPENSSL is called often (atm: 27x) to generate lists.

I was wondering whether it would make sense for longer terms to have a variable which contains everything  (``ALL:COMPLEMENTOFALL:@STRENGTH``) and to extract here the ones needed?

Issue #738, opened 2017-05-08
client simulation --ssl-native


drwetter:
This result (``./testssl.sh --client-simulation --ssl-native  --color 0 testssl.sh`` with supplied binary) is odd:

```
Android 2.3.7                    No connection
 Android 4.1.1                    No connection
 Android 4.2.2                    No connection
 Android 4.4.2                    No connection
 Android 5.0.0                    No connection
 Android 6.0                      TLSv1.0 ECDHE-RSA-AES256-SHA, 256 bit ECDH (P-256)
 Android 7.0                      TLSv1.0 ECDHE-RSA-AES256-SHA, 256 bit ECDH (P-256)
 Baidu Jan 2015                   No connection
 BingPreview Jan 2015             No connection
 Chrome 48 OS X                   TLSv1.0 ECDHE-RSA-AES256-SHA, 256 bit ECDH (P-256)
 Chrome 51 Win 7                  TLSv1.0 ECDHE-RSA-AES256-SHA, 256 bit ECDH (P-256)
 Edge 13 Win 10                   TLSv1.0 ECDHE-RSA-AES256-SHA, 256 bit ECDH (P-256)
 Edge 13 Win Phone 10             TLSv1.0 ECDHE-RSA-AES256-SHA, 256 bit ECDH (P-256)
 Firefox 45 Win 7                 TLSv1.0 ECDHE-RSA-AES256-SHA, 256 bit ECDH (P-256)
 Firefox 49 Win 7                 TLSv1.0 ECDHE-RSA-AES256-SHA, 256 bit ECDH (P-256)
 Firefox 49 XP SP3                TLSv1.0 ECDHE-RSA-AES256-SHA, 256 bit ECDH (P-256)
 Googlebot Feb 2015               No connection
 IE 11 Win 10                     TLSv1.0 ECDHE-RSA-AES256-SHA, 256 bit ECDH (P-256)
 IE 11 Win 7                      TLSv1.0 ECDHE-RSA-AES256-SHA, 256 bit ECDH (P-256)
 IE 11 Win 8.1                    TLSv1.0 ECDHE-RSA-AES256-SHA, 256 bit ECDH (P-256)
 IE 11 Win Phone 8.1              No connection
 IE 11 Win Phone 8.1 Update       No connection
 IE 6 XP                          No connection
 IE 7 Vista                       No connection
 IE 8 Win 7                       No connection
 IE 8 XP                          No connection
 Java 6u45                        No connection
[...]

PR #1028, opened 2018-04-09
Fix false connectivity problem error

dcooper16:
If OpenSSL reports an error, `sclient_connect_successful()` may incorrectly interpret it as a connectivity problem, leading testssl.sh to stop testing before it has completed.

When not using "--ssl-native" mode, this happens if $OPENSSL does not support SSLv3, as both `get_server_certificate()` and `run_beast()` will attempt to connect using SSLv3 even if $OPENSSL does not support it.

When using "--ssl-native" mode, this happens in multiple places if $OPENSSL does not support the protocol being used or if $OPENSSL does not support any of the ciphers that are specified in the command line.

This PR fixes the above problems by adding checks for protocol support or for support for at least one cipher before calling $OPENSSL.

`run_client_simulation()` also has a problem in "--ssl-native" mode of calling $OPENSSL with parameters that cause $OPENSSL to report an error, but this is already addressed by temporarily setting MAX_OSSL_FAIL to 100 during client simulation tests and then, after client simulation testing is complete, returning $MAX_OSSL_FAIL and $NR_OSSL_FAIL to the values they had before client simulation testing began.

NOTE: When using "--ssl-native" mode, there is a potential problem in `run_ssl_poodle()` and `run_lucky13()` if $OPENSSL does not support any CBC ciphers. (In the case of `run_ssl_poodle()` this only applies if $OPENSSL supports SSLv3.) This PR does not address this problem since it appears that all versions of OpenSSL support CBC ciphers.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4363,19 +4363,17 @@ locally_supported() {
 run_prototest_openssl() {
      local -i ret=0
 
+     ! locally_supported "$1" "$2" && return 7
      $OPENSSL s_client $(s_client_options "-state $1 $STARTTLS $BUGS -connect $NODEIP:$PORT $PROXY $SNI") >$TMPFILE 2>$ERRFILE </dev/null
      sclient_connect_successful $? $TMPFILE
      ret=$?
      debugme egrep "error|failure" $ERRFILE | egrep -av "unable to get local|verify error"
-     if ! locally_supported "$1" "$2" ; then
-          ret=7
-     else                                    # try again without $PROXY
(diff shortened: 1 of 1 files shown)


drwetter:
Thank you, David.

WRT ``run_client_simulation()``: yes. The hack (mine :-) ) is not the greatest thing on earth. Worse is however that "--ssl-native" mode gives different results, see #738. 

CBC ciphers: Thanks for the information. Practically you're right probably, lets keep this little catch in mind.

PR #1029, opened 2018-04-10
Fix typos found by codespell

knweiss:
`codespell` can be found here: https://github.com/lucasdemarchi/codespell


drwetter:
Appreciate, Thanks!

I can't change the OpenSSL license text however. Care to remove that part?

knweiss:
@drwetter I agree this makes sense, done.

PR #1030, opened 2018-04-10
Fix RFC 7919 DH groups

dcooper16:
This PR fixes three issues related to the testing for RFC 7919 DH groups in `run_pfs()`:

* If the RFC 7919 DH groups are supported for both TLSv1.3 cipher suites and non-TLSv1.3 cipher suites, then the list of supported groups is printed twice.

* The finding that is used for CSV/JSON files includes the word "offered" after the list of groups, which is inconsistent with other findings.

* Since the `$ffdhe_offered` is only used to determine whether to test for use of RFC 7919 DH groups with non-TLSv1.3 ciphers, this flag should only be set if a non-TLSv1.3 cipher that uses ephemeral DH is found.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7983,8 +7983,10 @@ run_pfs() {
                               fi
                          fi
                     fi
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks again!

PR #1032, opened 2018-04-12
Fix mass testing

dcooper16:
https://github.com/drwetter/testssl.sh/commit/b2be380b5465bcf8a8fc028d943a2687df10c8f8 inadvertently changed `MASS_TESTING_CMDLINE` to be a read-only variable. This causes mass testing to fail, since in mass testing the value of `MASS_TESTING_CMDLINE` is set to the command line for each child test.


drwetter:
Thank you

PR #1033, opened 2018-04-13
TLS 1.3 clients in update_client_sim_data.pl

dcooper16:
https://api.dev.ssllabs.com/api/v3/getClients incorrectly indicates a highestProtocol of 771 (TLS 1.2) for clients that support TLS 1.3, which leads `run_client_simulation()` to incorrectly report "No connection" if the client would have actually connected using TLS 1.3.

This has been addressed by manually editing etc/client-simulation.txt to set the highest_protocol to 0x0304 for the clients that support TLS 1.3.

This PR modifies update_client_sim_data.pl to automatically apply the fix for clients that support TLS 1.3 in order to avoid a possible regression when etc/client-simulation.txt is updated.


drwetter:
thx

drwetter:
screenshot_20180416_113235

cool!

PR #1034, opened 2018-04-13
update_client_sim_data.pl and GREASE ciphers

dcooper16:
Two GREASE ciphers currently appear in https://api.dev.ssllabs.com/api/v3/getClients: 0x3A3A for Chrome 57 and 0xAAAA for Chrome 65.

update_client_sim_data.pl currently only recognizes 0x3A3A as a GREASE cipher and so prints a "FIXME" for 0xAAAA. This PR fixes the problem by adding all 16 ciphers from https://tools.ietf.org/html/draft-ietf-tls-grease-00 to update_client_sim_data.pl.


drwetter:
thx

Issue #845, opened 2017-10-06
Testing server preferences


davidepic91:
If in my server (socket) I use the following configuration for the cipher suites:
```
#define SSL_CIPHERS "TLSv1.2:!eNULL:!SHA256"  
//other code
SSL_CTX_set_cipher_list(ctx, SSL_CIPHERS);
```

Then, in the test's output I obtain:

> **Testing server preferences**  
> Has server cipher order?     no matching cipher in this list found (pls report this): DES-CBC3-SHA:RC4-MD5:DES-CBC-SHA:RC4-SHA:AES128-SHA:AES128-SHA256:AES256-SHA:ECDHE-RSA-AES128-SHA:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA:ECDH-RSA-DES-CBC3-SHA:ECDH-RSA-AES128-SHA:ECDH-RSA-AES256-SHA:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:DHE-DSS-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-GCM-SHA256:AES256-SHA256 .
...
> **Testing 359 ciphers via OpenSSL plus sockets against the server, ordered by encryption strength**
>  x9d     AES256-GCM-SHA384                 RSA        AESGCM      256      TLS_RSA_WITH_AES_256_GCM_SHA384                    
>  x9c     AES128-GCM-SHA256                 RSA        AESGCM      128      TLS_RSA_WITH_AES_128_GCM_SHA256 
 
Is it OK ? Or is there an issue ?

drwetter:
Hi Davide,

On 06.10.2017 16:25, Davide Piccardi wrote:
> If in my server (socket) I use the following configuration for the cipher suites:
> 
> |#define SSL_CIPHERS "TLSv1.2:!eNULL:!SHA256" //other code SSL_CTX_set_cipher_list(ctx,
> SSL_CIPHERS); |
> 
> Then, in the test's output I obtain:
> 
>     *Testing server preferences*
>     Has server cipher order? no matching cipher in this list found (pls report this):
[...]

PR #1041, opened 2018-04-25
Update ciphers for server preference

dcooper16:
This PR was developed in response to #845. It adds to the list of ciphers used to determine whether the server has a cipher order in order to help avoid cases in which testssl.sh cannot determine a cipher order.

In order to create this list I scanned thousands of servers in order to determine what ciphers they support, including (1) about 20 thousand U.S. government web sites, (2) all of the sites listed at badssl.com, (3) all of the test servers listed at https://github.com/tlswg/tls13-spec/wiki/implementations, (4) about 30 additional non-U.S. government sites, and (5) one server configured as described in #845. I scanned each of these servers using OpenSSL 1.0.2-chacha, 1.0.2o, and 1.1.1.

Then I ran the collected information through a script that created the updated list. For each scanned server, and for each of the 3 versions of OpenSSL, the script checked whether $list_fwd contained at least two ciphers from the list of ciphers supported by the server. If it didn't, then it would add one of the ciphers supported by the server (and by OpenSSL) to $list_fwd. In choosing among the ciphers supported by the server that were not already in $list_fwd, it would choose the cipher that was supported by the most other servers.

The updated $list_fwd contains a few oddities as a result of the servers that I scanned. The script added two TLSv1.3 ciphers, since I scanned at least one server that only supports TLSv1.3. The list also includes ADH-AES256-GCM-SHA384 and AECDH-AES128-SHA, which may only be supported by null.badssl.com.

I made one manual change to the list - adding TLS_CHACHA20_POLY1305_SHA256. I did this since the number of TLSv1.3 servers scanned was so small, I didn't think it was safe to assume that all servers that support TLSv1.3 would support both TLS_AES_256_GCM_SHA384 and TLS_AES_128_GCM_SHA256.

Since most of the servers that I scanned were U.S. government servers, it may not be a representative sample. However, since the new list only adds to the current list, it can only be an improvement. Also, the updated list still only includes 37 ciphers, so many more could be added without creating any problems.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -5422,9 +5422,10 @@ run_server_preference() {
      local -a cipher proto
      local proto_ossl proto_txt proto_hex cipherlist i
      local -i ret=0 j sclient_success str_len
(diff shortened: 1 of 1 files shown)


drwetter:
cool, thank you! Also the two TLSv1.3 ciphers are another good step forward.

I suppose the US government servers are in a better shape as the average server out there.  The remaining maybe ~5-10% from https://github.com/drwetter/testssl.sh/issues/845#issuecomment-335862315 is IMO a matter of fine tuning after next release.

PR #1042, opened 2018-04-26
non-HTTP cRLDistributionPoints

dcooper16:
At the moment the code for downloading a CRL seems to only work if the URL is an HTTP or HTTPS URL. It fails if the URL is an LDAP URL. The `wget` command does not support LDAP and when `curl` retrieves data from an LDAP URL it stores the result in LDIF format, which `http_get()` cannot currently convert into a PEM-encoded CRL.

This PR addresses the issue by skipping the revocation check for any URL that does not begin with "http".

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1431,6 +1431,8 @@ check_revocation_crl() {
      local tmpfile=""
 
      "$PHONE_OUT" || return 0
+     # The code for obtaining CRLs only supports HTTP and HTTPS URLs.
+     [[ "$(tolower "${crl:0:4}")" == "http" ]] || return 0
      tmpfile=$TEMPDIR/${NODE}-${NODEIP}.${crl##*\/} || exit $ERR_FCREATE
 
      http_get "$crl" "$tmpfile"


drwetter:
Sure. 

Didn't know curl supports LDAP too. I stumbled often over it for internal CAs... (Microsoft product). And I didn't know/remember the scheme part can be HTTP**S** (according to RFC 5280 which you probably know better ;) )

drwetter:
Realistically the HTTPS part can be skipped. Not that I have the best statistics ever but I never encountered HTTPS but I am happy to learn ;-)  (In older revisions curl and wget had problems with the certificate chain. But as long as that is fine it _should_ work). 

LDAP OTOH seems interesting to me also because of it's usage and thus maybe a topic for 3.1dev.

You have any statistics during your scans? (the Alexa scans I did in December were just testing ROBOT).

PS: Can't remember seeing FTP.

dcooper16:
I just submitted #1045, which adds support for LDAP. It turned out to be easier than I thought it would be to extract the CRL from the LDIF file.

I haven't done any large-scale scans of servers recently, but I would hope that HTTPS isn't used. (The large-scale scan I did recently was just with the "-e" option.)

PR #1045, opened 2018-04-26
Support LDAP crLDistrubutionPoints

dcooper16:
This PR adds support for retrieving CRLs from an LDAP URL in the cRLDistributionPoints extension if curl if available.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1429,14 +1429,29 @@ check_revocation_crl() {
      local crl="$1"
      local jsonID="$2"
      local tmpfile=""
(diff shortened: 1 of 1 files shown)


drwetter:
thanks, merged to https://github.com/drwetter/testssl.sh/tree/crl_ocsp . 

Think it's better to make CRL (and maybe OCSP too) certificate revocation checks more solid here before merging it.

drwetter:
That is really cool btw!

Just checked it against a couple of servers with LDAP URI. Thanks

PR #1049, opened 2018-05-01
Don't offer TLS compression by default

dcooper16:
There is at least one server that will fail under some circumstances if the ClientHello offers a compression method other than null.

In OpenSSL 1.1.0 and 1.1.1, s_client will not offer any other compression methods unless the "-comp" option is provided. However, in earlier versions of OpenSSL, s_client will by default offer the DEFLATE compression method, however, this can be disabled using the "-no_comp" option.

This PR addresses the flaw in this server by having `s_client_options()` add a "-no_comp" option to the command line if "-no_comp" is supported and the test doesn't require offering compression.

Since run_crime() requires compression to be offered, `run_crime()` was changed to always add "-comp" to the command line, and then `s_client_options()` was changed to remove "-comp" from the command line, if that option isn't supported.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -316,6 +316,8 @@ HAS_PKEY=false
 HAS_NO_SSL2=false
 HAS_NOSERVERNAME=false
 HAS_CIPHERSUITES=false
+HAS_COMP=false
+HAS_NO_COMP=false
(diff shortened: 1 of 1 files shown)


drwetter:
thank you!

Issue #1046, opened 2018-04-28
fail to validate google.TLD


drwetter:
The recent feature of revocation checking via CRL has a problem when checking google.TLD hosts. It says "revoked" but in fact it is not.

it looked like the ``GeoTrust Global CA`` is the cuplrit:

```
prompt% openssl verify -verbose -crl_check -CAfile /tmp/testssl.rI7CYA/google.com-216.58.207.46-CRL-chain.pem /tmp/testssl.rI7CYA/host_certificate.pem
/tmp/testssl.rI7CYA/host_certificate.pem: C = US, O = GeoTrust Inc., CN = GeoTrust Global CA
error 2 at 2 depth lookup:unable to get issuer certificate
prompt%
```

To me it looks like the (old and rotten, thus retired) ``Equifax Secure Certificate Authority`` (https://www.geotrust.com/resources/root-certificates/retired.html) is missing.

At least when concatting this certificate too it says OK:

```
prompt% cat /tmp/equifax.pem /tmp/testssl.rI7CYA/google.com-216.58.207.46-CRL-chain.pem >/tmp/all.pem
prompt% openssl verify  -verbose -crl_check -CAfile /tmp/all.pem /tmp/testssl.rI7CYA/host_certificate.pem                            crl_ocsp
/tmp/testssl.rI7CYA/host_certificate.pem: OK
prompt%
```

This problem is no problem anymore soon (probably) as the root expires in August 2018.

PR #1051, opened 2018-05-02
Fix some problems with CRL revocation checking

dcooper16:
There are some circumstances in which `check_revocation_crl()` will incorrectly indicate that a CRL lists the server's certificate as revoked. #1046 is one of them. Another is any case in which the server's certificate cannot be validated using any of the certificates in the trust store that OpenSSL uses (e.g., the server's certificate was issued by a local CA). In both of these cases, "openssl verify" fails, for some reason other than "certificate revoked", and `check_revocation_crl()` assumes that any failure of "openssl verify" is the result of certificate revocation.

This PR addresses the problem in two ways. First, it adds the "-partial_chain" option to the "openssl verify" command line whenever `$OPENSSL` supports that option (it is not supported by LibreSSL or by versions of OpenSSL earlier than 1.0.2). This will fix most of the problems when a version of OpenSSL that supports "-partial_chain" is used.

However, even if the "-partial_chain" option is provided, OpenSSL needs to have at least one CA certificate so that it can get the public key needed to verify the signatures on the server's certificate and on the CRL. So, if the server doesn't send any CA certificates and the server's certificate was not issued by a CA in the trust store, then the verify command will fail even if the "-partial_chain" option is provided.

So, as a fail-safe, this PR changes `check_revocation_crl()` to check the error message that the verify command provides when it fails so that testssl.sh only reports a certificate a revoked if the verify command fails with a reason of "certificate revoked".

Note that this PR also fixes two other minor issues. It incorporates #1047, which corrects a typo, and it redirects $OPENSSL's output on line 1479 in order to suppress any error messages that $OPENSSL might print (e.g., "WARNING: can't open config file").

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -318,6 +318,7 @@ HAS_NOSERVERNAME=false
 HAS_CIPHERSUITES=false
 HAS_COMP=false
 HAS_NO_COMP=false
+HAS_PARTIAL_CHAIN=false
 HAS_ALPN=false
 HAS_NPN=false
 HAS_FALLBACK_SCSV=false
@@ -1451,7 +1452,7 @@ check_revocation_crl() {
      local crl="$1"
      local jsonID="$2"
      local tmpfile=""
-     local scheme
+     local scheme retcode partial_chain=""
      local -i success
 
      "$PHONE_OUT" || return 0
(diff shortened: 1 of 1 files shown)


dcooper16:
Closing this PR as #1067 provides an alternative solution.

PR #1052, opened 2018-05-04
TLS 1.3 draft 28

dcooper16:
https://github.com/tlswg/tls13-spec/wiki/implementations now lists a server that supports TLS 1.3 draft 28, so this PR adds support for drafts 27 and 28.

Since `run_protocols()` now checks for 11 different drafts of TLS 1.3 in addition to the final version, performing a separate test for each draft had become far too time consuming. So, this PR rewrites the check for TLS 1.3 drafts in `run_protocols()` so that the number of tests is proportional to the number of drafts that the server supports rather than the number of drafts that testssl.sh can check for.


drwetter:
cool thanks!

PR #1060, opened 2018-05-18
HTTP URLs with multiple components in path

dcooper16:
In some cases the OCSP URI contains multiple components in the path (e.g., http://www.example.com/OCSP/myOCSPresponder).

This PR changes `check_revocation_ocsp()` to remove all components in the path, rather than just the final component, when extracting the host name from the URI for the host header.


drwetter:
argh, you're right. Thx

PR #1064, opened 2018-05-23
OCSP improvements

dcooper16:
This PR fixes two issues with OCSP checking. First, the syntax for specifying a host header changed in OpenSSL between versions 1.0.2 and 1.1.0. With OpenSSL 1.0.2-chacha, 1.0.2o, and LibreSSL the syntax needs to be "-header Host \<hostname\>". With OpenSSL 1.1.0h and 1.1.1 the syntax needs to be "-header Host=\<hostname\>". I have not been able to test other versions of OpenSSL 1.0.2 or 1.1.0, but am assuming that all versions of OpenSSL 1.1.0 use the same syntax as 1.1.0h.

This PR also fixes a typo in the case of an error, which was causing $code to be set to "empty ocsp response" if the response was not empty rather than if it was empty.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1506,7 +1506,12 @@ check_revocation_ocsp() {
      tmpfile=$TEMPDIR/${NODE}-${NODEIP}.${uri##*\/} || exit $ERR_FCREATE
      host_header=${uri##http://}
      host_header=${host_header%%/*}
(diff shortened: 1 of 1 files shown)


drwetter:
Thx!

Your PR makes it probably necessary to readjust the / work on a correct error message ;-/

PR #1066, opened 2018-05-24
OCSP error handling

dcooper16:
This PR improves the handling of error responses when checking status using OCSP. It can handle a few types of errors:

* When the responder just returns an error (e.g., "Responder error: unauthorized").

* When the response cannot be verified (e.g., invalid signature, expired certificate).

* When the response is valid ("Response verify OK"), but there is a problem with the response for the individual certificate (e.g., information is too old, or status is "unknown").


drwetter:
Thank you, David

Issue #1057, opened 2018-05-17
Fail to do revocation checking on hosts, other than on google.TLD (and similar) (#1046)


drwetter:
Those hosts also show a problem (hosts from Alexa Top-something):

```
prompt% grep 'Certificate Revocation List' * | grep -v "not revoked" | grep revoked                                                                                   
adidas.com: Certificate Revocation List  http://crl.entrust.net/level1k.crl, revoked
banggood.com: Certificate Revocation List  http://cdp1.digicert.com/ssca-sha2-g6.crl, revoked
biobiochile.cl: Certificate Revocation List  http://crl.comodoca.com/COMODORSADomainValidationSecureServerCA.crl, revoked
ca.gov: Certificate Revocation List  http://crl.comodoca.com/COMODORSAOrganizationValidationSecureServerCA.crl, revoked
csdn.net: Certificate Revocation List  http://cdp1.digicert.com/ssca-sha2-g6.crl, revoked
expedia.com: Certificate Revocation List  http://crl.entrust.net/level1k.crl, revoked
hitcpm.com: Certificate Revocation List  http://crl.comodoca.com/COMODORSADomainValidationSecureServerCA.crl, revoked
mozilla.org: Certificate Revocation List  http://crl3.digicert.com/sha2-ev-server-g2.crl, revoked
nba.com: Certificate Revocation List  http://crl.comodoca.com/COMODORSAOrganizationValidationSecureServerCA.crl, revoked
okta.com: Certificate Revocation List  http://crl3.digicert.com/sha2-ha-server-g5.crl, revoked
onlinesbi.com: Certificate Revocation List  http://sr.symcb.com/sr.crl, revoked
sourceforge.net: Certificate Revocation List  http://crl.comodoca.com/COMODORSADomainValidationSecureServerCA.crl, revoked
thefreedictionary.com: Certificate Revocation List  http://crl.comodoca.com/COMODORSADomainValidationSecureServerCA.crl, revoked
tianya.cn: Certificate Revocation List  http://crl.entrust.net/level1k.crl, revoked
tribunnews.com: Certificate Revocation List  http://gp.symcb.com/gp.crl, revoked
ups.com: Certificate Revocation List  http://crl.comodoca.com/COMODORSAExtendedValidationSecureServerCA.crl, revoked
zhanqi.tv:   Certificate Revocation List  http://ss.symcb.com/ss.crl, revoked
[...]

Issue #1056, opened 2018-05-17
error querying OCSP responder


drwetter:
PR #1055 brought OCSP revocation checks (merged in crl_ocsp branch), see also #254. I added a host header to the query but still the feature doesn't seem to be production ready yet :-(

The first 1169  nodes (Alexa ranking) with TLS replies on port 443 TCP:

-  336 still return ``error querying OCSP responder`` (305 with Code=400, 31 without any code returned). 
- 152 of 336 are showing a Google responder (pki.goog)
- 833 are fine.

This issue is just informational.

dcooper16:
Hi Dirk,

Thanks for the improvements that you made to the code that I submitted. I thought it was in fairly good shape, but that it could use some improvements, which is why I submitted it for the crl_ocsp branch rather than the 2.9dev branch. I didn't test as many sites as you did, but it seemed to at least pass the initial test of not reporting any incorrect results. It seemed that the code would either report the correct result (revoked or not revoked) or it wouldn't print anything at all.

My testing produced mixed results. I tested the code against several sites using parallel mass testing, and then ran individual tests in debug mode for any server that didn't print a result during mass testing. It seemed that in every case, however, that the individual test resulted in a response of "not revoked."

To some degree we are going to have to accept less than perfection on these tests. There is a reason that browsers have basically given up on using CRLs or OCSP to check server certificates, and that even when they did try to check CRLs or OCSP, the browsers would be configured to accept a certificate as valid if no revocation information could be obtained. It is my understanding that the sources of revocation information just weren't reliable enough to treat the inability to obtain revocation information as a hard failure.

drwetter:
I would at least to understand the 400s which is probably the HTTP server status. The logic that it's "our problem" seems to me too far away yet. Why would a certificate issue operate a server which is broken and put this URI into the certificates.

> Thanks for the improvements that you made to the code that I submitted. I thought it was in fairly good shape, but that it could use some improvements, which is why I submitted it for the crl_ocsp branch rather than the 2.9dev branch. 

In fact I did realize this (with a smile) and I appreciate you seem always to pay a lot of attention what you're doing :+1: 

> My testing produced mixed results. I tested the code against several sites using parallel mass testing, and then ran individual tests in debug mode for any server that didn't print a result during mass testing. It seemed that in every case, however, that the individual test resulted in a response of "not revoked."

I need to pass to you tomorrow the servers I tested. Two servers I remembered off the top of my head: ca.gov, sourceforge.net .

[...]

PR #1067, opened 2018-05-29
Fix false "revoked" results for CRL and OCSP checking

dcooper16:
This PR fixes problems with `check_revocation_crl()` sometimes reporting that a certificate is revoked even when it isn't, and with `check_revocation_ocsp()` sometimes reporting "error querying OCSP responder" even if the OCSP responder provided a good response. The most common reason for this to happen is when OpenSSL cannot validate the server's certificate (even without status checking). PR #1051 attempted to get status checking to work even in cases in which the server's certificate could not be validated. Based on some feedback provided, this PR instead addresses the problem by not checking status if `determine_trust()` was unable to validate the server's certificate.

In some cases the server's certificate can be validated using some, but not all of the bundles of trusted certificates. For example, I have encountered some sites that can be validated using the Microsoft and Apple bundles, but not the Linux or Mozilla bundles. So, this PR introduces `GOOD_CA_BUNDLE` to store a bundle that could be used to successfully validate the server's certificate. If there is no such bundle, then neither `check_revocation_crl()` nor `check_revocation_ocsp()` is run. When `check_revocation_crl()` and `check_revocation_ocsp()` are called, the status checks within them closely match the validation check in `determine_trust()`, which helps to ensure that if the check fails it is because of the status information.

As noted in #1057, at least one CA provides incorrect information when the CRL is downloaded, so validation could fail for a reason other than the certificate being revoked. So, this PR adds a check of the reason that validation failed and only reports "revoked" if the validation failed for that reason.

As noted in #1056, it is not possible to perform an OCSP query without access to the certificate issuer's public key. So, with this PR `check_revocation_ocsp()` is only called if the server's provided at least one intermediate certificate (i.e., the issuer's certificate, which contains the issuer's public key).

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1456,6 +1456,7 @@ check_revocation_crl() {
      local -i success
 
      "$PHONE_OUT" || return 0
+     [[ -n "$GOOD_CA_BUNDLE" ]] || return 0
      scheme="$(tolower "${crl%%://*}")"
      # The code for obtaining CRLs only supports LDAP, HTTP, and HTTPS URLs.
      [[ "$scheme" == "http" ]] || [[ "$scheme" == "https" ]] || [[ "$scheme" == "ldap" ]] || return 0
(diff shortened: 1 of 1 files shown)


drwetter:
Hi David,

thanks! (Also for your patience). Let's try this, I'll start a broader test and my hope is that you're right :-) Originally I didn't plan to do any further checking if the certificate chain has a problem but maybe it helps to solve some edge cases. 

I'll get back on this. Thx again & cheers, Dirk

Issue #93, opened 2015-05-05
Insecure Client-Initiated Renegotiation


drwetter:

drwetter:
This can only be achieved with an old binary (=<0.98k) or bash sockets

Phil2407:
Is there an update on this issue? In #1070 I saw you were planning to implement a test for CVE-2009-3555 via bash sockets. Is this still something that's being worked on, or is it already implemented?

drwetter:
Nope / nope :-)

PR #1070, opened 2018-06-14
Avoid false-positives when testing for CVE-2011-1473 / maximum of 3 renegotiation attempts

poupas:
Hi,

testssl.sh may report that a server is vulnerable to CVE-2011-1473 (possible DoS due to client-side renegotiation) even if it only allows a limited number of renegotiation attempts. It concludes that after attempting only one renegotiation.

Contrary to other implementations, Node.js and apparently lighttpd, mitigate the issue by allowing a maximum of 3 renegotiation attempts. See [here](https://github.com/nodejs/node-v0.x-archive/issues/2726) for more details.

We propose this patch to handle this scenario. The limit of renegotiation attempts (4) was obtained empirically according to Node.js's behaviour.

João

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -12803,8 +12803,10 @@ run_renego() {
                fileout "$jsonID" "OK" "likely not vulnerable (timed out)" "$cve" "$cwe"
                sec_client_renego=1
           else
-               # second try in the foreground as we are sure now it won't hang
-               echo R | $OPENSSL s_client $(s_client_options "$proto $legacycmd $STARTTLS $BUGS -msg -connect $NODEIP:$PORT $SNI $PROXY") >$TMPFILE 2>>$ERRFILE
+               # Second try in the foreground as we are sure now it won't hang.
+               # Probe for a max renegotiation limit to avoid false positives. Some servers (e.g. Node.js)
+               # mitigate the DoS threat by restricting the number of allowed client renegotiations.
+               (for i in $(seq 4); do echo R; sleep 1; done) | $OPENSSL s_client $(s_client_options "$proto $legacycmd $STARTTLS $BUGS -msg -connect $NODEIP:$PORT $SNI $PROXY") >$TMPFILE 2>>$ERRFILE
                sec_client_renego=$?                                                  # 0=client is renegotiating & doesn't return an error --> vuln!
                case "$sec_client_renego" in
                     0)   if [[ $SERVICE == "HTTP" ]]; then


drwetter:
Thank @poupas for your PR and sorry for my delayed reply.

Can I check this against any recent node / lighty server? I am just hesitant merging it without testing it myself.

Jac2NL:
The PR did not work for me against a Node.js 6.12.2 based application (Kibana 6.2). A manual test shows that Node returns a 400 Bad Request after a few times. So it looks like they rate limit the number of renegotiations and are not vulnerable to the DoS. Of course one could argue what would be an acceptable limit.

The patch below attempts to detect this behaviour:

https://gist.github.com/Jac2NL/7b63a6d6bbf98a5b87bdc0cd066d1ece#file-badrequest-patch

poupas:
Hi @drwetter. No problem :)

I have tested it, but you can test it yourself, with the following script:

```bash
#!/bin/bash -e
docker > /dev/null 2>&1
if [ $? != 0 ]; then
        echo "Please install Docker first"
        exit 1
fi

cat > Dockerfile << 'EOF'
FROM node:latest

ENV HTTP_SERVER_VERSION 0.11.1
ENV PUBLIC_FOLDER /opt/www

RUN mkdir -p $PUBLIC_FOLDER && \
        npm install -g http-server@0.11.1

CMD http-server $PUBLIC_FOLDER

EXPOSE 8080
EOF

mkdir -p tls
openssl genrsa -out tls/key.pem 2048
openssl req -new -x509 -key tls/key.pem -out tls/cert.pem -days 365 \
        -subj '/CN=localhost'

docker build . -t local/nodejs-https-test:latest
docker run --rm -ti -v $(pwd)/tls:/tls -p 8080:8080 local/nodejs-https-test:latest \
        http-server -S -C /tls/cert.pem -K /tls/key.pem
```

Regarding what @Jac2NL mentioned, I could not reproduce it. Maybe it is a Kibana-specific behavior?

Jac2NL:
Could be something Kibana specific. What version of Node are you using?

Come to think of it, a potential solution should not just check for 400 Bad Request but for anything which breaks off the connection, which is what we want.

Having said that, perhaps even a rate limited server may be vulnerable to a DoS because the client can simply reconnect and have another go at it with more renegotiations?

poupas:
Hi @Jac2NL ,

At the time of this writing, I tested against v10.5.0 (provided by the node:latest Docker tag). If you have a Docker installation available, feel free to play around with the shell script above.

Regarding the potential for DoS, I agree with you. Nonetheless, some researchers claim that allowing indefinite renegotiations through a single TCP connection is worse that forcing multiple TCP connections, because this may sidestep rate-limiting countermeasures (e.g., on firewalls, IDS, etc.).

More info about this [here](https://blog.qualys.com/ssllabs/2011/10/31/tls-renegotiation-and-denial-of-service-attacks).

Could you please send more detailed logs of the Kibana behaviour? Or Kibana deployment instructions?

Thanks

Jac2NL:
Kibana can be downloaded from www.elastic.co. They also have Docker images at www.docker.elastic.co.

According to the Kibana logs, Node's rate limiter kicks in, so it's probably not Kibana specific after all. It looks like CLIENT_RENEG_LIMIT has been set to 3:

{"type":"error","@timestamp":"2018-06-25T11:46:17Z","tags":["connection","client","error"],"pid":877,"level":"error","error":{"message":"socket hang up","name":"Error","stack":"Error: socket hang up\n    at TLSSocket.<anonymous> (_tls_wrap.js:847:25)\n    at emitOne (events.js:101:20)\n    at TLSSocket.emit (events.js:188:7)\n    at _handle.close (net.js:497:12)\n    at Socket.done (_tls_wrap.js:332:7)\n    at Socket.g (events.js:292:16)\n    at emitOne (events.js:101:20)\n    at Socket.emit (events.js:188:7)\n    at TCP._handle.close [as _onclose] (net.js:497:12)","code":"ECONNRESET"},"message":"socket hang up"}
{"type":"error","@timestamp":"2018-06-25T11:46:17Z","tags":["connection","client","error"],"pid":877,"level":"error","error":{"message":"140077570651968:error:1408A0C1:SSL routines:ssl3_get_client_hello:no shared cipher:../deps/openssl/openssl/ssl/s3_srvr.c:1427:\n","name":"Error","stack":"Error: 140077570651968:error:1408A0C1:SSL routines:ssl3_get_client_hello:no shared cipher:../deps/openssl/openssl/ssl/s3_srvr.c:1427:\n\n    at Error (native)"},"message":"140077570651968:error:1408A0C1:SSL routines:ssl3_get_client_hello:no shared cipher:../deps/openssl/openssl/ssl/s3_srvr.c:1427:\n"}
{"type":"error","@timestamp":"2018-06-25T11:46:17Z","tags":["connection","client","error"],"pid":877,"level":"error","error":{"message":"140077570651968:error:1408A0C1:SSL routines:ssl3_get_client_hello:no shared cipher:../deps/openssl/openssl/ssl/s3_srvr.c:1427:\n","name":"Error","stack":"Error: 140077570651968:error:1408A0C1:SSL routines:ssl3_get_client_hello:no shared cipher:../deps/openssl/openssl/ssl/s3_srvr.c:1427:\n\n    at Error (native)"},"message":"140077570651968:error:1408A0C1:SSL routines:ssl3_get_client_hello:no shared cipher:../deps/openssl/openssl/ssl/s3_srvr.c:1427:\n"}
{"type":"response","@timestamp":"2018-06-25T11:46:18Z","tags":[],"pid":877,"method":"get","statusCode":302,"req":{"url":"/","method":"get","headers":{"host":"localhost","user-agent":"TLS tester from https://testssl.sh/dev/","accept-encoding":"identity","accept":"text/*","connection":"Close"},"remoteAddress":"127.0.0.1","userAgent":"127.0.0.1"},"res":{"statusCode":302,"responseTime":31,"contentLength":9},"message":"GET / 302 31ms - 9.0B"}
{"type":"error","@timestamp":"2018-06-25T11:46:23Z","tags":["connection","client","error"],"pid":877,"level":"error","error":{"message":"TLS session renegotiation attack detected","name":"Error","stack":"Error: TLS session renegotiation attack detected\n    at Immediate.<anonymous> (_tls_wrap.js:42:17)\n    at runCallback (timers.js:672:20)\n    at tryOnImmediate (timers.js:645:5)\n    at processImmediate [as _immediateCallback] (timers.js:617:5)"},"message":"TLS session renegotiation attack detected"}

Jac2NL:
By the way, SSLLabs distinguishes between  Secure Client-Initiated Renegotiation and  Insecure Client-Initiated Renegotiation.

I don't know how these checks have been implemented and how they could be mapped on testssl.sh.

Perhaps Secure Client-Initiated Renegotiation is "Yes" when the string "Secure Renegotiation IS supported" is present and the server disconnects at renegotiation? And Insecure Client-Initiated Renegotiation is "Yes INSECURE" if it does not disconnect?

drwetter:
> By the way, SSLLabs distinguishes between Secure Client-Initiated Renegotiation and Insecure Client-Initiated Renegotiation.

testssl.sh does to. It just hasn't implemented a check yet for the latter, see #93 and ``testssl.sh(1)``. This can only be done with a really old OpenSSH version or -- this is how it is planned -- via sockets.

poupas:
> Perhaps Secure Client-Initiated Renegotiation is "Yes" when the string "Secure Renegotiation IS supported" is present and the server disconnects at renegotiation? And Insecure Client-Initiated Renegotiation is "Yes INSECURE" if it does not disconnect?

Not quite. There are actually two related, but distinct, vulnerabilities. One is a plaintext injection attack, the other is a DoS, which affects Secure Client-Initiated Renegotiation.

- *CVE-2009-3555*: _allows a man-in-the-middle attackers to insert data into HTTPS sessions, and possibly other types of sessions protected by TLS or SSL, by sending an unauthenticated request that is processed retroactively by a server in a post-renegotiation context, related to a "plaintext injection" attack_. This issue is fixed by implementing RFC5746.

- *CVE-2011-1473*: this CVE refers to a DoS affecting servers that already implement *Secure Client-Initiated Renegotiation*. If a server implements RFC5746, but does not rate-limit the number of attempts, it will be vulnerable.

Thanks for the logs. When I have little time available, I'll try to check why Kibana sends a 400 error, instead of simply shutting down the connection, like everything else seems to do (nginx, apache, AWS load balancers, among others...).

drwetter:
The 400 might be difficult as an indicator. Other servers might return a 400 too.

Other than that and @poupas : Did you try lighty -- and maybe other servers?

poupas:
I've used the following [script](https://gist.github.com/poupas/11ef5334c5b5fdbe09a766dbc9832974) to start an instance of lighttpd.

At least with this version (latest stable version), I cannot reproduce the "allow multiple renegotiations" behavior. Either client-initiated renegotiating is disabled completely, or if specified in the config file, it imposes no limit on renegotiation.

At this time, only Nodejs seems to follow this behavior.

exploide:
> The limit of renegotiation attempts (4) was obtained empirically according to Node.js's behaviour.

Let's start discussing if this number needs an adjustment :D It's probably completely arbitrary what in the end gets chosen here.

But currently I see an Apache, marked as vulnerable by testssl while it aborts renegotiation on the 6th attempt, leaving openssl with `140003409954624:error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure:ssl/record/rec_layer_s3.c:1407:SSL alert number 40`.

Therefore, I propose increasing the number of renegotiation attempts to at least 6.

(Let's see if someone comes next telling there is a server accepting up to 7 :stuck_out_tongue: )

drwetter:
Best would be to make to configurable via environment variable.

But a proper default needs to be there. 

Res would be great which lists for which implementation how many connects would be necessary.
-- 

drwetter:
Res --> research
-- 
Sent from my mobile. Excuse the autocorrection, my brevity&typos

drwetter:
Sorry about the delayed response, I've actually forgot to look __into__ the issue.

AFAIR to me it seems that Apache never had this issue (probably after patching). I've tested also now and can't spot it, even not on rotten old machines (Ubuntu 12.04). Lighty: Using your dockerfile now and as you said before it works fine.

What I still see is that node really stops at the fourth time. However I can DoS that thing, see here

Screenshot_20200114_194217

I can't e.g. with lighty or Apache or nginx.

drwetter:
Next morning: despite the load it is still responsive. So maybe that is because of the threaD model of node.js.

Thanks!

drwetter:
I took this and merged it into the current branch. Thanks again for the patch and the discussion

PR #1074, opened 2018-06-21
Fix #615

### LINKED ISSUE #615: LUCKY13: 128 cipher limit
There are more than 128 CBC ciphers in this test. Thus it has to be split to not send more than this as some servers cannot cope with this

dcooper16:
This PR fixes #615 for the case in which `tls_sockets()` is used by splitting the list of CBC ciphers into two lists, each with fewer than 128 ciphers and then testing each list separately.

For the `--ssl-native case`, no changes were needed. Even though `$cbc_ciphers` contains 154 ciphers, no version of OpenSSL supports all of these ciphers, and so the actual ClientHello sent by every version of OpenSSL contains fewer than 128 ciphers.

I did, however, add the `-no_ssl2` flag to the `$OPENSSL s_client` command to prevent OpenSSL from sending an SSLv2-compatible ClientHello. As is noted in a comment in run_server_preference(), "the supplied openssl will send an SSLv2 ClientHello if $SNI is empty and the -no_ssl2 isn't provided."


drwetter:
Thanks, David!

> For the ``--ssl-native case``, no changes were needed. Even though ``$cbc_ciphers`` contains 154 ciphers, no version of OpenSSL supports all of these ciphers, and so the actual ClientHello sent by every version of OpenSSL contains fewer than 128 ciphers.

I am counting for the current version provided exactly 128 ciphers -- including SSLv2 ciphers. So: yes,. that should be fine.

dcooper16:
> I am counting for the current version provided exactly 128 ciphers -- including SSLv2 ciphers.

I get a much smaller number.
```
./bin/openssl.Linux.x86_64.krb5 ciphers -V "$cbc_ciphers" | wc
```
gives me 133, and
```
./bin/openssl.Linux.x86_64.krb5 ciphers -V "$cbc_ciphers" | grep -v SSLv2 | wc
```
gives me 132.

However, if I run
```
./bin/openssl.Linux.x86_64.krb5 s_client -connect 127.0.0.1:443 -cipher "$cbc_ciphers" -no_ssl2
```
and then look at the ClientHello using either Wireshark or ssldump, I see that the ClientHello only includes 111 cipher suites (including 0x00ff). (Without the "-no_ssl2" option, an SSLv2-style ClientHello is sent that only includes 81 cipher suites.)

When I compared the two lists, I saw that the 22 missing ciphers were: (1) all 10 KRB5 ciphers; (2) all 9 SRP ciphers; and (3) the 3 PSK ciphers that do not also use an RSA key.

dcooper16:
In the one server I tested that had the 128-cipher limit bug, the 0x00ff counted as part of the limit. So, if we make a call of the form
```
tls_sockets "03" "$ciphers_to_test, 00,ff"
```
we need to ensure that `$ciphers_to_test` includes at most 127 cipher suites.

drwetter:
> I get a much smaller number.

I did before a baseline check with ``./testssl.sh -V | grep -c CBC`` which gave 128. Minus 5 x SSLv2 CBC ciphers = 123.

That didn't include the kerberos binary though. That would give 138 - 5 = 133. That is approximately the same as you had. It's still one off though.

(the difference in you counting from 133 to 132 comes from EXP-RC2-CBC-MD5 which is both a SSLv2 and >=SSLv3 cipher).

> and then look at the ClientHello using either Wireshark or ssldump, I see that the ClientHello only includes 111 cipher suites

Confirmed (used -trace option of openssl as it is easier for the first mentioned case). But how come only 111 were sent and not 133 / 132, i.e. the 22 ciphers you mentioned? PSK+SRP: Ok, maybe that would require a key piggyback to the s_client command? If that's the case we need to have a look at other places too.

Wrt: Kerberos there's also something odd. ``/bin/openssl.Linux.x86_64.krb5 s_server -cipher KRB5-IDEA-CBC-SHA:KRB5-IDEA-CBC-MD5:KRB5-RC4-SHA:KRB5-RC4-MD5:KRB5-DES-CBC3-SHA:KRB5-DES-CBC3-MD5 -key /tmp/server.pem -www -cert /tmp/server.crt`` plus either ``testssl.sh localhost:4433`` or ``/bin/openssl.Linux.x86_64.krb5 s_client -connect ...`` don't get me anywhere.

Hmm...

PR #1075, opened 2018-06-24
Add option to skip offensive vulnerability tests

Jac2NL:
Some vulnerability tests such as Heartbleed may be flagged as offensive by an IDS. To skip such tests I propose an environment variable OFFENSIVE which can be set to false, if the user so desires.

I have not fully researched which tests are actually offensive (e.g. those injecting unusual payloads) and which tests mimic fairly normal user behaviour (e.g. those using RC4). But you probably get the idea.

Of course some people may consider any type of reconnaissance (with or without testssl.sh) as offensive... :-)

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -220,6 +220,7 @@ APPEND=${APPEND:-false}                 # append to csv/json file instead of ove
 [[ -z "$NODNS" ]] && declare NODNS      # If unset it does all DNS lookups per default. "min" only for hosts or "none" at all
 HAS_IPv6=${HAS_IPv6:-false}             # if you have OpenSSL with IPv6 support AND IPv6 networking set it to yes
 ALL_CLIENTS=${ALL_CLIENTS:-false}       # do you want to run all client simulation form all clients supplied by SSLlabs?
+OFFENSIVE=${OFFENSIVE:-true}            # do you want to include offensive vulnerability tests?
 
 ########### Tuning vars which cannot be set by a cmd line switch. Use instead e.g "HEADER_MAXSLEEP=10 ./testssl.sh <your_args_here>"
 #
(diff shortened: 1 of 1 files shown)


Jac2NL:
The VULN_COUNT=16 may need some adjustment. I'm not quite sure what it's counting.

drwetter:
Idea seems good to me, you PR may need some tuning though. Let me see...

drwetter:
The '16' is just for formatting the headlines 

```
VULN_THRESHLD=${VULN_THRESHLD:-1}       # if vulnerabilities to check >$VULN_THRESHLD we DON'T show a separate header line in the output each vuln. check
```

and according to my counting 16 is correct

Jac2NL:
I meant, you may want to update the VULN_COUNT, should you decide to apply the PR :-)
It seems to me that the number of tests will decrease once you set the environment variable to false.
Anyway, that's just formatting, as you said.

drwetter:
Ah, got ya
-- 

drwetter:
Thank you!

I liked the idea and even spent a switch (```--ids-friendly``) for this. However I had a difference stance wrt what is offensive. So currenlty heartbleed, CCS injection, ticketbleed and ROBOT are offensice tests. The remaining 12 tests are less offensive. If those fail I would assume that other parts of a default run will cause problems, too.

Jac2NL:
Actually, at one stage I picked the same set of tests that you did. But then I reversed because I wasn't quite sure. One more thing: do you think that the GREASE tests could be detected as offensive?

By the way, it's probably not that you would get blocked by an IDS. More likely your actions get reported and you may have to explain yourself... :-)

drwetter:
>  do you think that the GREASE tests could be detected as offensive?

The GREASE tests checks robustness of the server's TLS implementation and it's somehow similar to wha more or less strange clients send anyway. thus it shouldn't match typical IDS signatures. But IDS can do a lot of things, it just depends on the configuration. I think if it triggers with testssl.sh running GREASE tests the OPS people have a lot of work every day.

> By the way, it's probably not that you would get blocked by an IDS. More likely your actions get reported and you may have to explain yourself... :-)

maybe I should have rather said IPS.  Once a while that happens.

To explain "my"self: yeah could be. Depends on a lot of factors. But other than scanning activities from [privacyscore](https://github.com/PrivacyScore/PrivacyScore) I never heard that somebody got a complaint.

Jac2NL:
Mind you, I wasn't referring to scanning the public internet and then getting detected by some random IDS/IPS.

But rather to people scanning their own internal networks and then getting a visit or report from their organisation's security team...

PR #1078, opened 2018-06-28
Check stapled OCSP response for revocation status

dcooper16:
In cases in which the server offers a stapled OCSP response, this PR extracts the OCSP response and then checks the response for the status of the server's certificate. The check is performed in the same way as when the certificate includes an OCSP URI and the `--phone-out` option is set, except that the OCSP response is received from the TLS server rather than coming directly from the OCSP responder. Since this only involves additional processing of data that testssl.sh is already receiving, the check is performed whether or not the `--phone-out` flag is set.


drwetter:
looks good, thanks!

Never heard about the respin option before and couldn't find anything in the OpenSSL ChangeLog. Seems to be already supported by a 0.98k binary I have here.

PR #1079, opened 2018-06-28
Handle incorrectly populated certificate_list

dcooper16:
According to Section 7.4.2 of RFC 5246, when a server sends its certificate it MUST send a list in which the first certificate is the sender's certificate and "Each following certificate MUST directly certify the one preceding it." testssl.sh currently assumes that the server has populated the list this way and so places the second certificate in the list into `$TEMPDIR/hostcert_issuer.pem`.

However, not all servers have been following this requirement, and so draft-ietf-tls-tls13 (soon to be RFC 8446) only says that servers SHOULD list the certificates in this way and says that clients "SHOULD be prepared to handle potentially extraneous certificates and arbitrary orderings from any TLS version, with the exception of the end-entity certificate which MUST be first."

testssl.sh needs to place the correct certificate in `$TEMPDIR/hostcert_issuer.pem`, since otherwise any OCSP request it sends will be incorrect, and any attempt to verify and OCSP response will be incorrect as well.

This PR changes `extract_certificates()` and `parse_tls_serverhello()` to populate `$TEMPDIR/hostcert_issuer.pem` with the first certificate in certificate_list that has a subject DN that matches the issuer DN in the server's certificate, rather than simply populating $TEMPDIR/hostcert_issuer.pem with the second certificate in the list.

In testing a random sampling of U.S. government servers, of 57 servers tested 5 reported "unauthorized" for the OCSP URI using the current testssl.sh and all 5 of these reported "not revoked" with this PR. This PR also corrects the same issue in some servers on the Alexa Top 1000, but this was only a problem for 12 of those 1000 servers.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6555,7 +6555,8 @@ determine_tls_extensions() {
 extract_certificates() {
      local version="$1"
      local savedir
-     local -i success nrsaved=0
+     local -i i success nrsaved=0
+     local issuerDN CAsubjectDN
 
      # Place the server's certificate in $HOSTCERT and any intermediate
      # certificates that were provided in $TEMPDIR/intermediatecerts.pem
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks!

PR #1091, opened 2018-07-23
Update etc/common-primes.txt

dcooper16:
The primes in etc/common-primes.txt that were taken from https://github.com/cryptosense/diffie-hellman-groups/blob/master/gen/common.json were encoded in decimal rather than hexadecimal, preventing them from being matched against the primes offered by servers. This PR converts the primes from https://github.com/cryptosense/diffie-hellman-groups/blob/master/gen/common.json to hexadecimal, removing those that were duplicates from https://svn.nmap.org/nmap/scripts/ssl-dh-params.nse.


drwetter:
Thanks David for catching this.  I am not sure about are some primes like Sun JDK, Nginx, Apache, MODP groups.

PR #1110, opened 2018-08-28
Fix #1080

### LINKED ISSUE #1080: Check for incorrect certificate ordering
As noted in #1079, some servers are sending incorrectly formatted lists of certificates.  draft-ietf-tls-tls13 (soon to be RFC 8446) says:

> The sender's certificate MUST come in the first CertificateEntry in the list.  Each following certificate SHOULD directly certify the one immediately preceding it.

In RFC 5246, the "SHOULD" in this text was a "MUST".

testssl.sh should add a check that each certificate in the list of certificates sent by the server directly certifies the one immediately preceding it.

The change would be similar to the one in #1079, except that the issuer and subject DNs would be extracted from **every** certificate in the list and a check would be done to verify that the subject DN in each certificate was the same as the issuer DN in the preceding certificate. A low-level warning could then be issued if a problem was found in the list.

dcooper16:
RFC 8446 specifies the following for the list of certificates provided by the server:

> The sender's certificate MUST come in the first CertificateEntry in the list.  Each following certificate SHOULD directly certify the one immediately preceding it.

In RFC 5246 the "SHOULD" was a "MUST". This PR adds a check of whether the certificates provided by the server are in the correct order and issues a low severity warning if they are not.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -251,6 +251,8 @@ NO_ENGINE=${NO_ENGINE:-false}           # if there are problems finding the (ext
 declare -r CLIENT_MIN_PFS=5             # number of ciphers needed to run a test for PFS
 CAPATH="${CAPATH:-/etc/ssl/certs/}"     # Does nothing yet (FC has only a CA bundle per default, ==> openssl version -d)
 GOOD_CA_BUNDLE=""                       # A bundle of CA certificates that can be used to validate the server's certificate
+CERTIFICATE_LIST_ORDERING_PROBLEM=false # Set to true if server sends a certificate list that contains a certificate
+                                        # that does not certify the one immediately preceding it. (See RFC 8446, Section 4.4.2)
 STAPLED_OCSP_RESPONSE=""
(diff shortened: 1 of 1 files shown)


drwetter:
Cool, thanks

MindTooth:
Did this get removed?  I find this very useful information.

drwetter:
@MindTooth : ?

MindTooth:
I'm sorry, I maybe came off as rude.

I could not find `CERTIFICATE_LIST_ORDERING_PROBLEM` in the source using the [search](https://github.com/search?q=repo%3Adrwetter%2Ftestssl.sh%20CERTIFICATE_LIST_ORDERING_PROBLEM&type=code).  However, I see now that it's still present.  And I see the "Certificates provided" line in the output.

I also noticed that I was running an old version, so I submitted a patch to Homebrew pointing `HEAD` to `3.2`.

I'll open a new issue if I find a bug.  Thanks anyway. 😄

drwetter:
No worries, that didn't come across as rude. Just wasn't getting your point :-)

PR #1129, opened 2018-09-21
Fix #1118

### LINKED ISSUE #1118: Detect and fix TLS 1.3 handshake in sclient_connect_successful()
`` sclient_connect_successful()`` is for SSL 3 to TLS 1.2.  There's no Master Key in TLS 1.3. Also the logic for TLS 1.3 handshake is different. As a result:

```
./testssl.sh --ip=one -P --openssl /data/tmp/openssl.git/apps/openssl testssl.net

No engine or GOST support via engine with your <redacted>openssl
    testssl.sh       3.0rc1 from https://testssl.sh/dev/
    (5de89ae 2018-09-05 16:48:28 -- )

      This program is free software. Distribution and

 Using "OpenSSL 1.1.1-pre10-dev  xx XXX xxxx" [~189 ciphers]
 on <redacted>
 (built: "Aug 21 14:57:12 2018", platform: "linux-x86_64")

 Start 2018-09-10 07:54:43        -->> 104.18.34.81:443 (testssl.net) <<--

 Further IP addresses:   104.18.35.81 2400:cb00:2048:1::6812:2251 2400:cb00:2048:1::6812:2351 
 A record via:           supplied IP "104.18.34.81"
 rDNS (104.18.34.81):    --
 Service detected:       HTTP


 Testing server preferences 

 Has server cipher order?     yes (OK)
 Negotiated protocol          default proto empty
 Negotiated cipher            default cipher empty, 253 bit ECDH (X25519)
 Cipher order
    TLSv1:     ECDHE-ECDSA-AES128-SHA ECDHE-ECDSA-AES256-SHA 
    TLSv1.1:   ECDHE-ECDSA-AES128-SHA ECDHE-ECDSA-AES256-SHA 
    TLSv1.2:   ECDHE-ECDSA-CHACHA20-POLY1305-OLD ECDHE-ECDSA-CHACHA20-POLY1305 ECDHE-ECDSA-AES128-GCM-SHA256 ECDHE-ECDSA-AES128-SHA ECDHE-ECDSA-AES128-SHA256 ECDHE-ECDSA-AES256-GCM-SHA384
               ECDHE-ECDSA-AES256-SHA ECDHE-ECDSA-AES256-SHA384 
    TLSv1.3:   TLS_AES_256_GCM_SHA384 TLS_CHACHA20_POLY1305_SHA256 TLS_AES_128_GCM_SHA256 


 Done 2018-09-10 07:54:50 [  19s] -->> 104.18.34.81:443 (testssl.net) <<--
```

 See ``default proto empty`` and ``default cipher empty``.  

This is a leftover of #1115 .

dcooper16:
This PR is an attempt to fix the problem identified in #1118.

Currently, `get_cipher()` and `get_protocol()` attempt the extract the cipher and protocol from the SSL-Session information printed by OpenSSL s_client. This does not always work for TLSv1.3, however, since OpenSSL 1.1.1 will only print SSL-Session information for a TLSv1.3 connection if the server sends New Session Ticket. If the server doesn't, then `get_cipher()` and `get_protocol()` return empty strings.

For TLSv1.3 connections in which the server does not send a New Session Ticket, there seems to be only one other source for this information. A line of the form:
```
        New, TLSv1.3, Cipher is TLS_AES_256_GCM_SHA384
```
[Note that "New" would be "Reused" if the connection were created via session resumption.]

The use of this line seems to be okay for extracting the negotiated cipher, but it cannot be used in general to extract the negotiated protocol. The reason is that this line is created as follows:
```
    BIO_printf(bio, (SSL_session_reused(s) ? "---\nReused, " : "---\nNew, "));
    c = SSL_get_current_cipher(s);
    BIO_printf(bio, "%s, Cipher is %s\n",
               SSL_CIPHER_get_version(c), SSL_CIPHER_get_name(c));
```
While the cipher that is printed seems to be the negotiated cipher, the protocol that is printed is "the SSL/TLS protocol version that first defined the cipher." Since TLS 1.3 ciphers may only be used with TLS 1.3, protocol version printed on this line may be accepted as the negotiated protocol if and only if it is "TLSv1.3."

This PR addresses the problem by modifying `get_cipher()` and `get_protocol()` to check the "New, ..., Cipher is ..." line if lines from SSL-Session ("Cipher    : ...", "Protocol  : ...") cannot be found. In the case of get_protocol() the protocol on the "New, ..., Cipher is ..." will be accepted only if the protocol is "TLSv1.3" and the cipher is a TLSv1.3 cipher.

This PR also adds a check for the "New, ..., Cipher is ..." to `sclient_connect_successful()`. If this line is present, and the protocol and cipher are not "(NONE)", then this is accepted as an indication that the connection was successful even if the "Master-Key" line does not appear. It is not clear whether this extra test is needed, however, as `sclient_connect_successful()` will not even look at the text in the output of OpenSSL s_client if function's return value is 0, and OpenSSL s_client should return 0 if the connection was successful.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1225,13 +1225,36 @@ strip_trailing_space() {
 
 # retrieve cipher from ServerHello (via openssl)
 get_cipher() {
-     awk '/Cipher *:/ { a=$3 } END { print a }' "$1"
-     #awk '/\<Cipher\>/ && !/Cipher is/  && !/^New/ { print $3 }' "$1"
+     local cipher=""
+     local server_hello="$(cat "$1")"
+
+     if [[ "$server_hello" =~ Cipher\ *:\ ([A-Z0-9]+-[A-Z0-9\-]+|TLS_[A-Z0-9_]+) ]]; then
+          cipher="${BASH_REMATCH##* }"
+     elif [[ "$server_hello" =~ (New|Reused)", "(SSLv[23]|TLSv1(\.[0-3])?(\/SSLv3)?)", Cipher is "([A-Z0-9]+-[A-Z0-9\-]+|TLS_[A-Z0-9_]+) ]]; then
+          cipher="${BASH_REMATCH##* }"
+     fi
+     tm_out "$cipher"
 }
 
 # retrieve protocol from ServerHello (via openssl)
 get_protocol() {
-     awk '/Protocol *:/ { a=$3 } END { print a }' "$1"
+     local protocol=""
+     local server_hello="$(cat "$1")"
+
+     if [[ "$server_hello" =~ Protocol\ *:\ (SSLv[23]|TLSv1(\.[0-3])?) ]]; then
+          protocol="${BASH_REMATCH##* }"
+     elif [[ "$server_hello" =~ (New|Reused)", TLSv1.3, Cipher is "TLS_[A-Z0-9_]+ ]]; then
+          # Note: When OpenSSL prints "New, <protocol>, Cipher is <cipher>", <cipher> is the
+          # negotiated cipher, but <protocol> is not the negotiated protocol. Instead, it is
+          # the SSL/TLS protocol that first defined <cipher>. Since the ciphers that were
+          # first defined for TLSv1.3 may only be used with TLSv1.3, this line may be used
+          # to determine whether TLSv1.3 was negotiated, but if another protocol is specified
+          # on this line, then this line does not indicate the actual protocol negotiated. Also,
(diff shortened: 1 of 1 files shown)


drwetter:
Looks good, thanks! One minor thing to avoid cat : ``server_hello="$(< "$1")"`` should work too, right?

dcooper16:
> Looks good, thanks! One minor thing to avoid cat : server_hello="$(< "$1")" should work too, right?

I can try that. I just copied the `server_hello="$(cat "$2")"` from the first line of `sclient_connect_successful()`.

PR #1136, opened 2018-09-25
Revert #1113

dcooper16:
This PR removes the changes that were added in #1113. As noted in the conversation for #1113, it was eventually determined that the actual bug was related to the first message fragment being too short rather than the overall length of the message. So, the warning message that is displayed by the code is misleading. In addition, if changes are made to avoid TCP fragmentation, then this code will no longer even test for intolerance to short message fragments.

I have tried creating a test that intentionally sends the first part of the ClientHello as a short segment as a replacement for this test, but so far the code does not seem to produce consistent results.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -14532,12 +14532,12 @@ run_tls_truncation() {
 run_grease() {
      local -i success
      local bug_found=false
-     local normal_hello_ok=false clienthello_size_bug=false
+     local normal_hello_ok=false
      local cipher_list proto selected_cipher selected_cipher_hex="" extn rnd_bytes
      local alpn_proto alpn alpn_list_len_hex extn_len_hex
      local selected_alpn_protocol grease_selected_alpn_protocol
      local ciph list temp curve_found
-     local -i i j rnd alpn_list_len extn_len base_len clienthello_len debug_level=""
+     local -i i j rnd alpn_list_len extn_len debug_level=""
      local -i ret=0
      # Note: The following values were taken from https://datatracker.ietf.org/doc/draft-ietf-tls-grease.
      # These arrays may need to be updated if the values change in the final version of this document.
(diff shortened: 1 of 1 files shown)


drwetter:
> I have tried creating a test that intentionally sends the first part of the ClientHello as a short segment as a replacement for this test, but so far the code does not seem to produce consistent results.

Did you try separate bash internal printf function?

dcooper16:
I just posted the code that I've been working on to https://github.com/dcooper16/testssl.sh/tree/small_fragments. The code splits the ClientHello into two pieces, a small initial piece of 5, 6, 7, or 8 bytes, and then second piece with the remainder of the ClientHello, and it sends each piece with a separate call to printf. As far as I can tell from looking at Wireshark, the two pieces are being sent in separate fragments as expected.

The results, however, are not consistent. With the test in #1113, it seems that every server either fails every connection for which TLSCiphertext.length **mod** 256 = 10 or it succeeds on every connection. This is not what I am seeing, however, with https://github.com/dcooper16/testssl.sh/tree/small_fragments.

Servers that failed the test in #1113 seem to consistently fail the test in https://github.com/dcooper16/testssl.sh/tree/small_fragments as well. However, unlike with the test in #1113, there are some servers that will only intermittently fail if sent a ClientHello with in which the first TCP fragment is short. If you run the test in https://github.com/dcooper16/testssl.sh/tree/small_fragments with `--debug 2` on several servers that pass the test in #1113, you will see the occasional failed connection. Running the test on a server that passes #1113 twice will not necessarily produce the same results.

In rare cases a server that passed #1113 will fail on all eight tests in https://github.com/dcooper16/testssl.sh/tree/small_fragments and so the warning message will be printed: "Server fails if ClientHello sent with first message fragment less than 9 bytes."

Since the testing in https://github.com/dcooper16/testssl.sh/tree/small_fragments isn't producing consistent results, I don't think it is ready yet.

drwetter:
Okay, thanks for the heads up. Need to check that later.

Issue #1097, opened 2018-08-05
Name check fails for XMPP server


joto:
Please make sure that you provide enough information so that we understand what your issue is about.

1. testssl version from the banner (testssl.sh -b 2>/dev/null | head -4 | tail -2)

    testssl       2.8rc3 from https://testssl.sh/dev/
    (33dcb43 2018-08-04 19:09:51 -- 1.562)

Using "OpenSSL 1.1.0f  25 May 2017"

(From Debian stable)

2. what exactly was happening, output is needed

Tested an XMPP server with `testssl --quiet --server_defaults -t xmpp --xmpphost topf.org jabber.topf.org:5222`

So the server itself has the name `jabber.topf.org`, but the jabber ids would be `user@topf.org`. In this case, the certificate should use `topf.org` as SN and it does.

testssl now says:
```
...
Trust (hostname)             certificate does not match supplied URI
...
```

Maybe using the `--xmpphost` option was the wrong thing to do in this case? The docs are not 100% clear here.

3. what did you expect instead?

I would have expected it to say that everything is okay.

4. steps to reproduce

see above

dcooper16:
Hi @joto,

I can't access the server that you mentioned (probably a firewall issue), but it seems that you are correct. testssl.sh currently uses one set of matching rules for all types of servers - it checks for DNS names in the subjectAltName extension and it checks the common name attribute in the subject name.

According to Section 13.7.2.1 of [RFC 6120](https://tools.ietf.org/html/rfc6120#section-13.7.2.1), however, for XMPP servers there should also be a check against any XmppAddr names in the subjectAltName extension.

[RFC 6125](https://tools.ietf.org/html/rfc6125) also mentions checking for matching SRVName in the subjectAltName extension.

So, it seems that testssl.sh's rules for checking names in certificates are okay for HTTPS, and perhaps for some of the other non-HTTPS protocols it supports, but it isn't for XMPP and perhaps for some other protocols. To be complete, testssl.sh will need to add some protocol-specific matching rules.

I can try to help with this, but my ability may be limited, as I may be limited in what servers I can test against.

drwetter:
This seems loosely related to #316. Best would be if we think about protocol specific hooks in ``certificate_info()``

OT:

> testssl 2.8rc3 from https://testssl.sh/dev/
> (33dcb43 2018-08-04 19:09:51 -- 1.562)

[...]

Issue #316, opened 2016-03-09
Support SNI on mail tests


chrcoluk:
I am observing SNI seems to be non functional when using -t SMTP.

I see references to SNI in the code but the script is huge so I dont want to try and add the required arguments to the code.

My belief is because when I use the -S on its own just to see the certificate used, its not using the right certificate, yet my email client does use the right certificate.

Please update the script to utilise SNI on smtp and pop tests.

Thanks

chrcoluk:
I can confirm manual testing with the " openssl s_client -starttls smtp -servername " command shows SNI is functional.

drwetter:
SNI is used in HTTP.

For SMTP and SNI there's no such standard.  Preferred is to handle this via MX record. Not speaking about the fact that TLS encryption is a) opportunistic b) often cleints don't do strict certificate checking.

The only thing I know of is wrt to SMTP is TLSA records.

drwetter:
PS: What is POP ;-)

chrcoluk:
There is SNI in dovecot and exim, both very widely used server software.
Outlook, gmail, Thunderbird also all support SNI.

drwetter:
there's no SNI in sendmail or postfix, both widely used server software. So for an exim server it's BS to rely on SNI, unless that exim is sure it won't talk to sendmail or postfix.

drwetter:
@chrcoluk: I would need a detailed use case where testssl.sh currently fails and an explanation.

chrcoluk:
For reference I did disable SNI as I had issues with the android k9 client not supporting it, so now am using one cert per ip, however I still would like the test in place as SNI does exist for email, is on the software I mentioned and will only get more popular as time moves forward.
I have no issue if you disable it by default, but please add a flag thats optional to enable it.

afletch:
@drwetter SNI is now pretty widespread on mail (and even FTP), as services transition quickly to use free Let's Encrypt issued certificates. It would be great to see SNI support for these services in testssl.

drwetter:
If somebody feels doing so and I someone points me to test machines: Ok.

drwetter:
Just looking at the source codes from openssl 1.0.2 and openssl 1.1.1. Seems 1.1.1 has a ``-name`` switch for this? Looks to me like this would be also needed _currently_ for testssl.sh as there's are a few handshaskes left we do without sockets.

If anybody is willing to backport this please send a PR to https://github.com/drwetter/openssl-1.0.2.bad .

[...]

PR #1137, opened 2018-09-25
Name check for XMPP servers

dcooper16:
This PR is an attempt to address #1097. I have only been able to test it against jabber.topf.org and against locally created test certificates, so it needs more testing.

The main issue that this addresses is that testssl.sh currently checks against the wrong name for XMPP servers. According to RFC 6120, Section 13.7.2.1:

> - The initiating entity sets the source domain of its reference identifiers to the 'to' address it communicates in the initial stream header; i.e., this is the identity it expects the receiving entity to provide in a PKIX certificate.

So, if the `--xmpphost` option is provided, then the name provided in that option should be compared against the name in the certificate rather than the host name.

`compare_server_name_to_cert()` currently takes the server name to look for in the certificate as an parameter, but every call to `compare_server_name_to_cert()` uses `$NODE` as the argument. So, this PR removes the parameter and sets `$servername` to either `$XMPP_HOST` or `$NODE` as appropriate. This small change alone should fix the problem for most XMPP servers since the server's name SHOULD appear in the certificate encoded as a DNS name. That is the case for the one server I could test - jabber.topf.org.

The majority of the code in this PR is to address the other possibilities noted in RFC 6120, Section 13.7.1.2.1. This section notes that an XMPP server's identity may also appear in the subjectAltName extension as an otherName, either an SRV-ID or an XmppAddr identifier. Unfortunately, OpenSSL's certificate printer does not support otherName and just prints "othername:\<unsupported\>". So, this PR includes code to manually extract any SRV-ID or XmppAddr names from the certificate. This involves parsing the DER encoding of the certificate to look for the subjectAltName extension, looping through all of the names in the extension, and pulling out the names of these two name forms. This code is only run if the server is an XMPP server and the certificate does not have a matching DNS name. So, this code will rarely be executed.

This PR addresses one other issue. There is code in `certificate_info()` to set the variables `$has_dns_sans` and `$has_dns_sans_nosni.` These variables are needed to address the following requirement:

>      # Find out if the subjectAltName extension is present and contains
>      # a DNS name, since Section 6.3 of RFC 6125 says:
>      #      Security Warning: A client MUST NOT seek a match for a reference
>      #      identifier of CN-ID if the presented identifiers include a DNS-ID,
>      #      SRV-ID, URI-ID, or any application-specific identifier types
>      #      supported by the client.

While it is relatively easy to determine whether a certificate includes a DNS name in the subjectAltName extension, as noted above, it is not easy to determine whether it has an SRV-ID or an XmppAddr. So, this PR leverages the work `compare_server_name_to_cert()` does in parsing the subjectAltName extension by having `compare_server_name_to_cert()` set a global variable indicating whether the certificate has a subjectAltName extension with a relevant name form (DNS, SRV-ID, or XmppAddr for XMPP, or DNS for other servers). `$has_dns_sans` and `$has_dns_sans_nosni` are then just set to the value of this global variable.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -255,6 +255,7 @@ GOOD_CA_BUNDLE=""                       # A bundle of CA certificates that can b
 CERTIFICATE_LIST_ORDERING_PROBLEM=false # Set to true if server sends a certificate list that contains a certificate
                                         # that does not certify the one immediately preceding it. (See RFC 8446, Section 4.4.2)
 STAPLED_OCSP_RESPONSE=""
+HAS_DNS_SANS=false                      # Whether the certificate includes a subjectAltName extension with a DNS name or an application-specific identifier type.
 MEASURE_TIME_FILE=${MEASURE_TIME_FILE:-""}
 if [[ -n "$MEASURE_TIME_FILE" ]] && [[ -z "$MEASURE_TIME" ]]; then
      MEASURE_TIME=true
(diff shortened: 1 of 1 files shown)


drwetter:
Thx, David.

How would we deal with #316 in this context?

drwetter:
Thanks!

Issue #1139, opened 2018-10-04
Warn when DH parameters are less than 2048bit


fishouttah20:
1. testssl version from the banner (testssl.sh -b 2>/dev/null | head -4 | tail -2)
latest
2. what exactly was happening, output is needed
When doing a full scan or running just --logjam there is no warning that DH parameters are less than 2048 bit if they aren't using common primes. The only indication of this state is the DH 1024 (or similar) shown in orange in the per cipher table. 

3. what did you expect instead?
A warning that DH parameters were less than 2048bit. The logjam test seems the most relevant place for this information but ultimately a warning or even informational finding in the CSV/JSON/etc output would ensure we can scan for this without doing a complete scan and parsing individual ciphers.

I appreciate that some systems can't support 2048 DH and the original Logjam scope was Export ciphers and common primes, but 2048 DH is seen as a best practice and many security scanning tools report report it as a finding or penalise their grading when it's detected. (SSLLabs, etc).

drwetter:
FYI: See merges of David's PRs and commits to the ``dh_params`` branch. Still a little tweaking in the UI though which will happen in  ``dh_params``

drwetter:
See commit above

fishouttah20:
Thanks Dirk, much appreciated!

PR #1141, opened 2018-10-04
Send ECDHE quality to fileout()

dcooper16:
This PR is also based on #1139, but it addresses ECDH keys rather than DH keys. When `run_pfs()` prints the list of elliptic curves offered, it colors each curve according to its quality (based on key length). However, the severity level used when the list is sent to `fileout()` is always "INFO". This PR changes the call to `fileout()` to make the severity level be based on the quality of the shortest curve that the server offers.


drwetter:
Thanks!

PR #1142, opened 2018-10-05
Docker improvements

gandrade:
Adding **caching busting** technique, also adding `--no-cache` option which avoids using `apk update` instruction explicitly

References:

- https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run
- https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,21 @@
 FROM alpine:latest
 
-RUN apk update && apk upgrade
-RUN apk add bash procps drill git coreutils
-RUN apk add --no-cache curl
+WORKDIR /home/testssl/
+
+RUN apk add --no-cache \ 
+    bash \
+    procps \
+    drill \
+    git \
+    coreutils \ 
+    curl
 
 RUN addgroup testssl
-RUN adduser -G testssl -g "testssl user"  -s /bin/bash -D testssl
+RUN adduser -G testssl -g "testssl user" -s /bin/bash -D testssl
 
-RUN ln -s /home/testssl/testssl.sh /usr/local/bin/
+RUN ln -s testssl.sh /usr/local/bin/
 
 USER testssl
-WORKDIR /home/testssl/
 
 RUN git clone --depth=1 https://github.com/drwetter/testssl.sh.git .
 


drwetter:
Okay, thanks.

Two questions: the workdir is omitted. Any reason why you're doing do that?
Secondly:  ``apk upgrade`` would be missing. Reading the docker doc you pointed me to: How can I tell whether @ dockerhub the alpine image is up to date otherwise?

gandrade:
- Once `WORKDIR` instruction was moved to the top, there isn't any reason to replicate it during `ln -s` instruction. The Dockerfile behavior stays the same as before. Just leveraging it. (https://docs.docker.com/engine/reference/builder/#workdir) 
- According to Docker documentation: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run
> Avoid RUN apt-get upgrade and dist-upgrade, as many of the “essential” packages from the parent images cannot upgrade inside an unprivileged container. If a package contained in the parent image is out-of-date, contact its maintainers

drwetter:
> > According to Docker documentation: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run
> Avoid RUN apt-get upgrade and dist-upgrade, as many of the “essential” packages from the parent images cannot upgrade inside an unprivileged container. If a package contained in the parent image is out-of-date, contact its maintainers

Gustavo, you didn't answer my question. That is exactly what was referring to :-)

First: I can upgrade inside the image. Then I assume but I can't tell whether the image is up-to date. So also if I have some trust into dockerhub, in the end I don't know whether they are missing a crucial patch which might be relevant in the future. Also I even can't tell whether "parent image is out-of-date". So for me the ``upgrade`` statement makes sure that I only rely on Alpine doing their homework and not on Alpine and dockerhub.

I would be willing to loosen this if there's an security SLA which I feel comfortable with but I wasn't able to find it.

gandrade:
Yes, you can `upgrade` inside the image, however, following the [Docker Best Practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) documentation page, already mentioned, you shouldn't do it. If it makes sense for you, would be preferable, keep your base image up-to-date, instead of doing it over and over again. It's important to point out that the way Dockerfile was written, `apk update && apk upgrade`, followed by `apk add` it'll work for the first time. Let's say, a new package needs to be added, for instance, you might face build cache layer issues, once `apk update && apk upgrade` was already cached.
Another point, `upgrade` instruction could bring invisible side-effects hard to track it down. 
IMO, it's important to keep the notes above in mind.

Regarding security's SLA, there's nothing which the official documentation could argue here.

gandrade:
Based on what was discussed, how you want to proceed?

drwetter:
I need to do some research and this point has a lower priority

gandrade:
Thanks for your feedback and Let me know if you need something.

gandrade:
Do you have a chance to do some research on this topic? Just want to confirm if it's still a valid PR.

drwetter:
Hi Gustavo,

I didn't have much time to look into it yet. As indicated unfortunately there are currently more urgent requests waiting in here, besides my day-to-day work.

The PR is still valid. I would have closed it if that would be not the case.

Cheers, Dirk

gandrade:
Got it Dirk, thanks for the feedback.

Dockerfile
```diff
@@ -1,16 +1,21 @@
 FROM alpine:latest
 
-RUN apk update && apk upgrade
```
drwetter: This line is missing. Why not ``RUN apk update && apk upgrade && apk add --no-cache ...``?

gandrade: - `apk-update` instruction:
Once you're using the `--no-cache`  argument, it allows you to always get up-to-date packages; So, in that sense, the use of `update` instruction has no effects;

Reference: https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache

- `apk-upgrade` instruction:
If you need to upgrade a specific package for certain reasons, the recommended approach would be bumping `alpine` version for the one that provides you this updated package;
> Avoid RUN apt-get upgrade and dist-upgrade, as many of the “essential” packages from the parent images cannot upgrade inside an unprivileged container. If a package contained in the parent image is out-of-date, contact its maintainers. If you know there is a particular package, foo, that needs to be updated, use apt-get install -y foo to update automatically.

Reference: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

gandrade: Do you have a chance to take a look at it? Does it still valid to be open?

drwetter: Sorry for letting this hang here....

It was just a matter that I wanted to MAKE SURE that packages are up-to-date. It's just a security baseline, removing one instance in the chain we maybe are dependent on. I gave up on complaining about people doing their work properly, see e.g. a pending thing here https://github.com/alpinelinux/docker-alpine/issues/34 .

I just modified the line to ``RUN apk update && apk upgrade && apk add --no-cache curl bash procps drill git coreutils libidn curl``

```
(1/4) Upgrading musl (1.1.20-r4 -> 1.1.20-r5)
(2/4) Upgrading libcrypto1.1 (1.1.1b-r1 -> 1.1.1d-r0)
(3/4) Upgrading libssl1.1 (1.1.1b-r1 -> 1.1.1d-r0)
(4/4) Upgrading musl-utils (1.1.20-r4 -> 1.1.20-r5)
```

If I omit the statement ``apk upgrade`` those lines are missing.

So I will modify the line, boil down the RUN statements to one and close this.

Thanks for the fruitful discussion.

drwetter:
Sorry for letting this hang here....

It was just a matter that I wanted to MAKE SURE that packages are up-to-date. It's just a security baseline, removing one instance in the chain we maybe are dependent on. I gave up on complaining about people doing their work properly, see e.g. a pending thing here alpinelinux/docker-alpine#34 .

I just modified the line to RUN apk update && apk upgrade && apk add --no-cache curl bash procps drill git coreutils libidn curl

(1/4) Upgrading musl (1.1.20-r4 -> 1.1.20-r5)
(2/4) Upgrading libcrypto1.1 (1.1.1b-r1 -> 1.1.1d-r0)
(3/4) Upgrading libssl1.1 (1.1.1b-r1 -> 1.1.1d-r0)
(4/4) Upgrading musl-utils (1.1.20-r4 -> 1.1.20-r5)

If I omit the statement apk upgrade those lines are missing.

So I will modify the line, boil down the RUN statements to one and close this.

Thanks for the fruitful discussion.

PR #1147, opened 2018-10-23
Dev fixcsv

C0FFEEC0FFEE:
Fixes the CSV file output (there were two duplicate commas).

Example of erroneous output:
`"service","hostname.example.com/www.xxx.yyy.zzz,","443,","INFO","HTTP","",""`

Example of fixed output:
`"service","hostname.example.com/www.xxx.yyy.zzz","443","INFO","HTTP","",""`

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6697,7 +6697,7 @@ extract_stapled_ocsp() {
      local response="$(cat $TMPFILE)"
      local ocsp tmp
      local -i ocsp_len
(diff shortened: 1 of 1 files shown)


drwetter:
Thx for the PR! 

Could you remove the "fixes" for the trailing spaces in the EOF section pls? They break the output.

drwetter:
Thanks again.

PR #1151, opened 2018-10-31
Create testssl-nmap-auto.sh

B-kay:
automate nmap and testssl scans


drwetter:
Thx but I am afraid this won't make it

1) The code quality and sense is questionable
2) Have you looked into ~/utils/gmap2testssl.sh
3) ``./testssl.sh --help | grep gnmap``

PR #1156, opened 2018-11-19
WIP: Update client sim program

seccubus:
@drwetter this is my attempt to make this process more robust. Can you check the output?


drwetter:
If you could help with this #1158 that would be great.

seccubus:
@drwetter the size difference may be due to me removing some outdated custom signatures.

PR #1166, opened 2018-11-28
Don't run TLSv1.1 tests if server doesn't support TLSv1.1

dcooper16:
As a result of #276, `run_server_defaults()` makes several attempts to find certificates that a server offers if the ClientHello is for TLSv1.2 and no SNI is offered. However, these tests are unnecessary if it is already known that the server does not support TLSv1.1.

This PR modifies `run_server_defaults()` so that the the TLSv1.1-only tests are skipped if the server is known to not support TLSv1.1.


drwetter:
Thx

PR #1168, opened 2018-11-30
Fix typos

dcooper16:
The primary purpose of this PR is to fix the typo on line 14157, but it also corrects a number of typos that appear in comments.


drwetter:
thx

PR #1171, opened 2018-12-04
Fix #1165

### LINKED ISSUE #1165: Bug in TLSv1.3 client simulations
There is a bug in `client_simulation_sockets()` when it comes to handling TLSv1.3 connections in which the server sends a HelloRetryRequest.

After sending the initial ClientHello, `client_simulation_sockets()` calls:
```
resend_if_hello_retry_request "$tls_hello_ascii" "$cipher_list_2send" "$4" "$process_full"
```
A minor issue is that neither "$4" nor "$process_full" are defined ("$process_full" should just be replaced by "ephemeralkey"). The more serious problem is that `resend_if_hello_retry_request()` will basically just call `prepare_tls_clienthello()` to create the second ClientHello, and this second ClientHello may not look anything like the initial ClientHello.

I would rate this bug as minor for two reasons: (1) not many servers support TLSv1.3 at the moment and (2) I think it would be very uncommon for a TLSv1.3 server to respond to an initial ClientHello with a HelloRetryRequest rather than a ServerHello. However, this will need to be fixed at some point, which may require writing a new version of `resend_if_hello_retry_request()` that is specifically designed to take a provided ClientHello and modify it as required by a HelloRetryRequest.

dcooper16:
This PR fixes #1165 by changing `resend_if_hello_retry_request()` to modify the initial ClientHello rather than having it call `prepare_tls_clienthello()` to try to generate a new ClientHello that is almost the same as the first. The modification is done using a revised version of `create_client_simulation_tls_clienthello()`, which is now renamed as `modify_clienthello()`.

Since `prepare_tls_clienthello()` is no longer used to create a second ClientHello message, argument 7 to that function is no longer needed.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4186,12 +4186,22 @@ run_cipher_per_proto() {
 }
 
 # arg1 is an ASCII-HEX encoded SSLv3 or TLS ClientHello.
-# If the ClientHello contains a server name extension, then
-# either:
+# arg2: new key_share extension (only present to response to HelloRetryRequest)
+# arg3: cookie extension (if needed for response to HelloRetryRequest)
+#
+# This function may be used to either modify a ClientHello for client simulation
+# or to create a second ClientHello in response to a HelloRetryRequest.
+# If arg2 is present, then this is a response to a HelloRetryRequest, so the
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks!

PR #1192, opened 2019-02-08
Fix TLSv1.3 second ClientHello

dcooper16:
This PR fixes two problems with `modify_clienthello()`.

First, the function was incorrectly using the variable `$key_share` instead of `$new_key_share`. Since `$key_share` is defined when `modify_clienthello()` is called from `resend_if_hello_retry_request()`, but not when it is called from `client_simulation_sockets()`, this bug does not seem to result in incorrect behavior, but it should still be fixed.

Second, when this function is used to create a second ClientHello in response to a HelloRetryRequest, it removes the key_share extension from the original ClientHello and then appends the new key_share extension at the end. According to https://mailarchive.ietf.org/arch/msg/tls/8ZKCyamcYFaV90h6nf4MUnSPkEE, however, extensions must appear in the second ClientHello in the same order in which they appeared in the first ClientHello.

I am not aware of any servers that will actually complain if the extensions in the second ClientHello do not appear in the same order as in the first ClientHello, but this fix helps to ensure that the ClientHello messages testssl.sh sends are in compliance with the standard.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4301,7 +4301,7 @@ modify_clienthello() {
           offset+=+4
           len_extension=2*$(hex2dec "${tls_handshake_ascii:$offset:4}")
 
-          if [[ "$extension_type" == 0000 ]] && [[ -z "$key_share" ]]; then
+          if [[ "$extension_type" == 0000 ]] && [[ -z "$new_key_share" ]]; then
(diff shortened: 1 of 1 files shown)


drwetter:
Thank you!

PR #1193, opened 2019-02-12
Fix get_cipher()

dcooper16:
This PR fixes a bug in `get_cipher()` - one that also appears in `sclient_connect_successful()`.

The code currently assumes that cipher names contain only uppercase letters and numbers. However, ciphers that do not provide authentication include "anon" in the name, which is written in lowercase.

This PR fixes the problem by allowing lowercase letters to appear in cipher names (except in the first portion of the name).

Note that no change was made to similar code in `get_protocol()`, since the line in `get_protocol()` only matches TLSv1.3 ciphers, which do not contain any lowercase letters.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1246,9 +1246,9 @@ get_cipher() {
      local cipher=""
      local server_hello="$(< "$1")"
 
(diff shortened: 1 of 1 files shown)


drwetter:
oops, thanks for catching + fixing this!

PR #1194, opened 2019-02-12
Fix determine_optimal_proto()

dcooper16:
This PR fixes two bugs in `determine_optimal_proto()`.

First, `sslv2_sockets()` returns 3 if the connection was successful.

Second, if all connection attempts using `tls_sockets()` were unsuccessful, it is possible that `$TEMPDIR/$NODEIP.parse_tls_serverhello.txt` will not exist, so copying it or grepping it will lead to an error. Checking that `$proto` is not `22` will fix this as `$proto` will be empty if `$OPENSSL s_client` was used and it will be `00`, `01`, `02`, `03`, or `04` if `tls_sockets()` was used and the connection was successful with some protocol higher than SSLv2.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -17161,11 +17161,11 @@ determine_optimal_proto() {
                                    break
                               fi ;;
                          22)  sslv2_sockets
-                              [[ $? -eq 0 ]] && all_failed=false && add_tls_offered ssl2 yes && OPTIMAL_PROTO="-ssl2"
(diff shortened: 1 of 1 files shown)


drwetter:
Thank you, forget that.

At a certain point we should consider  fixing ``parse_sslv2_serverhello()`` or `` sslv2_sockets()`` to return what is expected.  A return value of 3 is counterintuitive .

PR #1196, opened 2019-02-12
Reorganize run_cipher_per_proto()

dcooper16:
This PR reorganizes `run_cipher_per_proto()`. Currently `run_cipher_per_proto()` runs a for loop, which loops over each protocol and prints the set of supported ciphers for each protocol. This PR simply places the body of the for loop in a separate function from the loop itself. This allows the body of the loop to be called for just a single protocol.

While this PR does not change the way that testssl.sh functions, it would allow for a future change in which `run_server_preferences()` called `cipher_pref_check()` for protocols in which the server enforces a cipher order and calls `ciphers_by_strength()` for protocols in which the server does not enforce a cipher order.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3955,8 +3955,12 @@ run_allciphers() {
 }
 
 # test for all ciphers per protocol locally configured (w/o distinguishing whether they are good or bad)
-run_cipher_per_proto() {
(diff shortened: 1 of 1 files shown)


drwetter:
Wonderful! Thank you!

PR #1197, opened 2019-02-12
Fix run_server_preference() for SSLv2 only server

dcooper16:
When `run_server_preference()` is run on a server that only supports SSLv2 it incorrectly reports that the server has a cipher order. The reason for this is that `$list_fwd` and `$list_reverse` only include one SSLv2 cipher.

In SSLv2 the server sends a list of all ciphers it supports in common with the client and the client chooses which cipher to use. As a result, the server cannot enforce a cipher order for SSLv2.

So, this PR fixes the problem in `run_server_preference()` by skipping the test for whether the server enforces a cipher order if `$OPTIMAL_PROTO` is -ssl2 and simply declares that the server does not enforce a cipher order.

Note that this PR is somewhat dependent on #1194, as #1194 needs to be applied in order for `$OPTIMAL_PROTO` to be set to -ssl2 when testing an SSLv2-only server.


drwetter:
Thank you!

PR #1198, opened 2019-02-12
Fix run_server_defaults() for SSLv2-only servers

dcooper16:
This PR makes a few improvements to `run_server_defaults()` when run on an SSLv2-only server.

First, it uses `sslv2_sockets()` to test the server rather than `$OPENSSL`, so that it will work even if `$OPENSSL` does not support SSLv2.

Second, it changes `run_server_defaults()` to only call `get_server_certificate()` once if `$OPTIMAL_PROTO` is -ssl2, since calling more than once is a waste - SSLv2 only supports ciphers that use RSA key exchange.

Finally, as some code assumes that `$TEMPDIR/intermediatecerts.pem` will exist, even if it is empty, this PR changes a couple of places that delete `$TEMPDIR/intermediatecerts.pem` to instead make the file empty.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7005,11 +7005,11 @@ get_server_certificate() {
      success=7
 
      if [[ "$OPTIMAL_PROTO" == -ssl2 ]]; then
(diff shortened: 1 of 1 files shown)


drwetter:
Thx!

PR #1199, opened 2019-02-14
Use sockets for get_server_certificate()

dcooper16:
This PR modifies  `get_server_certificate()` to use `tls_sockets()` rather than `$OPENSSL` for finding certificates using SSLv3 - TLSv1.2, unless `$SSL_NATIVE` is true. Using `tls_sockets()` allows testssl.sh to find certificates used by the server even if the server is only using cipher suites not supported by `$OPENSSL`. This may happen, for example, if the server only supports TLS_ECDHE_ cipher suites with curve X25519 and a version of OpenSSL prior to 1.1.0 is being used. A less likely possibility would be if the server had a certificate with a DH key, and a newer version of OpenSSL that does not support TLS_DH_ cipher suites is being used.

Since `tls_sockets()` cannot be used to obtain session tickets from the server, an additional test for session ticket lifetime needed to be added.

In order to reduce the number of times the server needs to be queried for certificates, this PR bundles the testing in a similar way to what is already done to test for cipher suites. Currently, each call to `get_server_certificate()` only tests for one type of certificate. This PR has each call test for more than one type of certificate. For example, one call is made to test for ECDSA, ECDH, DH, DSA, and GOST certificates. If the test is unsuccessful, then the server has none of these certificates. If the test finds a certificate (e.g., an ECDSA) certificate, then another test is run looking for the remaining types (ECDH, DH, DSA, and GOST) until a test is unsuccessful.

For most servers, this will reduce the number of calls to `get_server_certificate()` from 8 or 9 to 4 or 6.

Note that this PR does not completely eliminate the run_server_defaults()'s reliance on `$OPENSSL s_client`, as that is still needed to obtain a session ticket and to test session resumption. In addition, if `$OPENSSL` is OpenSSL 1.1.1, then `get_server_certificate()` uses `$OPENSSL` to find certificates using TLSv1.3, since `$OPENSSL s_client` is much faster than `tls_sockets()` and OpenSSL 1.1.1 implements all of the TLSv1.3 cipher suites as well as all of the curves permitted for use with TLSv1.3. Using OpenSSL 1.1.1 may cause a problem, however, if `$OPENSSL` and the server implement different drafts of TLSv1.3 (e.g., if `$OPENSSL` implements the final version of TLSv1.3 and the server implements draft26).

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6989,20 +6989,38 @@ extract_stapled_ocsp() {
      return 0
 }
 
-# arg1 is "-cipher <OpenSSL cipher>" or empty
+# arg1 is "<OpenSSL cipher>"
 # arg2 is a list of protocols to try (tls1_2, tls1_1, tls1, ssl3) or empty (if all should be tried)
 get_server_certificate() {
      local protocols_to_try proto
-     local success
+     local success ret
      local npn_params="" line
+     local ciphers_to_test=""
+     # Cipher suites that use a certifiate with an RSA (signature) public key
+     local -r a_rsa="cc,13, cc,15, c0,30, c0,28, c0,14, 00,9f, cc,a8, cc,aa, c0,a3, c0,9f, 00,6b, 00,39, c0,77, 00,c4, 00,88, c0,45, c0,4d, c0,53, c0,61, c0,7d, c0,8b, 16,b7, 16,b9, c0,2f, c0,27, c0,13, 00,9e, c0,a2, c0,9e, 00,67, 00,33, c0,76, 00,be, 00,9a, 00,45, c0,44, c0,4c, c0,52, c0,60, c0,7c, c0,8a, c0,11, c0,12, 00,16, 00,15, 00,14, c0,10"
+     # Cipher suites that use a certifiate with an RSA (encryption) public key
+     local -r e_rsa="00,b7, c0,99, 00,ad, cc,ae, 00,9d, c0,a1, c0,9d, 00,3d, 00,35, 00,c0, 00,84, 00,95, c0,3d, c0,51, c0,69, c0,6f, c0,7b, c0,93, ff,01, 00,ac, c0,a0, c0,9c, 00,9c, 00,3c, 00,2f, 00,ba, 00,b6, 00,96, 00,41, c0,98, 00,07, 00,94, c0,3c, c0,50, c0,68, c0,6e, c0,7a, c0,92, 00,05, 00,04, 00,92, 00,0a, 00,93, fe,ff, ff,e0, 00,62, 00,09, 00,61, fe,fe, ff,e1, 00,64, 00,60, 00,08, 00,06, 00,03, 00,b9, 00,b8, 00,2e, 00,3b, 00,02, 00,01"
+     # Cipher suites that use a certifiate with a DSA public key
+     local -r a_dss="00,a3, 00,6a, 00,38, 00,c3, 00,87, c0,43, c0,57, c0,81, 00,a2, 00,40, 00,32, 00,bd, 00,99, 00,44, c0,42, c0,56, c0,80, 00,66, 00,13, 00,63, 00,12, 00,65, 00,11"
+     # Cipher suites that use a certifiate with a DH public key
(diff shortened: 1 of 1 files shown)


drwetter:
Excellent, thanks!

testssl.sh
```diff
@@ -7575,6 +7654,20 @@ certificate_transparency() {
...
+     # Cipher suites that use a certifiate with a DH public key
+     local -r a_dh="00,a5, 00,a1, 00,69, 00,68, 00,37, 00,36, 00,c2, 00,c1, 00,86, 00,85, c0,3f, c0,41, c0,55, c0,59, c0,7f, c0,83, 00,a4, 00,a0, 00,3f, 00,3e, 00,31, 00,30, 00,bc, 00,bb, 00,98, 00,97, 00,43, 00,42, c0,3e, c0,40, c0,54, c0,58, c0,7e, c0,82, 00,10, 00,0d, 00,0f, 00,0c, 00,0b, 00,0e"
+     # Cipher suites that use a certifiate with an ECDH public key
+     local -r a_ecdh="c0,32, c0,2e, c0,2a, c0,26, c0,0f, c0,05, c0,79, c0,75, c0,4b, c0,4f, c0,5f, c0,63, c0,89, c0,8d, c0,31, c0,2d, c0,29, c0,25, c0,0e, c0,04, c0,78, c0,74, c0,4a, c0,4e, c0,5e, c0,62, c0,88, c0,8c, c0,0c, c0,02, c0,0d, c0,03, c0,0b, c0,01"
+     # Cipher suites that use a certifiate with an ECDSA public key
+     local -r a_ecdsa="cc,14, c0,2c, c0,24, c0,0a, cc,a9, c0,af, c0,ad, c0,73, c0,49, c0,5d, c0,87, 16,b8, 16,ba, c0,2b, c0,23, c0,09, c0,ae, c0,ac, c0,72, c0,48, c0,5c, c0,86, c0,07, c0,08, c0,06"
+     # Cipher suites that use a certifiate with a GOST public key
+     local -r a_gost="00,80, 00,81, ff,00, 00,82, 00,83"
```
drwetter: Haven't looked at all a_* variables above but as the day before yesterday I was looking at GOST ciphers:
FF,00 is a GOST cipher but seems to me rather RSA related. Isn't 00,85  missing, 00,82: Does that have a public key?

dcooper16: I created a script to create all of those lists using the same logic that now appears in `run_server_defaults()` (see below).

I just submitted a PR that moves ff,00 from a_gost to e_rsa, as it appears that this cipher suites uses RSA for both authentication and key exchange.

I found 00,85 (TLS_DH_DSS_WITH_CAMELLIA_256_CBC_SHA) in the a_dh list, so that one seems okay.

According to etc/cipher-mapping.txt, 00,82 (GOST94-NULL-GOST94) uses GOST94 for authentication, GOST for key exchange, and no encryption. So, it does seem to belong on the list of ciphers that use a GOST certificate.

```
if [[ "$ciph" == TLS_DHE_RSA_* ]] || [[ "$ciph" == TLS_ECDHE_RSA_* ]] || [[ "$ciph" == TLS_CECPQ1_RSA_* ]]; then
     ... RSA signature certificate ...
elif [[ "$ciph" == TLS_RSA_* ]] || [[ "$ciph" == SSL_* ]]; then
     ... RSA encryption certificate ...
elif [[ "$ciph" == TLS_DHE_DSS_* ]]; then
     ... DSA certificate ...
elif [[ "$ciph" == TLS_DH_* ]]; then
     ... DH certificate ...
elif [[ "$ciph" == TLS_ECDH_* ]]; then
     ... ECDH certificate ...
elif [[ "$ciph" == TLS_ECDHE_ECDSA_* ]] || [[ "$ciph" == TLS_CECPQ1_ECDSA_* ]]; then
     ... ECDSA certificate ...
elif [[ "$ciph" == TLS_GOST* ]]; then
     ... GOST certificate ...
fi
```

drwetter: > According to etc/cipher-mapping.txt, 00,82 (GOST94-NULL-GOST94) uses GOST94 for authentication, GOST for key exchange, and no encryption. So, it does seem to belong on the list of ciphers that use a GOST certificate.

In my humble practical experience with _Enc=Null ciphers_ way back with a customers F5 no certificate was returned. That was the reason of my second comment. Unfortunately the nginx+apache server I have currently compiled refused to serve a NULL cipher (w/o complaints when I do).

drwetter: My memory seems to be too weak ;-)

Just tried with ``s_server`` and a certificate was returned. So pls ignore my comment

PR #1200, opened 2019-02-15
TLS_GOSTR341094_RSA_WITH_28147_CNT_MD5 uses an RSA certificate

dcooper16:
As Dirk commented in #1199, TLS_GOSTR341094_RSA_WITH_28147_CNT_MD5 uses an RSA certificate, not a GOST certificate. So, this PR moves that cipher suite (0xff, 0x00) from the GOST list to the RSA list.

According to etc/cipher-mapping.txt, TLS_GOSTR341094_RSA_WITH_28147_CNT_MD5 uses RSA for both authentication and key exchange, so this PR places it on the list of cipher suites that uses RSA for encryption rather than signatures.


drwetter:
Thank you for the thorough (L8625) fix!

PR #1201, opened 2019-02-15
Fix check_tls12_pref()

dcooper16:
There is a minor bug when performing `run_server_preference()` if the server cannot handle ClientHello messages with more than 128 ciphers (i.e., `$SERVER_SIZE_LIMIT_BUG` is true) and the server supports at least one cipher in 'CAMELLIA:IDEA:KRB5:PSK:SRP:aNULL:eNULL'.

The problem is that `$OPENSSL s_client` is called with a cipher list such as
```
ECDHE-RSA-AES256-GCM-SHA384:CAMELLIA256-SHA:AES256-SHA256
```
then
```
ECDHE-RSA-AES256-GCM-SHA384:CAMELLIA256-SHA:AES256-SHA256:-CAMELLIA256-SHA
```
then
```
ECDHE-RSA-AES256-GCM-SHA384:CAMELLIA256-SHA:AES256-SHA256:-CAMELLIA256-SHA:-AES256-SHA256
```
and finally
```
ECDHE-RSA-AES256-GCM-SHA384:CAMELLIA256-SHA:AES256-SHA256:-CAMELLIA256-SHA:-AES256-SHA256:-ECDHE-RSA-AES256-GCM-SHA384
```
The last call to `$OPENSSL s_client` produces an error since the list of ciphers to send is empty, and this results in `connectivity_problem()` being called to print a "openssl s_client connect problem" warning. 

This PR fixes the problem by constructing a list of ciphers to test for and by not calling `$OPENSSL s_client` if the list is empty.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6270,7 +6270,7 @@ run_server_preference() {
 check_tls12_pref() {
      local batchremoved="-CAMELLIA:-IDEA:-KRB5:-PSK:-SRP:-aNULL:-eNULL"
      local batchremoved_success=false
-     local tested_cipher=""
+     local tested_cipher="" cipher ciphers_to_test
      local order=""
      local -i nr_ciphers_found_r1=0 nr_ciphers_found_r2=0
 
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks!

PR #1203, opened 2019-02-22
Update make-openssl.sh: Darwin compatibility

sxxs:
- Darwin doesn't build with -static (removed; file name suffix changed to "dynamic" in this case)
- Darwin has a different openssldir (/private/etc/ssl)
- script doesn't fail any more at make clean step in case there is no Makefile yet
- Darwin 64 bit compilation needs ./Configure instead of ./config and an explicit reference to darwin64-x86_64-cc

### DIFF
--- a/utils/make-openssl.sh
+++ b/utils/make-openssl.sh
@@ -5,7 +5,9 @@
 # instructions @ https://github.com/drwetter/testssl.sh/tree/2.9dev/bin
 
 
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks!

drwetter:
> Darwin 64 bit compilation needs ./Configure instead of ./config and an explicit reference to darwin64-x86_64-cc

I beautified the code a little bit (your change was not the reason). I forgot about that sentence though and thus accidentally changed 64 Bit to ``./config``. I am about to change it back but I wonder whether for 32 Bit it really should stay ``./config`` and not ``./Configure``?

PR #1205, opened 2019-02-26
Separate server preference test (cipher order) for TLS 1.3

### LINKED ISSUE #1163: Separate server preference test (cipher order) for TLS 1.3
Testing the server cipher order with openssl 1.1.1 against a server which supports TLS 1.3 may lead to the assumption that the server doesn't have a cipher order configured at all. However in some cases this assumption is not correct, presumably because the server is not using the correct calls to configure the cipher order (see  https://wiki.openssl.org/index.php/TLS1.3#Ciphersuites). 

Using an s_client connect with <= TLS 1.2 all is fine against such a server. 

Current example is testssl.net.

See also https://github.com/drwetter/testssl.sh/issues/1118#issuecomment-442407527

(On the long run we should get rid of OpenSSL here)

dcooper16:
This PR is an attempt to fix #1163 by running separate tests for a server cipher order preference for TLSv1.3 and for SSLv3 - TLSv1.2.

If the server supports TLSv1.3, then a test is performed to determine whether the server enforces a cipher order to TLSv1.3. A separate test is performed for SSLv3 - TLSv1.2 unless it is known that the server does not support any of these protocols.

If the server enforces a cipher order for SSLv3 - TLSv1.2, but not for TLSv1.3, then `cipher_pref_check()` is not called for TLSv1.3, since `cipher_pref_check()` is intended to show the cipher order that the server enforces. As TLSv1.3 will be the negotiated protocol if it is supported, the negotiated cipher for TLSv1.3 will already be presented.

This PR still has one major flaw, which may create a problem when testing a TLSv1.3-only server. If `run_protocols()` is run before `run_server_preference()`, then everything will be okay, as `run_server_preference()` will be able to determine that SSLv3 - TLSv1.2 are not supported. However, if `run_server_preference()` is run by itself, run_server_preference() will not know that SSLv3 - TLSv1.2 are not supported and so it will try to determine whether the server enforces a cipher preference order for these protocols. The attempt to connect to the server will fail, but at the moment `run_server_preference()` doesn't know whether the failure is because the server does not support SSLv3 - TLSv1.2 or because the server supports at least one of these protocols, but does not support any ciphers in `$list_fwd`. At the moment, `run_server_preference()` incorrectly flags an error.

One option would be to perform additional tests against the server in this case to determine the reason for the connection failure. Another option would be to have some code that is always run earlier, such as `determine_optimal_proto()`, test whether a server that supports TLSv1.3 supports any earlier protocols (SSLv3 - TLSv1.2).

A minor issue with this PR is that if the server enforces a cipher order for TLSv1.3, but not for SSLv3 - TLSv1.3, then `run_server_preference()` will just list the negotiated cipher for each protocol, but will not show the cipher order for TLSv1.3. However, as no servers have been encountered yet that work this way, it is not a significant problem.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6280,7 +6280,7 @@ run_server_preference() {
           pr_svrty_good "yes (OK)"; out " -- only for < TLS 1.3"
           fileout "$jsonID" "OK" "server -- TLS 1.3 client determined"
      elif ! "$has_cipher_order" && "$has_tls13_cipher_order"; then
-          pr_svrty_high "nope (NOT ok)"; out " -- only for TLS 1.3"
+          pr_svrty_high "no (NOT ok)"; out " -- only for TLS 1.3"
           fileout "$jsonID" "HIGH" "server -- < TLS 1.3 client determined"
      else
           if "$has_tls13_cipher_order"; then
--- a/testssl.sh
+++ b/testssl.sh
@@ -6292,6 +6292,10 @@ run_server_preference() {
                     out " -- TLS 1.3 and below"
                     fileout "$jsonID" "OK" "server"
                fi
+          else
+               # we don't have TLS 1.3 at all
+               pr_svrty_best "yes (OK)"
+               fileout "$jsonID" "OK" "server"
           fi
      fi
      outln
--- a/testssl.sh
+++ b/testssl.sh
@@ -6264,8 +6264,14 @@ run_server_preference() {
 
           [[ $cipher1 == $cipher2 ]] && has_cipher_order=true
      fi
+     debugme echo "has_cipher_order: $has_cipher_order"
+     debugme echo "has_tls13_cipher_order: $has_tls13_cipher_order"
 
-     if ! "$has_cipher_order" && ! "$has_tls13_cipher_order"; then
+     if "$TLS13_ONLY" && ! "$has_tls13_cipher_order"; then
+          out "nope"
+          limitedsense=" (limited sense as client will pick)"
+          fileout "$jsonID" "INFO" "NOT a cipher order configured"
+     elif ! "$has_cipher_order" && ! "$has_tls13_cipher_order"; then
           # server used the different ends (ciphers) from the client hello
           pr_svrty_high "nope (NOT ok)"
           limitedsense=" (limited sense as client will pick)"
@@ -6274,10 +6280,11 @@ run_server_preference() {
           pr_svrty_good "yes (OK)"; out " -- only for < TLS 1.3"
           fileout "$jsonID" "OK" "server -- TLS 1.3 client determined"
      elif ! "$has_cipher_order" && "$has_tls13_cipher_order"; then
-          pr_svrty_good "yes (OK)"; out " -- only for TLS 1.3"
-          fileout "$jsonID" "OK" "server -- < TLS 1.3 client determined"
+          pr_svrty_high "nope (NOT ok)"; out " -- only for TLS 1.3"
+          fileout "$jsonID" "HIGH" "server -- < TLS 1.3 client determined"
      else
           pr_svrty_best "yes (OK)"
+          "$has_tls13_cipher_order" && out " -- TLS 1.3 and below"
           fileout "$jsonID" "OK" "server"
      fi
      outln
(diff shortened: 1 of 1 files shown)


drwetter:
Hi @dcooper16,

thanks!

Problem no 1 

> This PR still has one major flaw, which may create a problem when testing a TLSv1.3-only server. If run_protocols()

is easily solvable, maybe a bit ugly. But that's ok.

Problem no 2

> A minor issue with this PR is that if the server enforces a cipher order for TLSv1.3, but not for SSLv3 - TLSv1.3, then run_server_preference() will just list the negotiated cipher for each protocol, but will not show the cipher order for TLSv1.3. However, as no servers have been encountered yet that work this way, it is not a significant problem.

... is more problematic. Have you tried testssl.net or google.com? (for google you might want to supply an IP address like ``./testssl.sh -q  --ip=216.58.211.3  -P google.com`` as they seem to have servers at least in Europe which sometimes don't have TLS 1.2).

dcooper16:
Hi Dirk,

If #1216 is merged, then the information collected by `determine_optimal_sockets_params()` can be used to address problem 1.

I just tested testssl.net and google.com, and they both seem to have the opposite of problem 2. They enforce a cipher order for SSLv3 - TLSv1.2, but not for TLSv1.3. This PR handles servers like these by simply skipping over TLSv1.3 when showing "Cipher Order" since there is no server-enforced cipher order for TLSv1.3.

I think problem 2 should be addressed, even if we don't know of any servers that work that way. I just believe that addressing problem 2 will require a more substantial change to `run_server_preference()`, and so it should wait until after 3.0 has been released.

drwetter:
ok, thanks

drwetter:
sorry, David. I didn't mean to close it! 

I was just deleting the branch and as a consequence github must have closed it. There were more PRs waiting (another one from you). I try to recover the branch first

dcooper16:
Now that #1216 has been merged, I changed this PR to use the information `determine_optimal_sockets_params()` collects to determine if the server only supports TLSv1.3.  This addresses the flaw that I mentioned when first posting this PR. So, I believe that this PR is now ready to be merged.

drwetter:
Hi David,

thanks.

When I test a server which has no TLS < 1.3 I get the following (nginx under debian buster with ``ssl_prefer_server_ciphers off`` which probably sets both cipher orders):

```
Has server cipher order?     nope (NOT ok)
 Negotiated protocol          TLSv1.3
 Negotiated cipher            TLS_AES_256_GCM_SHA384, 253 bit ECDH (X25519) (limited sense as client will pick)
 Negotiated cipher per proto  (limited sense as client will pick)
     TLS_AES_128_GCM_SHA256:        TLSv1.3
 No further cipher order check has been done as order is determined by the client
```

which I think is a bit of a tough verdict. I think here the cipher order doesn't matter as much. Maybe either informational  or low severity would be better? 

And, related to that, I think when there's no cipher order for TLS <=1.2 and there's one for TLS 1.3 the severity is too good. If we assume the worst case (client with TLS 1.2 maximum) this should be high.

FYI: It looks to me like the cipher order setting changed. IIRC previously the admin had no influence on the TLS 1.3 cipher order (nginx at least, apache: I don't remember). No if I set ``ssl_prefer_server_ciphers off`` as said it seems to set both. So I can't test the other cases (which should not concern you, but worries me a bit).

Cheers, Dirk

dcooper16:
Good points. The first seems related to #1311. If all TLS 1.3 ciphers are considered strong, then it doesn't really matter whether the server enforces a cipher order. (However, [draft-camwinget-tls-ts13-macciphersuites](https://tools.ietf.org/html/draft-camwinget-tls-ts13-macciphersuites) proposes to define authentication-only cipher suites for TLS 1.3.)

At the same time, if a server enforced a cipher order for TLS 1.3, but not for earlier protocol versions, that should be considered just as bad as a server that doesn't enforce a cipher order for any protocol version.

In the future, something more sophisticated could be done, where we first determine the full set of cipher suites supported for each protocol version, and then only mark the lack of a server-enforced cipher order as an issue if the server supports any "weak" ciphers. But, that would have to come after 3.0.

drwetter:
Hi David,

just about to add a few bits..

> (However, draft-camwinget-tls-ts13-macciphersuites proposes to define authentication-only cipher suites for TLS 1.3.)

but atm we don't test for them. So this we can keep in mind for later.

testssl.sh
```diff
@@ -6283,9 +6283,16 @@ run_server_preference() {
...
+               if "$TLS13_ONLY"; then
+                    out "yes (TLS 1.3 only)"
+                    fileout "$jsonID" "INFO" "server (TLS 1.3)"
+               else
+                    pr_svrty_best "yes (OK)"
+                    out " -- TLS 1.3 and below"
+                    fileout "$jsonID" "OK" "server"
+               fi
+          fi
      fi
```
dcooper16: I believe that there is a problem here. Given the above code, if a server enforces a cipher order but does not support TLS 1.3, nothing is printed.

The code above applies whenever the server enforces a cipher order for all supported protocol versions. Given the previous `if` statements, `$has_cipher_order` must be `true` and either (1) the server supports TLS 1.3 and `$has_tls_cipher_order` is `true` or (2) the server does not support TLS 1.3 and `$has_tls_cipher_order` is `false`.

The above code handles the case in which the server supports TLS 1.3, but not the case in which TLS 1.3 isn't supported.

drwetter:
Thanks, David!

PR #1211, opened 2019-03-06
Extend determine optimal proto sockets

dcooper16:
At the moment, `determine_optimal_proto_sockets_helper()` will flag a protocol as supported if it determines that the protocol is supported (i.e., it will call "`add_tls_offered p yes`" for the protocol). However, it will not flag any protocols as not supported.

`determine_sizelimitbug()` needs to know whether TLSv1.2 is supported. If "`has_server_protocol 03`" returns 2 (unknown), then `determine_sizelimitbug()` runs a test to determine whether TLSv1.2 is supported. This PR extends `determine_optimal_proto_sockets_helper()` to call "`add_tls_offered tls1_2 no`" if TLSv1.2 is not supported, so that the extra test in `determine_sizelimitbug()` is not needed.

Similarly, when checking for whether the server supports a cipher order, #1205 needs to know whether the server supports any protocol in SSLv3 - TLSv1.2. So, this PR extends `determine_optimal_proto_sockets_helper()` to continue running tests against the server until it can call "`add_tls_offered p yes`" for some protocol in SSLv3 - TLSv1.2 or until it has called "`add_tls_offered p no`" for all of them.

Another issue this PR addresses is that there are some servers that support TLSv1.2, but do not support any of the ciphers in `$TLS12_CIPHER` (see #806 and #807), which is why `$TLS12_CIPHER_2ND_TRY` was defined. So, if "`tls_sockets 03 $TLS12_CIPHER`" does not return 0, this PR has `determine_optimal_proto_sockets_helper()` try "`tls_sockets 03 $TLS12_CIPHER_2ND_TRY`" before determining that TLSv1.2 is not supported. If `$TLS12_CIPHER_2ND_TRY`, but `$TLS12_CIPHER` did not, then `TLS12_CIPHER` is changed to `$TLS12_CIPHER_2ND_TRY` so that later tests using `$TLS12_CIPHER` will succeed. With this change, `run_protocol()` and `run_grease()` no longer need to try using `$TLS12_CIPHER_2ND_TRY` if `$TLS12_CIPHER` failed.

Note that this PR does not address the problems mentioned above if --ssl-native is used,


drwetter:
Hi David,

Tthanks for the PR.

There a a couple of changes not necessarily related. Maybe we can move forward doing one step after another.

> Similarly, when checking for whether the server supports a cipher order, #1205 needs to know whether the server supports any protocol in SSLv3 - TLSv1.2. So, this PR extends determine_optimal_proto_sockets_helper() to continue running tests against the server until it can call "add_tls_offered p yes" for some protocol in SSLv3 - TLSv1.2 or until it has called "add_tls_offered p no" for all of them.

Can't this go into #1205?

> Another issue this PR addresses is that there are some servers that support TLSv1.2, but do not support any of the ciphers in $TLS12_CIPHER (see #806 and #807), which is why $TLS12_CIPHER_2ND_TRY was defined. So, if "tls_sockets 03 $TLS12_CIPHER" does not return 0, this PR has determine_optimal_proto_sockets_helper() try "tls_sockets 03 $TLS12_CIPHER_2ND_TRY" before determining that TLSv1.2 is not supported. 

fine.

> If $TLS12_CIPHER_2ND_TRY, but $TLS12_CIPHER did not, then TLS12_CIPHER is changed to $TLS12_CIPHER_2ND_TRY so that later tests using $TLS12_CIPHER will succeed.

sounds like a good idea. Can this go into a separate PR?

Another thing --> see code comment I'll do in a minute.

testssl.sh
```diff
@@ -17379,51 +17373,92 @@ sclient_auth() {
...
+                              all_failed=false
+                         fi
+                    fi
+                    "$all_failed" || break ;;
                04) tls_sockets "$proto" "$TLS13_CIPHER"
                     if [[ $? -eq 0 ]]; then
                          add_tls_offered tls1_3 yes; optimal_proto="-tls1_3"
                          all_failed=false
-                         break
+                         # Don't break if this test succeeds, as some tests
```
drwetter: Don't want to do a full protocol check here. For this task run_protocols() should be used.

determine_optimal_proto() is ONLY for dealing with broken servers which otherwise -- no protocol supplied in a openssl clienthello -- refuse serve their expected serverhello properly. See comment above determine_optimal_proto().

dcooper16:
I decided to close this PR in favor of #1216. #1216 addresses the need to determine what value to use for KEY_SHARE_EXTN_NR for a TLSv1.3 ClientHello and what set of ciphers to include in a TLSv1.2 ClientHello, but does so in a separate function rather than as part of determine_optimal_proto().

PR #1215, opened 2019-03-13
determine_optimal_proto_sockets_helper() speedup and bug fix

dcooper16:
There is currently a bug in `determine_optimal_proto_sockets_helper()`. In two places there is code of the form:
```
   tls_sockets ...
   if [[ $? -eq 0 ]]; then
        ...
   elif [[ $? -eq 2 ]]; then
        ...
   fi
```
This code does not work as intended since the second check (`elif [[ $? -eq 2 ]]`) is actually comparing the results of the first check to 2 rather than the results of the call to `tls_sockets()`.

This PR fixes that problem and also speeds up the code. Since `tls_sockets()` sets `$DETECTED_TLS_VERSION` to the protocol version that was negotiated, there is no need to scan `$TEMPDIR/$NODEIP.parse_tls_serverhello.txt` for this information.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -17382,19 +17382,18 @@ determine_optimal_proto_sockets_helper() {
      local proto=""
      local optimal_proto=""
      local starttls="$1"
+     local -i ret
 
      for proto in 03 01 04 00 02 22; do
           case $proto in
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks!

PR #1216, opened 2019-03-13
Add determine_optimal_sockets_params()

dcooper16:
This PR adds a new helper function that is run just prior to `determine_optimal_proto()` and that determines the what information `tls_sockets()` should include in a ClientHello.

For a TLSv1.3 ClientHello, `determine_optimal_sockets_params()` determines whether `tls_sockets()` should use 0x33 or 0x28 are the extension number for the key_share extension. 0x33 should be used with servers that support RFC 8446 or drafts 23-28. 0x28 should be used with servers that support drafts 18-22.

For a TLSv1.2 ClientHello, `determine_optimal_sockets_params()` determines what cipher list should be provided to tls_sockets(). For most servers, the list of ciphers in `$TLS12_CIPHER` works best. But, there are some servers that do not support any ciphers in `$TLS12_CIPHER`, but do support one or more ciphers in `$TLS12_CIPHER_2ND_TRY`.


dcooper16:
I just added another commit to this PR adds some more checks to determine_optimal_sockets_params(). These additional checks will almost never need to be run, and so will not slow down the typical run of testssl.sh, but adding them will provide information that can be useful for other parts of testssl.sh.

These additional checks will only be run if the server does not support TLSv1.3 and a TLSv1.2 ClientHello is not successful. This means that either:
   * the server is not an TLS/SSL enabled server;
   * the server only supports SSLv2; or
   * the server supports some protocol in SSLv3 - TLSv1.1, but does not handle version negotiation correctly.

The additional checks determine which of the above is the case.

Adding these additional checks helps in at least the following ways.

If determine_optimal_proto() (assuming it is reverted to using OpenSSL) is unable to connect to the server using OpenSSL, it will be possible determine whether the problem is that server does not support TLS/SSL or that a different version of OpenSSL is needed to test the server, and so the appropriate error message can be provided.

If the code in #1205 for run_server_preference() is unable to connect when checking for a cipher order, the reason for the failure will be known, making it possible to determine the correct response to the failure.

This commit also defines a new global, OPTIMAL_SOCKETS_PROTO, that gets set if the server does not handle version negotiation correctly, but it does not make use of the variable.

drwetter:
Thank you!

At some point we should merge somehow the two ``determine_optimal_*()`` with ``run_protocols()``.  How would need to be defined.

dcooper16:
> At some point we should merge somehow the two determine_optimal_*() with run_protocols(). How would need to be defined.

Agreed. I don't think it's exactly a matter of merging them though. I think it's just a matter of using `has_server_protocol()` in `run_protocols()` (and perhaps elsewhere) to avoid testing something for which the results are already known.

At the moment, however, I am working on `determine_optimal_proto()`. Since the point of  `determine_optimal_proto()` is to set the value of `$OPTIMAL_PROTO`, I think the testing should be done using `$OPENSSL`. After all, my understanding is that `$OPTIMAL_PROTO` was defined because with some servers and some versions of OpenSSL, `$OPENSSL s_client` did not succeed unless a protocol was specified. The best way to determine what will work best with the `$OPENSSL s_client` command line is to do the testing with `$OPENSSL s_client`.

`determine_optimal_proto()` can then use the results of `determine_optimal_sockets_params()`. If `determine_optimal_sockets_params()` was able to connect to the server using `tls_sockets()` or `sslv2_sockets()`, but `determine_optimal_proto()` could not connect using `$OPENSSL`, then this can be reported to the user, perhaps with a recommendation of what the user should do:

* TLSv1.3-only server - try using OpenSSL 1.1.1.
* SSLv2 or SSLv3-only server - try using an older OpenSSL, such as the one distributed with testssl.sh.

I'm also thinking about using the results of `determine_optimal_sockets_params()` in `determine_sizelimitbug()`. I believe that we've determined that the bug occurs when ClientHello contains more than 128 ciphers. `determine_optimal_sockets_params()` determines whether TLSv1.2 is supported and sets `$TLS12_CIPHER` to a list of exactly 128 ciphers that is guaranteed to include at least one cipher supported by the server. So, `determine_sizelimitbug()` could just make one call to `tls_sockets()` of the form:
```
tls_sockets 03 "${TLS12_CIPHER}, ${overflow_cipher}"
```
where `${overflow_cipher}` is chosen to be some cipher not in `$TLS12_CIPHER`.

I see that `$test_ciphers` in `determine_sizelimitbug()` includes some ciphers that are not in either `$TLS12_CIPHER` or `$TLS12_CIPHER_2ND_TRY`. If you think there may be some servers that support one of these ciphers, but none of the ciphers in `$TLS12_CIPHER` or `$TLS12_CIPHER_2ND_TRY`, then we should redefine `$TLS12_CIPHER` and `$TLS12_CIPHER_2ND_TRY` (and possibly add a `$TLS12_CIPHER_3ND_TRY`), but still leave it to `determine_optimal_sockets_params()` to find a set of 128 ciphers that works with the server with a TLSv1.2 ClientHello.

PR #1222, opened 2019-04-01
Fix escaping error in JSON output for OCSP stapling check

andrewbonney:
This is a very minor fix to remove some unnecessary escape characters which cause an error generating JSON output for an OCSP error state.


drwetter:
Thank you!

PR #1224, opened 2019-04-01
Fix #1223

### LINKED ISSUE #1223: OCSP Stapling check causes non-zero exit code when --phone-out not set
I'm not certain if this is a bug or simply an error in my setup, but just in case it's the former...

When testing using `testssl.sh -S` and with `--phone-out` not set because I'm aware my OCSP server is unreachable, I was expecting the OCSP stapling check to be skipped around https://github.com/drwetter/testssl.sh/blob/2.9dev/testssl.sh#L8500. Instead, the `$ocsp_response` variable was set to the following:

```
OCSP response:
===============================================================================
Responder Error: trylater (3)
```

Which resulted in the tests hitting the following error handler https://github.com/drwetter/testssl.sh/blob/2.9dev/testssl.sh#L8520

Linux 4.4.0-142-generic #168-Ubuntu SMP Wed Jan 16 21:00:45 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
testssl.sh       3.0rc4 from https://testssl.sh/dev/
(4fd69fc 2019-04-01 11:49:17 -- )

dcooper16:
This PR fixes #1223 by checking whether the stapled OCSP response from the server is an error message.

Another way to fix #1223 would be to just change line 8510 to:
```
if grep -a "OCSP Response Status" <<< "$ocsp_response_status" | grep -q successful || \
     [[ "$ocsp_response" =~ Responder\ Error: ]]; then 
```
However, I believe this alternative would lead to confusing results, since testssl.sh would then print

     offered, error querying OCSP responder (tryLater)

I'm not sure whether it makes sense to say "offered" when the stapled response that is provided is just an error message, but I think it is important to make clear that the error response was received from the TLS server, and that it wasn't testssl.sh that tried querying the OCSP responder.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7808,6 +7808,7 @@ certificate_info() {
      local days2warn1=$DAYS2WARN1
      local provides_stapling=false
      local caa_node="" all_caa="" caa_property_name="" caa_property_value=""
+     local response=""
 
      if [[ $number_of_certificates -gt 1 ]]; then
           [[ $certificate_number -eq 1 ]] && outln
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, David. 

@andrewbonney : Speak up if this doesn't resolve the problem

andrewbonney:
That does the job, thanks both.

PR #1230, opened 2019-04-03
[WIP] add log color switch support

noraj:


drwetter:
Closing as you can do that manually if you want it. It's too much and we rather focus on the core functions

PR #1244, opened 2019-04-25
Fix #1243

### LINKED ISSUE #1243: JSON file error - No separators (,) between targetHosts objects
On the JSON output file, there is no delimiter between targetHosts objects, so the parsing of the file fails with python json library with the following message:

```
In [23]: json.load(open("testssl.targets.port.443.json","r"))
[...]
ValueError: Expecting , delimiter: line 8068 column 11 (char 519745)
```

Example of current JSON file:
```
[...]
                           },{
                                "id"           : "clientsimulation-openssl_102e",
                                "severity"     : "INFO",
                                "finding"      : "TLSv1.2 ECDHE-RSA-AES256-SHA384"
                           }
                    ]
-->       }
-->       {
                    "targetHost"      : "<IP>",
                    "ip"              : "<IP>",
                    "port"            : "443",
[...]
```
Between the two pointed curly braces there should be a comma.

1. uname -a
Linux felipe-host 4.9.0-8-amd64 #1 SMP Debian 4.9.144-3 (2019-02-02) x86_64 GNU/Linux
   
2. testssl version from the banner: testssl.sh -b 2>/dev/null | head -4 | tail -2
testssl.sh       3.0rc4 from https://testssl.sh/dev/

3. git log | head -1 (if running from git repo)
commit bc6b2c6f94dc6a76f9cc9951a601dbc020f97b6b

4. openssl version used by testssl.sh: 
Using "OpenSSL 1.0.2-chacha (1.0.2k-dev)"

5. steps to reproduce: testssl.sh or docker command line, if possible incl. host
Scan multiple IP hosts and output the results to JSON with the flag -oA.

dcooper16:
This PR fixes #1243 by modifying `create_mass_testing_cmdline()` to handle the `--outfile`, `-oa`, `--outFile`, and `-oA` options in the case that the filename provided is not "auto." It also modifies `create_mass_testing_cmdline()` so that in the case of serial testing `-oj` and `-oJ` are treated the same as `--jsonfile` and `--jsonfile-pretty`, respectively.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -17941,7 +17941,8 @@ run_mx_all_ips() {
 create_mass_testing_cmdline() {
      local testing_type="$1"
      local cmd test_number
-     local -i nr_cmds=0
+     local outfile_arg
+     local -i nr_cmds=0 index=0
      local skip_next=false
 
      MASS_TESTING_CMDLINE=()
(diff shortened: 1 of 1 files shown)


drwetter:
great, David!

PR #1245, opened 2019-04-25
The -outfile, -oa, -outFile, and -oA options should accept a directory

dcooper16:
Currently the `--outfile`, `-oa`, `--outFile`, and `-oA` options assume that the \<fname\> being provided is to be used as a filename, unless it is "auto." However, all of the individual options (e.g., `--logfile`) allow for a directory name to be provided instead of a file name.

This PR changes the handling of the `--outfile`, `-oa`, `--outFile`, and `-oA` options so that if a directory name is provided, the files are created in that directory.


drwetter:
You're right! Thank you.

PR #1252, opened 2019-04-29
Mark only TLSv1.3 final as pr_svrty_best

dcooper16:
This PR changes `run_protocols()` so that, when using `tls_sockets()`, support for TLSv1.3 is only marked as `pr_svrty_best()` if the final (RFC 8446) version is supported. It also changed run_protocols() so that support for TLSv1.3 is marked as `pr_svrty_best()` if OpenSSL is used (i.e., if the `--ssl-native` option is specified).

One potential issue is that the `--ssl-native` version assumes that if OpenSSL supports TLSv1.3 it supports the final (RFC 8446) version of the protocol. If the tester is using a development version of OpenSSL 1.1.1 rather than the final version, then the protocol test will actually be indicating whether the server supports the same draft version of OpenSSL as the `$OPENSSL` being used to perform the tests.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4906,7 +4906,7 @@ run_protocols() {
                ;;
           2)   pr_svrty_medium "not offered"
                add_tls_offered tls1 no
-               if [[ "$DETECTED_TLS_VERSION" == "0300" ]]; then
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, @dcooper16 .

Issue #1158, opened 2018-11-21
Provide more client data


drwetter:
... compatible with SSLlab API format, so that this data can be merged into ``etc/client-simulation.txt``.

Also: Documentation on usage would be needed.

drwetter:
Missing is clienthello data at least from Android 8 and 9, OpenSSL 1.1.1, Safari on OSX 11 and 12. Java 10 and 11. See also #1104

drwetter:
I started with that in a separate branch: https://github.com/drwetter/testssl.sh/compare/more_client_sim.

Help would be appreciated. See idea here and howto 
![Screenshot_20190418_113959](https://user-images.githubusercontent.com/8036727/56352951-d6afbd00-61bf-11e9-96af-33fec2b64f45.png) 
![Screenshot_20190418_114104](https://user-images.githubusercontent.com/8036727/56352961-db747100-61bf-11e9-9387-485026a1bea1.png)

drwetter:
Result in this branch: 
![Screenshot_20190418_114316](https://user-images.githubusercontent.com/8036727/56353034-05c62e80-61c0-11e9-9d7b-f6bf74b13ccb.png)

drwetter:
see c183c213e5309df12cb2b3a9f74aacd0e20e23f1 (added FF 66, OpenSSL 1.1.{0,1}, Android 8.1)

drwetter:
Safari @ OSX 12 done in PR #1260.

Open: Safari @ OSX 11. Java 10+11

drwetter:
closing now, OSX 11. Java 10+11 are EOL

PR #1268, opened 2019-05-06
Add Safari 12.1 on macOS 10.13.6

csett86:
manually wiresharked, one more for #1158.


drwetter:
Thanks. Could you add the section to the wireshared file too, pls?

csett86:
What exactly do you mean? the new section is already part of both client-simulation files.

drwetter:
sorry, seems that wasn't enough coffee this morning yet ;-)

Thanks!

PR #1269, opened 2019-05-06
Deprecate Java 9, its EOL since March 2018

csett86:
No current distro (Ubuntu, Debian, Fedora) is still shipping it,
Oracle has EOLed it in March 2018 according to

https://www.oracle.com/technetwork/java/java-se-support-roadmap.html


drwetter:
Thanks!

Issue #1176, opened 2018-12-18
pwnedkeys


schuetzm:
Matt Palmer just anounced a public database of compromised keys:

https://groups.google.com/forum/#!topic/mozilla.dev.security.policy/mSXCMmw8x8M
https://pwnedkeys.com/

Might be useful to integrate this.

drwetter:
Thx!

That wouldn't be difficult to set up. I am only wondering why the returned JSON payload value is base64 encoded :-( But that's no real issue.

From the practical perspective I am wondering how many keys are in his DB and it's hard to understand why he doesn't include RSA keys < 2048 bit. Yeah, it's weak, but still...

drwetter:
PoC: 

```curl -s https://v1.pwnedkeys.com/9e03b56749abe821a6f5299d6f634b35404975f0552eb3347bf3adfad9af1109 | sed -e 's/.*payload":"//g' -e 's/","protected".*$//' | openssl base64 -d -A
```

drwetter:
... and what is weird, if I try with the certificate from testssl.shl

```
prompt% curl -s https://v1.pwnedkeys.com/ca0f134494ea7cad519a95e2595714d82fac2a1bb9aa04476b7590cf1886c38c
<html>
<head><title>404 Not Found</title></head>
<body bgcolor="white">
<center><h1>404 Not Found</h1></center>
<hr><center>nginx</center>
</body>
</html>
prompt% 
```

So, the JSON has base64 encoded cleartext values, keys not in his DB return a server error and he refuses to list small keys. Thus at the moment I would rather say, let's wait until this matures. Until then this will carry a ``wontfix``. 

@schuetzm : if you happen to see a change, please let me know.

dcooper16:
The behavior is consistent with the information at https://pwnedkeys.com/search.html:
> Make a request to https://v1.pwnedkeys.com/<fingerprint\>. If the response to the request is a 200 OK, then the key appears in the database, while a 404 response indicates that the key doesn’t appear in the pwnedkeys database.

So, one could just do the query as:
```
curl --head https://www.google.com/ca0f134494ea7cad519a95e2595714d82fac2a1bb9aa04476b7590cf1886c38c
```
and the search the response for either "200 OK" or "404 Not Found". There doesn't seem to be any need to look at the JSON.

The value does seem to be limited, though. Presumably the key is bad if the server responds "200 OK", but a response of "404 Not Found" doesn't really provide much assurance.

[...]

PR #1274, opened 2019-05-22
Check pwnedkeys.com database

dcooper16:
This PR addresses #1176 by adding a check of whether the server's public key appears in the https://pwnedkeys.com database.


drwetter:
Thanks a lot, David!

PR #1279, opened 2019-06-06
Initial support for certificate compression

dcooper16:
[draft-ietf-tls-certificate-compression](https://tools.ietf.org/html/draft-ietf-tls-certificate-compression) specifies a new TLS extension that allows a client to indicate support for receiving the server's certificates in compressed form. This PR adds initial support for that extension to testssl.sh. It was developed based on an [announcement](https://mailarchive.ietf.org/arch/msg/tls/idYj6mgvrlYBbdSfAOcw19pOv4A) that facebook.com has implemented support for the extension.

This PR does not add a test for a server's support for the new extension, it just adds code to `parse_tls_serverhello()` that will parse a compressed certificate message if one is present in the server's response. So, in practice, the code in this PR will not actually be used until additional code has been added that sends a ClientHello with the compress_certificate extension.

The code in this PR can be tested by using the `--devel` option and by changing line 19347 from
```
tls_sockets "$TLS_LOW_BYTE" "$HEX_CIPHER" "ephemeralkey"
```
to 
```
tls_sockets "$TLS_LOW_BYTE" "$HEX_CIPHER" "all+" "00,1b, 00,03, 02, 00,01"
```
testssl.sh can then be called using
```
testssl.sh --debug 3 --devel 04 "13,01" facebook.com
```
Note that this PR adds a test for zlib support to `find_openssl_binary()`. The test uses `$OPENSSL zlib` in the same way it is used in `parse_tls_serverhello()`. The test asks `$OPENSSL zlib` to uncompress the compressed version of the string "zlib" and the checks to see whether the decompression was performed correctly. There is already a test for zlib support in `run_crime()`, but I did not check to see whether the check in `run_crime()` could be replaced with a check of the new `$HAS_ZLIB` variable.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -351,6 +351,7 @@ HAS_IRC=false
 HAS_CHACHA20=false
 HAS_AES128_GCM=false
 HAS_AES256_GCM=false
+HAS_ZLIB=false
 OSSL_CIPHERS_S=""
 PORT=443                                # unless otherwise auto-determined, see below
 NODE=""
@@ -11587,6 +11588,7 @@ parse_tls_serverhello() {
                     16) tmln_out " (certificate_status)" ;;
                     17) tmln_out " (supplemental_data)" ;;
                     18) tmln_out " (key_update)" ;;
+                    19) tmln_out " (compressed_certificate)" ;;
                     FE) tmln_out " (message_hash)" ;;
                     *) tmln_out ;;
                esac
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -16556,6 +16588,8 @@ find_openssl_binary() {
      $OPENSSL enc -aes-256-gcm -K 0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef -iv 0123456789abcdef01234567  > /dev/null 2> /dev/null <<< "test"
      [[ $? -eq 0 ]] && HAS_AES256_GCM=true
 
+     [[ "$(echo -e "\x78\x9C\xAB\xCA\xC9\x4C\xE2\x02\x00\x06\x20\x01\xBC" | $OPENSSL zlib -d 2>/dev/null)" == zlib ]] && HAS_ZLIB=true
+
```
drwetter: Albeit this is certainly cool, it is problematic on systems whose OpenSSL don't have zlib support:

```
prompt> echo -e "\x78\x9C\xAB\xCA\xC9\x4C\xE2\x02\x00\x06\x20\x01\xBC" | openssl zlib -d 2 > /dev/null
openssl:Error: 'zlib' is an invalid command.

Standard commands
asn1parse         ca                ciphers           crl               
[..]
```

I suspect MacOS is one of them

drwetter:
Hi David,

good idea. Also for HAS_ZLIB I thought about that a while back too. But please see inline comment there.

> The code in this PR can be tested by using the --devel option and by changing line 19347 from

No quite sure I get this. Is there any catch changing this line (it's L19313, right?) 

Dirk

testssl.sh

The above code worked for me, even with versions of OpenSSL that did not have zlib support. I believe the problem is the extra spaces in the test that you performed. By writing "2 >" rather than "2>", the output of stderr is not redirected to /dev/null. Even if OpenSSL has zlib support, the test you ran would not work, as it would treat "2" as a parameter to openssl rather than as part of the redirect command.

dcooper16:
> > The code in this PR can be tested by using the --devel option and by changing line 19347 from
> 
> No quite sure I get this. Is there any catch changing this line (it's L19313, right?)

Yes, it would be L19313 in the current code, and L19347 after this PR is applied.

This line needs to be changed for two reasons:

1. The ClientHello needs to include the compress_certificate extension, and there is no way to specify additional extensions at the command line.

2. If the third parameter to `tls_sockets()` is "ephemeralkey" rather than "all" or "all+", then testssl.sh won't even decrypt the CompressedCertificate message, let alone try to uncompress it. If you specify "all" rather than "all+", then the facebook.com will select X25519 even if OpenSSL cannot handle that curve. Using "all+" ensures that the ClientHello will not offer X25519 unless you are using an OpenSSL that supports it.

I didn't make this change in the PR, however, since it didn't seem appropriate to change the `--devel` option to always offer the compress_certificate extension.

testssl.sh
drwetter: Sorry. I tried it under an old FreeBSD but I forgot the default shell is /bin/csh :-/

A little bit strange though - as I pasted your code into the command line the paste was "1:1" but threw the error. When I pasted the code my alleged proof I used the cmd line repetition which then showed the blank inserted.

drwetter:
Thanks

PR #1285, opened 2019-06-24
Fix typo in check of $COLOR

dcooper16:
In `emphasize_stuff_in_headers()` there is a check of the `$COLOR` value, which checks whether `$COLOR` is 2. It should check whether `$COLOR` is 2 or higher.


drwetter:
Thanks!

Issue #1294, opened 2019-07-09
[make-openssl111.sh] "enable-tls13downgrade" doesn't exist anymore


pihug12:
https://github.com/drwetter/testssl.sh/blob/3.0/utils/make-openssl111.sh#L15

`enable-tls13downgrade` doesn't exist since July 18, 2018: https://github.com/openssl/openssl/commit/9f22c527232d8babfa4827dff34a6707e8880dd9#diff-6a755e7d228a2b491321dbae0a7a248b

It makes the script/build failed.

____

And are the lines 85-87 correct (https://github.com/drwetter/testssl.sh/blob/3.0/utils/make-openssl111.sh#L85-L87) without the `-static`?

drwetter:
Hi @pihug12,

the first point I discovered too but didn't bother yet to change. A while back this file was provided as a courtesy. It's a good idea to change that now.

The second point was just my sloppiness as one distribution won't link the binary otherwise. But that line should change.

Mind providing a PR?

Cheers, Dirk

PR #1295, opened 2019-07-10
Fix "make-openssl111.sh"

pihug12:
Fix https://github.com/drwetter/testssl.sh/issues/1294


drwetter:
Thanks! Any reason you disabled ``enable-ec_nistp_64_gcc_128``? I think it doesn't matter much. For the kerberos it can't hurt either

pihug12:
`enable-ec_nistp_64_gcc_128` seems already set in `OPT11` (line 15).

PR #1307, opened 2019-08-14
Output files to host when using Docker

petegallagher:
Provides an example of how to create output files in the host filesystem when using Docker.

### DIFF
--- a/Dockerfile.md
+++ b/Dockerfile.md
@@ -11,7 +11,13 @@ You can also supply command line options like:
 ``docker run -t mytestssl -p --header example.com``
 
 Please keep in mind that any output file (--log, --html, --json etc.) will be created
-in the container. Also if you don't provide a user, this docker container uses a non-root user.
+in the container. If you wish to have this created in your local directory you can mount a volume and change the output prefix e.g.:
+
+```
+docker run -t -v `pwd`:/data drwetter/testssl.sh --html --outprefix /data/ example.com
+```
+
+Also if you don't provide a user, this docker container uses a non-root user.
 
 
 You can also pull the image from dockerhub and run:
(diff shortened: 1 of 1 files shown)


drwetter:
Hi @petegallagher ,

that is an important hint, thanks!

Could add 1-2 points though or do you want me to do that?

1) it should work with copy & paste or using place holders to explain what it is doing.
2) `pwd` works but I am preaching in the coding style guide it's deprecated. Pls use $PWD


Cheers, Dirk

petegallagher:
Hi @drwetter 

1. The command already works with copy/paste and I provided and explanation of what is happening in the sentence prior to the example. So I'm not sure what else you require here?
2. I have updated the command according to your style preferences.

Thanks,

Peter

drwetter:
Hi Peter,

thanks for the update.

You assume the container user is allowed to write into $PWD which only works with very loose (e.g. 777) permissions [1]:

```
prompt:/data/tmp/d|2% ls -la                                                                                                                                         
total 16
drwxrwxrwx 2 me root   94 Aug 29 11:37 .
drwxr-xr-x 4 root  root  100 Aug 29 11:33 ..
-rw-r--r-- 1  1000 1000 6072 Aug 29 11:38 -example.com_p443-20190829-0935.html
-rw-r--r-- 1  1000 1000 4747 Aug 29 11:38 -example.com_p443-20190829-0937.html
prompt:/data/tmp/d|% 
```

Without any further comment people will stumble over this. So at least the should get a hint. OTOH I do not want people to recommend to set _any_ directory to 777 so that they can do the docker mount, as this is security-wise a bad recommendation. So either the suggested commandline needs to be amended or a big warning needs to be there when saying it only works with 777.

I don't know what's the best is, let me sleep over it. Opinions?

Cheers, Dirk


[1]: It might work by coincidence if the local user and the container UID match.

PS: Why not picking ``--htmlfile`` instead of ``--html --outprefix``?

Issue #1321, opened 2019-09-17
[Feature Request] Support punycode if `idn` command is available


teward:
This is a **feature request** and not a bug report.

Currently, `testssl` does NOT support punycode, or the conversion of UTF8 strings *to* punycode.

However, there is *NO* guarantee that any given system will have IDN support - i.e. no installed `idn` or `libidn` to call upon.

In Ubuntu and Debian variants, we can install `idn` from the repositories, however this isn't a guarantee that this will be available under the same name.

Depending on how we wish to include IDN support, we can either go the route of:

(1) Checking if the `idn` command exists, and if it does, run the given domain through `idn` before processing it throughout the program - however if `idn` is not available or installed this **will fail hard and will continue to have issues** a la #1319 and #1320 

(2) External dependencies beyond just the `idn` program, a la a Python call to do the encode/decode.  Which has the requisite of needing Python.

(3) Try and write an IDN encoder/decoder ourselves in Bash (which sounds like the hellish approach).

Using approach 1 here may be the best approach, however do we want to have that dependency added?

If we *do* want to go the route of incorporating the `idn` command support I would like to attempt to provide some mechanism for this in the code (so please let me have some time to dig into this for a solution, my Bash is slightly rusty and I need to reaffirm myself with bits of the code...)

teward:
Fixed with #1322 merge

drwetter:
There's a problem: 

Debian Buster:
```
prompt% dig +timeout=2 +tries=2 +short -t a xn--v4h.com
dig: 'xn--v4h.com.' is not a legal IDNA2008 name (string contains a disallowed character), use +noidnout
prompt% dig +noidnout  +timeout=2 +tries=2 +short -t a xn--v4h.com
54.36.56.87
```

Debian stretch
```
prompt% dig  +timeout=2 +tries=2 +short -t a xn--v4h.com
54.36.56.87 from server 192.168.122.1 in 0 ms.
prompt% dig +noidnout  +timeout=2 +tries=2 +short -t a xn--v4h.com
Invalid option: +noidnout
```

Opensuse 15.X
```
prompt% dig +noidnout  +timeout=2 +tries=2 +short -t a xn--v4h.com
;; IDN support not enabled [on stderr]
54.36.56.87
prompt% dig +timeout=2 +tries=2 +short -t a xn--v4h.com 2>/dev/null
```


On Debian Buster ``host`` also hiccups. Drill seems to be straight forward so far. 

[...]

Issue #1320, opened 2019-09-17
fails to detect http service on port 443 when using IDN


bjmgeek:
Please make sure that you provide enough information so that we understand what your issue is about.

1. uname -a
   Linux xxxxxx 5.1.12 #1 SMP Thu Jun 20 08:36:36 EDT 2019 x86_64 GNU/Linux

2. testssl version from the banner: testssl.sh -b 2>/dev/null | head -4 | tail -2
    testssl.sh       3.0rc5 from https://testssl.sh/dev/
    (3e4cd58 2019-09-13 10:56:15 -- )

3. git log | head -1 (if running from git repo)
  3e4cd58f7b4918bab1ce2be6ee48963113040606

4. openssl version used by testssl.sh: testssl.sh -b 2>/dev/null | awk -F':' '/openssl/ { print $2}'
  $PWD/bin/openssl.Linux.x86_64

5. steps to reproduce: testssl.sh or docker command line, if possible incl. host
  testssl.sh --ip 23.236.62.147 ♨️.com

6. what exactly was happening, output is needed
Nothing happened for a while, then  
````
 Start 2019-09-17 16:37:06        -->> 23.236.62.147:443 (♨️.com) <<--

 A record via:           supplied IP "23.236.62.147"
 rDNS (23.236.62.147):   147.62.236.23.bc.googleusercontent.com.

23.236.62.147:443 doesn't seem to be a TLS/SSL enabled server
 The results might look ok but they could be nonsense. Really proceed ? ("yes" to continue) --> 
````

7. what did you expect instead?
Using the punycode name, (which still needs the ip address, see #1319 ), the correct http detection occurs.

````
 Start 2019-09-17 16:33:32        -->> 23.236.62.147:443 (xn--j6h.com) <<--

 A record via:           supplied IP "23.236.62.147"
 rDNS (23.236.62.147):   147.62.236.23.bc.googleusercontent.com.
 Service detected:       HTTP

````

drwetter:
That should work too now.

Issue #1319, opened 2019-09-17
incorrect domain resolution with IDN


bjmgeek:
1. uname -a
   Linux xxxx 5.1.12 #1 SMP Thu Jun 20 08:36:36 EDT 2019 x86_64 GNU/Linux

2. testssl version from the banner: testssl.sh -b 2>/dev/null | head -4 | tail -2
    testssl.sh       3.0rc5 from https://testssl.sh/dev/
    (3e4cd58 2019-09-13 10:56:15 -- )

3. git log | head -1 (if running from git repo)
commit 3e4cd58f7b4918bab1ce2be6ee48963113040606

4. openssl version used by testssl.sh: testssl.sh -b 2>/dev/null | awk -F':' '/openssl/ { print $2}'
$PWD/bin/openssl.Linux.x86_64

5. steps to reproduce: testssl.sh or docker command line, if possible incl. host
testssl.sh ☮️.com

6. what exactly was happening, output is needed
Fatal error: No IPv4/IPv6 address(es) for "☮️.com" available

7. what did you expect instead?
testssl.sh should run and test the server.  It should resolve the domain name as xn--v4h.com

drwetter:
thanks

teward:
Note the specified URL/IDN **is not valid** for a URL.  This is confirmed by dig, etc.

bjmgeek:
> Note the specified URL/IDN is not valid for a URL. This is confirmed by dig, etc.

And yet, it works in Firefox and Chrome, as well as curl:
````
 curl -v http://xn--v4h.com
* Uses proxy env variable no_proxy == 'localhost,127.0.0.0/8,::1'
*   Trying 54.36.56.87:80...
* TCP_NODELAY set
* Connected to xn--v4h.com (54.36.56.87) port 80 (#0)
> GET / HTTP/1.1
> Host: xn--v4h.com
> User-Agent: curl/7.65.3
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 301 Moved Permanently
< Cache-Control: public, must-revalidate, proxy-revalidate, max-age=3600
< Location: https://www.geniusofpeace.org/
< Content-type: text/html; charset=UTF-8
< Content-Length: 0
< Date: Wed, 18 Sep 2019 20:21:52 GMT
< X-IPLB-Instance: 16677
< 
* Connection #0 to host xn--v4h.com left intact
````

teward:
[...]

PR #1322, opened 2019-09-17
Add IDN/punycode support for non-ASCII URIs (#1319, #1320, #1321)

teward:
This should address the issues #1319 and #1320 *if* `idn` is available on the system used to test those cases, and my corresponding feature request in #1321 

This takes approach number 1 for #1321 and does several tests to determine how to handle the entered URI:

1. We test if there are non-ASCII characters in the provided URI.  If there are not any non-ASCII characters in the provided URI, then we accept the URI as entered and continue on, processing the URI as we would normally in the 3.0 branch.  Otherwise, we continue on, as we have a second test to run before we can handle the IDN and convert to punycode.

2. If the test revealed that there are non-ASCII characters in the URI, the next test we do is to see if the `idn` binary is available, using Bash's `command -v` call.

    1. If `idn` is available, then `URI` is set to the output of the command `echo $1 | idn`, which will return punycode.  In the case of #1319 and #1320, using their example of `♨️.com`this properly will detect the non-ASCII characters and convert to Punycode, and then properly handle the URI as `xn--j6h.com`.

    2. If `idn` is not available, then error with a Fatal call indicating that the URI contains non-ASCII characters and there is no IDN handler available to convert to Punycode.

This can be tested *directly* at my fork of this through https://github.com/teward/testssl.sh/tree/idn-support which contains the code in this PR.  Just check out the `idn-support` branch.  That branch was forked from the `3.0` branch here.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -19340,7 +19340,15 @@ parse_cmd_line() {
           fatal "URI missing" $ERR_CMDLINE
      else
      # left off here is the URI
-          URI="$1"
+          if [[ $1 = *[![:ascii:]]* ]]; then
+              if [[ "$(command -v idn)" == "" ]]; then
+                  fatal "URI contains non-ASCII, IDN not available."
+              else
+                  URI="$(echo $1 | idn)"
+              fi
+          else
+              URI="$1"
+          fi
           # parameter after URI supplied:
           [[ -n "$2" ]] && fatal "URI comes last" $ERR_CMDLINE
      fi
--- a/testssl.sh
+++ b/testssl.sh
@@ -19342,7 +19342,7 @@ parse_cmd_line() {
      # left off here is the URI
           if [[ $1 = *[![:ascii:]]* ]]; then
               if [[ "$(command -v idn)" == "" ]]; then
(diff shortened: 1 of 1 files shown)


teward:
Note that I did not write tests for this when I committed to it, so tests will need to be written if we want to verify that IDN support does in fact exist/work as part of the unittests.

drwetter:
@teward: first feedback: Thanks, cool idea!

drwetter:
Thank you @teward!

drwetter:
a small unit test is always appreciated (wink).

Code was fine, too, thanks. For consistency reasons though I will change later ``command`` to ``type -p`` and ``idn "$1" should do it better (saves one subshell)

teward:
> a small unit test is always appreciated (wink).
> 
> Code was fine, too, thanks. For consistency reasons though I will change later `command` to `type -p` and ``idn "$1" should do it better (saves one subshell)

Yep, all's good.

I'm still wrapping my head around the structure of the unit tests here, which is why I didn't write a test yet :)  This said, it shouldn't be *too* hard to do...

PR #1325, opened 2019-09-18
remove duplicate header name for other_header in file output

Phylu:
The Finding of other_headers such as "Referrer-Policy" during file output are displayed as `$header: $HEADERVALUE` instead of only `$HEADERVALUE` as the good_headers. This leads to duplicate information e.g. in the JSON output file.


drwetter:
good catch, thanks!

PR #1327, opened 2019-09-20
Idn improvements

drwetter:
see each comment

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -17453,7 +17453,9 @@ check_resolver_bins() {
 get_a_record() {
      local ip4=""
      local saved_openssl_conf="$OPENSSL_CONF"
+     local noidnout=""
 
+     [[ "$HAS_DIG_NOIDNOUT" ]] && noidnout="+noidnout"
      [[ "$NODNS" == none ]] && return 0      # if no DNS lookup was instructed, leave here
      if [[ "$1" == localhost ]]; then
           # This is a bit ugly but prevents from doing DNS lookups which could fail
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -17384,11 +17425,26 @@ get_local_a() {
...
 }
 
-# does a hard exit if no lookup binary is provided
+# Does a hard exit if no lookup binary is provided
+# Checks for IDN capabilities also
+#
 check_resolver_bins() {
-     if ! type -p dig &> /dev/null && ! type -p host &> /dev/null && ! type -p drill &> /dev/null && ! type -p nslookup &>/dev/null; then
+     type -p dig   &> /dev/null &&  HAS_DIG=true
+     type -p dig   &> /dev/null &&  HAS_HOST=true
```
teward: Think you meant this to be `host`

drwetter: done

testssl.sh
```diff
@@ -355,7 +355,16 @@ HAS_CHACHA20=false
...
+HAS_DIG=false
+HAS_HOST=false
+HAS_DRILL=false
+HAS_NSLOOKUP=false
+HAS_IDN=false
+HAS_IDN2=false
+HAS_AVAHIRESOLVE=false
+HAS_DIG_NOIDNOUT=false
+
+ OSSL_CIPHERS_S=""
```
teward: Style: Extra space in the beginning is probably not supposed to be here, so it's all on one line.  Just being nitpicky at this point, but still.

drwetter: :-)

done

PR #1329, opened 2019-09-20
Fix run_protocols()

dcooper16:
This PR fixes a problem in `run_protocols()` that was introduced by https://github.com/drwetter/testssl.sh/commit/7ec3c6ab997b012b1518903137d84043cf9825a8.

https://github.com/drwetter/testssl.sh/commit/7ec3c6ab997b012b1518903137d84043cf9825a8 changes `run_protocols()` to perform the initial testing for TLSv1.3 support before testing for TLSv1.2 support. The problem with this is that the code for testing TLSv1.3 makes use of the results of the TLSv1.2 testing.

In the current code, Line 5183 looks at the value of `$subret` to determine whether the TLSv1.2 ClientHello resulted in a successful connection. However, `$subret` has not yet been set (it has just been initialized to 0 at the beginning of the function). Since `$subret` will always be 0, the code will try to extract a cipher from `$TEMPDIR/$NODEIP.parse_tls_serverhello.txt`. This may work, since `$TEMPDIR/$NODEIP.parse_tls_serverhello.txt` may have been populated earlier (perhaps before `run_protocols()` was called), but this is not how the code was intended to work.

This PR fixes the problem by doing the TLSv1.2 testing before the TLSv1.3 testing is done. It still waits until both have been tested, however, before outputting the results, so that the output for TLSv1.2 can be modified depending on whether TLSv1.3 is supported.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4866,7 +4866,7 @@ run_protocols() {
      local lines nr_ciphers_detected
      local tls13_ciphers_to_test=""
      local i drafts_offered=""  drafts_offered_str="" supported_versions debug_recomm=""
-     local -i ret=0 subret=0 ret_val_tls13=0
+     local -i ret=0 ret_val_tls12=0 ret_val_tls13=0
      local offers_tls13=false
      local jsonID="SSLv2"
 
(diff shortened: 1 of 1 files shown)


dcooper16:
While I didn't make a change in this PR, I don't agree with the change that https://github.com/drwetter/testssl.sh/commit/7ec3c6ab997b012b1518903137d84043cf9825a8 made in the case that the TLSv1.2 test returns 2 (lines 5243 - 5247). To get to this point, it means that a TLSv1.2 ClientHello resulted in a TLSv1.1 or earlier connection.

It seems _highly_ unlikely that a server would support TLSv1.3 and TLSv1.1 (or TLSv1 or SSLv3) and not support TLSv1.2. However, if that did happen, I believe we should flag that as a problem. A TLSv1.3-only server that reject TLSv1.2 ClientHello messages may be okay. But, a server that responds to a TLSv1.2 ClientHello by downgrading to TLSv1.1 (or TLSv1 or SSLv3) seems bad, even if that server supports TLSv1.3.

drwetter:
Hi David,

great hearing from you, thanks for reviewing + PR!

You're right regarding the mixture of TLS 1.3 and lower protocols. Feel free to add a PR. Also regarding the downgrade.

Thx, Dirk

PR #1330, opened 2019-09-23
More run_protocol() fixes

dcooper16:
This PR fixes a few issues with `run_protocol()`:

* In the case that the call to `tls_sockets "03" "$TLS12_CIPHER"` had a return value of 2, the code determining what results to print was looking at `$DETECTED_TLS_VERSION`. However, the value of this variable was set by the later call to `tls_sockets "04" "$tls13_ciphers_to_test"`. This caused incorrect results in the case of a server that supports TLSv1.3 and TLS1.1 (or earlier), but not TLSv1.2. This PR saves the value of `$DETECTED_TLS_VERSION` in `$tls12_detected_version` and then uses this variable later rather than `$DETECTED_TLS_VERSION`.

* When running in debug mode with a server that does not support TLSv1.3, testssl.sh was printing

           TLS 1.3     -- downgradednot offered and downgraded to a weaker protocol"

    This PR fixes the output by not printing the "--downgraded"

* As noted in #1329, run_protocols() was treating a downgrade from TLSv1.2 as less bad if the server supports TLSv1.3. This PR changes this code back to treat any downgrade from TLSv1.2 as equally bad.

* In order to be consistent with the TLSv1.3 test, this PR changes the TLS1.2 test output to say "not offered and downgraded to a weaker protocol" if a TLSv1.2 ClientHello results in a downgraded connection.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4866,6 +4866,7 @@ run_protocols() {
      local lines nr_ciphers_detected
      local tls13_ciphers_to_test=""
      local i drafts_offered=""  drafts_offered_str="" supported_versions debug_recomm=""
+     local tls12_detected_version
      local -i ret=0 ret_val_tls12=0 ret_val_tls13=0
      local offers_tls13=false
      local jsonID="SSLv2"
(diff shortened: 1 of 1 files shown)


drwetter:
Look good, thank you!

PR #1332, opened 2019-09-25
Check ignore_no_or_lame() return value

dcooper16:
In the calls to i`gnore_no_or_lame()` at the end of `run_protocols()` the return value is not checked and the program continues running even if the user asked it to stop.


dcooper16:
I just noticed a second problem with the code at the end of `run_protocols()`, and so added a fix for it to this PR.

If the `[[ -x /usr/bin/openssl ]]` test fails, then the code moves to the `else` case, just as if /usr/bin/openssl existed and supported TLSv1.3. Then if `$OSSL_SHORTCUT` is true, testssl.sh will try to continue on using the non-existent /usr/bin/openssl.

drwetter:
Thanks, David. WRT: Ignore or lame: yes. The other: Looks you're right. Take care next few days

drwetter:
closing, see #1336

PR #1334, opened 2019-09-30
Reset OpenSSL dependent variables

dcooper16:
There is a problem if a TLSv1.3-only server is tested using the OpenSSL 1.0.2-chacha binary and `$OSSL_SHORTCUT` is true.

`$HAS_NO_SSL2` is set to `true` when `find_openssl_binary()` is called with OpenSSL 1.0.2-chacha. `/usr/bin/openssl` does not have the `-no_ssl2` option, but the second call to `find_openssl_binary()`, after setting `$OPENSSL` to `/usr/bin/openssl`, does not set `$HAS_NO_SSL2` to `false`. So, later calls to `$OPENSSL s_client` include the `-no_ssl2` option, resulting in connection failures.

This PR fixes the problem by modifying `find_openssl_binary()` to ensure that every OpenSSL-dependent variable is set by this function.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -16597,6 +16597,35 @@ find_openssl_binary() {
          OPENSSL_LOCATION="$openssl_location"
      fi
 
+     OSSL_CIPHERS_S=""
+     HAS_SSL2=false
+     HAS_SSL3=false
+     HAS_TLS13=false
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks!

PR #1335, opened 2019-09-30
Fix session ticket issues

dcooper16:
This PR fixes two issues with finding session tickets when using OpenSSL 1.1.1.

First, if OpenSSL connects to the server using TLSv1.3 and it receives more than one Post-Handshake New Session Ticket, then the "TLS session ticket lifetime hint" will appear more than once in `$TMPFILE`. This will cause the line to appear more than once in `$sessticket_lifetime_hint`, which causes problems when trying to extract the `$lifetime` and `$unit` from `$sessticket_lifetime_hint`.

This PR fixes the first problem by changing the `awk` expression in the lines that set sessticket_lifetime_hint so that only the first line with "session ticket lifetime" is extracted.

The second issue is that some servers (e.g., google.com) return a session ticket for TLSv1.2, but not for TLSv1.3. For such servers, testssl.sh will miss the session ticket if `$OPTIMAL_PROTO` is empty or `-tls1_3` and the `--ssl-native` flag is not set.

This PR addresses the second issue with the changes in lines 9047 - 9053 -- the code that is intended to provide a last chance to find a session ticket.

If `$OPENSSL` supports TLSv.1.3 and the server returns session tickets for TLSv1.3 connections, then the session ticket would have already been found by `get_server_certificate()`, since `get_server_certificate()` uses `$OPENSSL` for TLSv1.3, if `$OPENSSL` supports TLSv1.3. So, in such circumstances, the code in liens 9047 - 9053 should not try again with TLSv1.3. So, if `$OPENSSL` supports TLSv1.3 and `$OPTIMAL_PROTO` is empty or is set to `-tls1_3` (either of which would result in a TLSv1.3 ClientHello), the `$OPENSSL s_client` call is changed to specify -no_tls1_3 rather than `$OPTIMAL_PROTO`.

The code on line 9047 is also changed to only make this final try is `$TLS13_ONLY` is false. If `$TLS13_ONLY` is true, then either:
* `$OPENSSL` does not support TLSv1.3 and the connection attempt would fail anyway; or
* `$OPENSSL` supports TLSv1.3, in which case any session ticket would have been found by `get_server_certificate()`, since `get_server_certificate()` uses `$OPENSSL` for TLSv1.3 if `$OPENSSL` supports TLSv1.3.

In either case, there is no reason to try again to find a session ticket.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -8889,7 +8889,7 @@ run_server_defaults() {
                     cp "$TEMPDIR/$NODEIP.get_server_certificate.txt" $TMPFILE
                     >$ERRFILE
                     if [[ -z "$sessticket_lifetime_hint" ]]; then
-                         sessticket_lifetime_hint=$(awk '/session ticket life/' $TMPFILE)
+                         sessticket_lifetime_hint=$(awk '/session ticket life/ { if (!found) print; found=1 }' $TMPFILE)
                     fi
 
                     if [[ $n -le 7 ]]; then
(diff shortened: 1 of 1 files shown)


drwetter:
Cool, thanks for your analysis and your PR

PR #1336, opened 2019-09-30
Use OpenSSL for determine_optimal_proto()

dcooper16:
This PR reverts `determine_optimal_proto()` to use OpenSSL again rather than `tls_sockets()`.

The primary reason for this is that the primary purpose of `determine_optimal_proto()` is to set `OPTIMAL_PROTO`, which is only used with `$OPENSSL s_client`. So, the best way to determine what works best on the `$OPENSSL s_client` command line is to use `$OPENSSL s_client`.

In most cases, `determine_optimal_proto_sockets_helper()` would set `OPTIMAL_PROTO` to an acceptable value, but it might not always do so. For example, suppose that a server
* supports different cipher suites with different protocols, 
* supports TLSv1.2, but only with cipher suites not supported by `$OPENSSL`, but
* supports TLSv1.1 with at least one cipher suite supported by `$OPENSSL`.

In the above case, `determine_optimal_proto_sockets_helper()` would set `OPTIMAL_PROTO` to "-tls1_2", but testing using `$OPENSSL` would result in `OPTIMAL_PROTO` being set to "-tls1_1".

Using `$OPENSSL` for `determine_optimal_proto()` also allows for edge cases to be detected earlier:
* If the server only supports TLSv1.3, and `$OPENSSL` does not support TLSv1.3, then the code in this PR will detect that (rather than waiting until `run_protocols()` is executed).
* The code in this PR can also detect if the server only supports SSLv3 (and possibly also SSLv2), but `$OPENSSL` does not support SSLv3.
* This code can also detect the (rare) case in which connections using `$OPENSSL` succeed, but connections using `tls_sockets()` fails.

[Note also that in the current code, if `$all_failed` is true, then a message may be printed that `$OPENSSL` is not IPv6 aware, even if testing was performed using `tls_sockets()` rather than `$OPENSSL`.]


dcooper16:
If you are okay with this PR, then you don't need to bother with #1332. If this PR is merged, then I'll propose to delete the code at the end of `run_protocols()`.

drwetter:
Ok, thanks for the info. Don't know I'll get to it today

dcooper16:
Hi Dirk,

While working on #1337 I noticed another issue with the current `determine_optimal_proto()`. The function `sclient_auth()` is only called by `determine_optimal_proto()`, and only if `$using_sockets` is true. `sclient_auth()` in turn is the only function that sets `$CLIENT_AUTH` and `$NO_SSL_SESSIONID`. So, at the moment, `$CLIENT_AUTH` and `$NO_SSL_SESSIONID` are not set properly unless "--ssl-native" is used.

drwetter:
okay... :-( but thanks

testssl.sh
```diff
@@ -17982,91 +17922,105 @@ determine_optimal_proto_sockets_helper() {
...
+          TLS13_ONLY=true
+     fi
 
-     if "$all_failed"; then
+     if [[ "$optimal_proto" == -ssl2 ]]; then
+          prln_magenta "$NODEIP:$PORT appears to only support SSLv2."
+          ignore_no_or_lame " Type \"yes\" to proceed and accept false negatives or positives" "yes"
+          [[ $? -ne 0 ]] && exit $ERR_CLUELESS
+     elif "$all_failed" && ! "$ALL_FAILED_SOCKETS"; then
+          if ! "$HAS_TLS13" && "$TLS13_ONLY"; then
```
drwetter: My original idea here was when testssl.sh discovers a scenario that   

if the server side offers TLS 1.3 only && the used $OPENSSL doesn't support that && /usr/bin/openssl does that an switch is taking place like

```
 Testing protocols via sockets except NPN+ALPN

 SSLv2      not offered (OK)
 SSLv3      not offered (OK)
 TLS 1      not offered
 TLS 1.1    not offered
 TLS 1.2    not offered
 TLS 1.3    offered (OK): final
 172.17.0.3:443 appears to support TLS 1.3 ONLY. You better use --openssl=<path_to_openssl_supporting_TLS_1.3>
 Type "yes" to proceed and accept all scan problems -->
```
This hack doesn't work any more. I'll look into it

PR #1337, opened 2019-10-01
Fix sub_session_resumption()

dcooper16:
This PR fixes an issue with `sub_session_resumption()` when using OpenSSL 1.1.1.

As noted in #1335, some servers will return a session ticket for TLSv1.2, but not for TLSv1.3.

OpenSSL 1.1.1 does not support the "-no_ssl2" option, and so when using OpenSSL 1.1.1 `sub_session_resumption()` adds `$OPTIMAL_PROTO` to the `$OPENSSL s_client` command line. When `determine_optimal_proto_sockets_helper()` is called, `$OPTIMAL_PROTO` will generally be set to "-tls1_2" (or "-tls1_1" or "-tls1") unless the server is a TLSv1.3-only server. As a result  `sub_session_resumption()` will specify that same protocol on the command line, if OpenSSL 1.1.1 is being used.

If "--ssl-native" is used, however, then `determine_optimal_proto()` will set `$OPTIMAL_PROTO` to "-tls1_3" if the server supports TLSv1.3 (and doesn't use STARTTLS). Similarly, if the version of `determine_optimal_proto()` in #1336 is used, then `$OPTIMAL_PROTO` will usually be empty. In either case, `sub_session_resumption()` will send a TLSv1.3 ClientHello, even if the server only supports session tickets for TLSv1.2 and below.

This PR appears to fix the problem. This PR makes no changes when using a version of OpenSSL that supports "-no_ssl2". When using a version of OpenSSL that does not support "-no_ssl2", however, rather than using `$OPTIMAL_PROTO`, this PR has `sub_session_resumption()` use whatever protocol version the server connected with when `$sessticket_lifetime_hint` was set.

For an example of the problem, try
```
testssl.sh --openssl=/usr/bin/openssl --ssl-native -S testssl.net
```

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6068,10 +6068,21 @@ sub_session_resumption() {
           fi
      fi
      "$CLIENT_AUTH" && return 3
-     protocol=${protocol/\./_}
-     protocol=${protocol/v/}
-     protocol="-$(tolower $protocol)"
-     "$HAS_NO_SSL2" && addcmd+=" -no_ssl2" || addcmd+=" $protocol"
+     if "$HAS_NO_SSL2"; then
+          addcmd+=" -no_ssl2"
+     else
+          protocol=${protocol/\./_}
+          protocol=${protocol/v/}
+          protocol="-$(tolower $protocol)"
+          # In some cases a server will not support session tickets, but will support session resumption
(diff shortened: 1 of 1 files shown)


drwetter:
Thank you, David

PR #1339, opened 2019-10-02
Simplify determine_sizelimitbug()

dcooper16:
This PR takes advantage of the testing done by `determine_optimal_sockets_params()` in order to simplify `determine_sizelimitbug()`.

By the time that `determine_sizelimitbug()` is called, `determine_optimal_sockets_params()` has already determined whether TLSv1.2 ClientHello with 128 ciphers (including 00FF) sent by `tls_sockets()` works, and it has set `$TLS12_CIPHER` to a list of exactly 128 ciphers (including 00FF) that works with the server. So, `determine_sizelimitbug()` doesn't have to check whether the server supports TLSv1.2 and no longer needs to send tests using 127 or 128 ciphers. `determine_sizelimitbug()` can just perform one test with 129 ciphers, if the server supports TLSv1.2, and use the results to set `$SERVER_SIZE_LIMIT_BUG`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -18193,68 +18193,39 @@ determine_service() {
 # Return value is 0 unless we have a problem executing
 #
 determine_sizelimitbug() {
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -18193,68 +18193,39 @@ determine_service() {
...
-          debugme echo "Sending 127 ciphers"
-          tls_sockets 03 "${test_ciphers}, 00,FF"
-          if [[ $? -eq 0 ]]; then
-               debugme echo "Sending 128 ciphers"
-               tls_sockets 03 "${test_ciphers}, ${overflow_cipher1}, 00,FF"
-               if [[ $? -ne 0 ]]; then
-                    SERVER_SIZE_LIMIT_BUG=true
+     else
+          if [[ "$DEBUG" -ge 1 ]]; then
+               nr_ciphers="$(tr ' ' '\n' <<< "${overflow_cipher}, $TLS12_CIPHER" | sed -e '/^$/d' | wc -l)"
```
drwetter: can you please check whether this works under Darwin? IIRC FreeBSD+Darwin add a blank sometimes with ``wc -l``. ``count_lines()`` would be preferred otherwise.

It worked for me on MacOS 10.14.6. `$TLS12_CIPHER` includes a blank line, so I add to add `sed -e '/^$/d'` to remove any blank lines before counting. Perhaps that also prevents any problems with FreeBSD+Darwin

drwetter:
Thanks, David! Didn't see we used ~it before.

Note to self: At some time I should use a separate count function ``count_ciphers()`` does that for openssl formatted ciphers.

PR #1341, opened 2019-10-04
Use determine_optimal_sockets_params() in run_protocols()

dcooper16:
This PR modifies `run_protocols()` to use the information collected by `determine_optimal_sockets_params()`. If it has already been determined that a protocol is supported, then no test is run by `run_protocols()` for that protocol. `run_protocols()` will still run a test for a protocol even if it has been determined that the server does not support that protocol. The reason for running the test is to verify that the server handles version negotiation correctly. This could be a TLSv1 server that rejects a TLSv1.2 or TLSv1.3 ClientHello, or it could happen in the opposite direction. At one point there was a server that would respond to an SSLv3 ClientHello with a TLSv1.2 ServerHello.

This PR required a couple of changes to `determine_optimal_sockets_params()` so that additional information could be passed to `run_protocols()`. If the server supports TLS 1.3, then `run_protocols()` needs to know which version (RFC 8446, draft 28, draft 27, etc.) rather than just that TLS 1.3 is supported. If the server supports TLS 1.2, but not TLS 1.3, then `run_protocols()` needs to know about at least one TLS 1.2 cipher that the server supports so that it can form a TLS 1.3 ClientHello that has no more than 128 ciphers and that should result in the server returning a TLS 1.2 ServerHello.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -285,6 +285,7 @@ NR_SOCKET_FAIL=0                        # Counter for socket failures
 NR_OSSL_FAIL=0                          # .. for OpenSSL connects
 NR_HEADER_FAIL=0                        # .. for HTTP_GET
 PROTOS_OFFERED=""                       # This keeps which protocol is being offered. See has_server_protocol().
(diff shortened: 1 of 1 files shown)


drwetter:
cool, thank you!

PR #1344, opened 2019-10-10
Add --connect-timeout parameter

goncalor:
Currently there is an `--openssl-timeout` option. However, there is no option to control TCP handshake timeouts. When testing many hosts if any of them is unexpectedly unresponsive (TCP 443 port closed, host down, behind a firewall, etc.) `testssl.sh` will hang for about 2 minutes for each host.

This PR adds the parameter `--connect-timeout` to limit the maximum amount of time to wait for the TCP handshake to complete successfully. The default value is 3 minutes, so that current users of `testssl.sh` should not be affected by this change.

I tested this feature and it seems to be working as intended.


drwetter:
Thanks @goncalor,

that's a good idea! The PR needs to wait though until next week. There are a few things which need to be added.

Do you want me to go through that with you (review feature here) or should I just amend the PR later?

Cheers, Dirk

goncalor:
Hi @drwetter,

Thank you for the quick reply. :slightly_smiling_face:  I'm fine with you ammending the PR when you can. I will check what you had to add afterwards and learn from it.

Cheers, Gonçalo

drwetter:
Thanks and sorry for the delay.

I merged and pushed this to the _connect-timeout_ branch, see https://github.com/drwetter/testssl.sh/tree/connect-timeout where the work continuous. 

I had this almost ready this week but there was a misunderstanding between me and git ;-)

drwetter:
see #1408. Thanks again.

goncalor:
You're welcome. Also thank you for your extra work on it!

PR #1346, opened 2019-10-16
Update Safari to 13.0 and macOS to 10.14

csett86:
manually wiresharked, now with TLS1.3 for macOS as well.


drwetter:
Thank you!

drwetter:
How about iOS? ;-)

csett86:
Nothing has changed in Safari on iOS, TLS 1.3 was already working with the iOS version in it. But maybe I‘ll doublecheck ATS...

drwetter:
Even if it hasn't changed: Feel free to submit that with another label.

I am not really up to date with versions in the Mac world. For Android I looked at the usage and included a reasonable old version and newer versions which are still in use.

If you or anybody else feel we should drop e.g. "Safari 10 OS X 10.12" to be listed by default because nobody's using it pls. let me know.

PR #1351, opened 2019-10-23
'IPv4 address in header' shows body content

tkaehn:
'IPv4 address in header' shows body content if there is no match with the specified patterns.
Probably the HEADERFILE should end on the first empty line.

Please check if this works for all supported platforms or if there are any side effects caused by this approach before merging.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2175,10 +2175,8 @@ run_http_header() {
      # Populate vars for HTTP time
      debugme echo "$NOW_TIME: $HTTP_TIME"
 
-     # delete from pattern til the end. We ignore any leading spaces (e.g. www.amazon.de)
-     sed -e '/<HTML>/,$d' -e '/<html>/,$d' -e '/<\!DOCTYPE/,$d' -e '/<\!doctype/,$d' \
-         -e '/<XML/,$d' -e '/<xml/,$d' -e '/<\?XML/,$d' -e '/<?xml/,$d' $HEADERFILE >$HEADERFILE.tmp
-         # ^^^ Attention: filtering is for html body only as of now, doesn't work for other content yet
+     # Quit on first empty line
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks. 

It would be definitely more elegant, thanks. Should work also for non-GNUish sed, see https://edoras.sdsu.edu/doc/sed-oneliners.html.
I'll check.

But the current check should for most cases stop after the header too, right?

tkaehn:
Thanks for your quick response. I assumed that HEADERFILE is supposed to contain only the header for further checks done by testssl.sh. So quitting on the first empty line should be better than hoping to see e.g. an HTML/XML tag. 
However I do not know if ^$ matches empty lines for all operating systems and web servers. Maybe in some cases something like ^\r?$ might be required which is probably not compatible with all versions of sed. Alternatively both methods could be chained - first try to quit on the first empty line - in case this fails proceed with looking for HTML/XML tags.
To my mind testssl.sh shouldn't rely solely on searching for HTML/XML tags as some web servers might only respond with JSON or someone might want to test URLs pointing to e.g. image data.

drwetter:
I am afraid your pattern doesn't work

drwetter:
Thanks. Can't check cases with ``^\r$`` right now but with your change first and the former filter followed it's more robust than before. As indicated  ``'/^$/q'`` works under BSD also and doesn't appear GNUish to me. 

On regular servers '/^$/ didn't work. I added a pattern there.

> To my mind testssl.sh shouldn't rely solely on searching for HTML/XML tags as some web servers might only respond with JSON or someone might want to test URLs pointing to e.g. image data.

Yes. There was a comment. I put this pattern in again and changed that slightly out of date line I did ages ago by something which is better.

Still, if somebody provides an URL which has no HTML or JSON content and not a blank line following the HTTP header, this part won't work. For the time being it is acceptable to me but I also accept PRs for this ;-)

PR #1352, opened 2019-10-23
Fix client simulation bug

dcooper16:
This PR fixes a bug in `modify_clienthello()` that occurs when client simulation is being performed, the ClientHello contain an SNI extension, and `$SNI` is empty. In the case, `modify_clienthello()` should just skip over the SNI extension and not include one in the modified ClientHello. However, the code currently only skips over the 2-byte extension type. The result being that the remainder of the extension is included in the modified ClientHello. This PR fixes the problem by ensuring the `$offset` is advanced whether or not `$SNI` is empty.


drwetter:
Thanks for catching it, David!

Issue #1353, opened 2019-10-26
SSLv3 check fail 


uwe-schwarz:
Please make sure that you provide enough information so that we understand what your issue is about.

1. uname -a
Linux app 4.15.0-64-generic #73-Ubuntu SMP Thu Sep 12 13:16:13 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
   
2. testssl version from the banner: testssl.sh -b 2>/dev/null | head -4 | tail -2
testssl.sh       3.0rc5 from https://testssl.sh/dev/

3. git log | head -1 (if running from git repo)
commit f5c3b4e41da3ded6431b9e3562e2d53f57932346

4. openssl version used by testssl.sh: testssl.sh -b 2>/dev/null | awk -F':' '/openssl/ { print $2}'
./bin/openssl.Linux.x86_64

5. steps to reproduce: testssl.sh or docker command line, if possible incl. host
testssl.sh sip.dvag.com

6. what exactly was happening, output is needed
```
 Testing protocols via sockets except NPN+ALPN

 SSLv2      not offered (OK)
 SSLv3      ^C
```
It just hangs when testing for SSLv3 (I waited a few hours, because this hang when running over night). It is working with --ssl-native.

7. what did you expect instead?
Not hanging, passing the test.

drwetter:
Can reproduce it, thanks.

```
SSLv3      
sending client hello... sending client hello... 
[...]

PR #1356, opened 2019-10-28
Fix parse_tls_serverhello() bug

dcooper16:
This PR fixes a minor bug in `parse_tls_serverhello()`. In some cases the server's entire response is not retrieved. In these cases, it is possible that the response from the server ends with a portion of a handshake message.

The loop at the beginning of `parse_tls_serverhello()` extracts the various handshake and alert messages from the server's response. If it gets to the end of the response, and what is at the end is not a small (< 10 byte) fragment of a message, it should just ignore that fragment and `break` out of the loop. At the moment, however, `parse_tls_serverhello()` just continues in the loop rather than breaking out. This has not been a problem up to now, since `$msg_len` is usually set to a positive value from a previous iteration of the loop, which causes the loop to end.

In the case of the server identified in #1353, however, `$msg_len` is 0 and so the `continue` rather than `break` results in an endless loop.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -11527,7 +11527,7 @@ parse_tls_serverhello() {
                else
                     # This could just be a result of the server's response being
                     # split across two or more packets.
-                    continue
+                    break
                fi
           fi
           tls_content_type="${tls_hello_ascii:i:2}"


drwetter:
Thanks!

It stops the loop, yes. Any idea why openssl returns an alert for this server? (not indicating that it must be fixed as the server looks to me as a rare species)

dcooper16:
When I try sending an SSLv3 ClientHello with OpenSSL I get the same results as with `tls_sockets()`. Below is what I see on the command line, and Wireshark shows the the server is sending a large number of null bytes.
```
 ./bin/openssl.Linux.x86_64 s_client -connect 185.64.159.72:443 -servername sip.dvag.com -ssl3 < /dev/null
CONNECTED(00000003)
40482880:error:1408F10B:SSL routines:SSL3_GET_RECORD:wrong version number:s3_pkt.c:365:
---
no peer certificate available
---
---
SSL handshake has read 5 bytes and written 7 bytes
---
New, (NONE), Cipher is (NONE)
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : SSLv3
    Cipher    : 0000
    Session-ID: 
    Session-ID-ctx: 
    Master-Key: 
    Key-Arg   : None
    PSK identity: None
    PSK identity hint: None
    SRP username: None
    Start Time: 1572289999
    Timeout   : 7200 (sec)
    Verify return code: 0 (ok)
---
```

drwetter:
-msg shows the tls alert

drwetter:
PS: I didn't use -servername
-- 

dcooper16:
Okay, now I am seeing it. However, what I am seeing is that `openssl s_client` sends an SSLv3 ClientHello, the server responds with a bunch of null bytes, and then the _client_ responds with a handshake_failure alert.

drwetter:
yes, thanks. Overlooked that.

PR #1358, opened 2019-10-28
Use $jsonID rather than literal string

dcooper16:
In `run_drown()`, `$jsonID` is set to "DROWN" and most calls to `fileout()` are of the form
```
fileout "$jsonID" ...
```
However, one call is written as
```
fileout "DROWN" ...
```
This PR changes this one call to be consistent with the others. This does not change the functionality of the program.


drwetter:
Thanks!

PR #1364, opened 2019-11-02
fix bug in f5_ip_oldstyle() printf format string

scottj:
The printf format string in `f5_ip_oldstyle()` is incorrect. It works most of the time, but in cases where the hexadecimal output is less than 8 characters, it fails.


drwetter:
Thank you!

PR #1369, opened 2019-11-06
Fix issues with run_protocols() in --ssl-native mode

dcooper16:
This PR fixes a minor problem with `run_protocols()` in "--ssl-native" mode if `$OPENSSL` does not support TLS 1.3. Currently, the warning message that `$OPENSSL` does not support a protocol is printed when `run_prototest_openssl()` is called. This causes a problem for the output if `$OPENSSL` does not support TLS 1.3, since the `run_prototest_openssl()` is called before the results for TLS 1.2 are printed. The result is something like this:
```
 SSLv2      not offered (OK)
 SSLv3      not offered (OK)
 TLS 1      offered (deprecated)
 TLS 1.1    offered (deprecated)
Local problem: /home/cooper/Desktop/testssl.sh/bin/openssl.Linux.x86_64 doesn't support "s_client -tls1_3"
 TLS 1.2    offered (OK)
 TLS 1.3     NPN/SPDY   not offered
 ALPN/HTTP2 http/1.1 (offered)
```
This PR moves the printing of the warning message to run_protocols() in order to fix the problem.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4802,12 +4802,18 @@ run_prototest_openssl() {
      sclient_connect_successful $? $TMPFILE
      ret=$?
      debugme grep -E "error|failure" $ERRFILE | grep -Eav "unable to get local|verify error"
-     # try again without $PROXY
-     $OPENSSL s_client $(s_client_options "-state $1 $STARTTLS $BUGS -connect $NODEIP:$PORT $SNI") >$TMPFILE 2>&1 </dev/null
-     sclient_connect_successful $? $TMPFILE
-     ret=$?
-     debugme grep -E "error|failure" $ERRFILE | grep -Eav "unable to get local|verify error"
-     grep -aq "no cipher list" $TMPFILE && ret=5       # <--- important indicator for SSL2 (maybe others, too)
+     if [[ $ret -ne 0 ]]; then
+          if grep -aq "no cipher list" $TMPFILE; then
+               ret=5       # <--- important indicator for SSL2 (maybe others, too)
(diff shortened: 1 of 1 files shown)


dcooper16:
I just added a few more commits that address issues related to `run_protocols()` in `--ssl-native` mode. Most of the issues are in the `run_prototest_openssl()` function.

- The first commit just fixes a couple of typos in a comment.
- The second commit fixes run_prototest_openssl()'s ability to detect when the server supports SSLv2, but does not support any SSLv2 ciphers.
- The third commit fixes a missing line break if the server supports SSLv2, but does not support any SSLv2 ciphers, and `--ssl-native` is being used.
- The fourth commit fixes a problem in `run_prototest_openssl()` where `$OPENSSL s_client` is being called twice and the results of the first call are ignored.

In the case of the fourth commit, it is not clear why `run_prototest_openssl()` needs to try calling `$OPENSSL s_client` twice. If the attempt to connect to the server using `$PROXY` is unsuccessful, is it necessary to try again without `$PROXY`? This doesn't seem to be done anywhere else?

drwetter:
Thanks, David!

drwetter:
> If the attempt to connect to the server using $PROXY is unsuccessful, is it necessary to try again without $PROXY? This doesn't seem to be done anywhere else?

Can't tell anymore. I suspect it had something to do with the protocol but a) the use case of a proxy normally is that a proxy is needed as otherwise one can't get beyond the proxy. And b) I can't reproduce that

PR #1374, opened 2019-11-11
Fix missing negation in check_resolver_bins

a666:
Fix minor mistake introduced in 61238f1a4f71ac077b414a502f3092c9655572ab


drwetter:
Thanks for catching this!

PR #1375, opened 2019-11-13
Enhance run_protocols() in --ssl-native mode

dcooper16:
When `tls_sockets()` is used for `run_protocols()`, for each protocol version the results will indicate whether the server responded to a ClientHello for that protocol version with (1) a ServerHello for that same protocol version, (2) a ServerHello for a different (hopefully lower) protocol version, or (3) a handshake failure.

Currently, however, `run_prototest_openssl()` does not distinguish between cases in which the server responds with a ServerHello for a different (hopefully lower) protocol version and cases in which the server responds with a handshake failure. This PR changes `run_prototest_openssl()` so that it distinguishes between these two cases (as long as `$OPENSSL` supports the protocol version specified in the ServerHello).

Making use of the additional information provided by `run_prototest_openssl()`, this PR also modifies `run_protocols()` to check that version negotiation was performed correctly even if `$using_sockets` is false.

Note that one special case needed to be addressed. If an SSLv3-only server is being tested using an `$OPENSSL` that does not support SSLv3, then `$latest_supported` must not be set to SSLv3. In the case of a server like this, it is possible that support for SSLv3 will be determined by `determine_optimal_sockets_params()`, which will cause `run_protocols()` to report that the server supports SSLv3, even though `$OPENSSL` does not support SSLv3 and testing is being performed in --ssl-native mode. If `$latest_supported` were set, then later tests in `run_protocols()` would incorrectly report a version negotiation failure, even though the failure to connect was a result of a limitation of `$OPENSSL` rather than a fault of the server.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4795,10 +4795,19 @@ locally_supported() {
 #
 run_prototest_openssl() {
      local -i ret=0
+     local protos proto
 
      # check whether the protocol being tested is supported by $OPENSSL
      $OPENSSL s_client "$1" -connect x 2>&1 | grep -aq "unknown option" && return 7
-     $OPENSSL s_client $(s_client_options "-state $1 $STARTTLS $BUGS -connect $NODEIP:$PORT $PROXY $SNI") >$TMPFILE 2>&1 </dev/null
+     case "$1" in
+          -ssl2) protos="-ssl2" ;;
+          -ssl3) protos="-ssl3" ;;
+          -tls1) protos="-no_tls1_2 -no_tls1_1 -no_ssl2"; "$HAS_TLS13" && protos+=" -no_tls1_3" ;;
+          -tls1_1) protos="-no_tls1_2 -no_ssl2"; "$HAS_TLS13" && protos+=" -no_tls1_3" ;;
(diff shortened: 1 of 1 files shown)


drwetter:
Looks good, thank you!

OT: Any idea what's wrong with the HTML failures in Travis? Test 6 from ``t/08_isHTML_valid.t``  seems to fail randomly, happened also to me.

dcooper16:
> OT: Any idea what's wrong with the HTML failures in Travis? Test 6 from t/08_isHTML_valid.t seems to fail randomly, happened also to me.

Hi Dirk,

In the case of this PR, I downloaded the raw log from Travis and compared the two HTML outputs:
```
> diff expected.txt got.txt 
135c135
< #  Ticketbleed (CVE-2016-9244), experiment.  not vulnerable (OK)
---
> #  Ticketbleed (CVE-2016-9244), experiment.  not vulnerable (OK), reply empty
```
Same thing for the Travis log for PR #1329.

I'm not familiar with the Ticketbleed tests, so I haven't tried to figure out why the results for a server are not always the same.

drwetter:
Ah, ok. Thanks.

The cause seems to be in L14141 and is kind of strange to me: 1x an empty reply and the other time a non-empty reply.

PR #1377, opened 2019-11-14
Support new OpenSSL/LibreSSL versions

dcooper16:
This PR enhances support for the latest versions of OpenSSL and LibreSSL.

The development version of OpenSSL at https://github.com/openssl/openssl/ is version 3.0.0-dev. So, checks for OpenSSL versions need to support this version as well. At the same time, the latest versions of LibreSSL are 3.0.0, 3.0.1, and 3.0.2, so version number alone will no longer be sufficient to distinguish between OpenSSL and LibreSSL.

In addition to checks for these new version numbers, this PR addresses a couple of other issues:

 - In LibreSSL, the "$OPENSSL ciphers" command will not accept any protocol version other than "-tls1" as a parameter (and even including "-tls1" as an option is described as "deprecated").  So, this PR ensures that "$OPENSSL ciphers" is not passed any protocol version, other than "-tls1", as an option if LibreSSL is being used.
- In OpenSSL 3.0.0-dev, "$OPENSSL dgst" can no longer be used to compute HMACs, but a new "$OPENSSL mac" function has been created. So, this PR changes `hmac()` to use "$OPENSSL mac" with OpenSSL 3.0.0-dev.

Note that I have not tested the modified version of `sub_session_resumption()`. I am just assuming that for this function OpenSSL 3.0.0-dev works the same as OpenSSL 1.1.1 and that all versions of LibreSSL work the same as OpenSSL 1.1.0 and earlier.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1797,7 +1797,10 @@ check_revocation_ocsp() {
      else
           host_header=${uri##http://}
           host_header=${host_header%%/*}
-          if [[ $OSSL_VER_MAJOR.$OSSL_VER_MINOR == "1.1.0"* ]] || [[ $OSSL_VER_MAJOR.$OSSL_VER_MINOR == "1.1.1"* ]]; then
+          if [[ "$OSSL_NAME" =~ LibreSSL ]]; then
+               host_header="-header Host ${host_header}"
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, David!

> Note that I have not tested the modified version of ``sub_session_resumption()``.

I guess time will tell. It took a while until 1.1.1 was ~working.

PR #1378, opened 2019-11-14
Don't call s_client for unsupported protocol versions

dcooper16:
This PR fixes a couple of places where "$OPENSSL s_client" is called with "-ssl3" even if SSLv3 is not supported.

The fix in `ciphers_by_strength()` is easy, as the issue only occurs if "$using_sockets" is true. If SSLv3 (or TLSv1.3) is not supported, then testing using "$OPENSSL s_client" is skipped and all of the supported ciphers are found using `tls_sockets()`.

The fix for `run_tls_fallback_scsv()` is more complicated. While it is easy to avoid calling "$OPENSSL s_client" with "-ssl3" if SSLv3 is not supported, it is not easy to determine the correct message to present to the user if support for SSLv3 (and possibly also TLSv1.3) is unknown.

For the case in which $high_proto cannot be set, I believe that I have covered all of the possibilities, but am not sure if the correct message/rating is used in every case.

For the case in which it is not possible to determine whether SSLv3 is the $low_proto, more could be done. If $high_proto is TLS 1.1 or TLS 1, then this PR is okay, as it is possible that SSLv3 would be the fallback protocol, but there is no way to tell. However, it seems unlikely that a server would support TLS 1.2 and SSLv3, but not TLS 1.1 or TLS 1. So, perhaps if $high_proto is TLS 1.2 and the server does not support TLS 1.1 or TLS 1, it should just be assumed that SSLv3 is not supported, even if it cannot be tested.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -14840,16 +14840,12 @@ run_tls_fallback_scsv() {
 
      # Next find a second protocol that the server supports.
      for p in $protos_to_try; do
-          if [[ "$p" == ssl3 ]] && ! "$HAS_SSL3"; then
-               prln_local_problem "Can't test: $OPENSSL does not support SSLv3"
-               fileout "$jsonID" "WARN" "Can't test: $OPENSSL does not support SSLv3"
-               return 1
-          fi
           [[ $(has_server_protocol "$p") -eq 1 ]] && continue
           if [[ $(has_server_protocol "$p") -eq 0 ]]; then
                low_proto="$p"
                break
           fi
+          [[ "$p" == ssl3 ]] && ! "$HAS_SSL3" && continue
           $OPENSSL s_client $(s_client_options "-$p $STARTTLS $BUGS -connect $NODEIP:$PORT $PROXY $SNI") >$TMPFILE 2>$ERRFILE </dev/null
           if sclient_connect_successful $? $TMPFILE; then
                low_proto="$p"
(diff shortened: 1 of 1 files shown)


drwetter:
Hi @dcooper16, 

when you think it's ready, please just remove "WIP" in the title or give me another hint.

Cheers, Dirk

dcooper16:
> when you think it's ready, please just remove "WIP" in the title or give me another hint.

Okay. I won't be able to work on it until at least the week after next. In the meantime, however, if you have any suggestions, please let me know.

dcooper16:
Hi Dirk,

I removed "WIP" from the title. I believe it is ready, but lines 14795 - 14821 in particular need some extra review. I believe that the logic of each of the if/elif statements is correct and that the corresponding comment correctly describes each case. However, the text that is printed for each case and the rating that is given is a matter of opinion. It is not always clear what to report when some information about the server is known and some is unknown.

drwetter:
Thanks! Will have a thorough look.

Like;y not related but Travis seems to have a problem again with one of the badssl certificates.  Haven't looked at it, only spotted the mail.

dcooper16:
> Like;y not related but Travis seems to have a problem again with one of the badssl certificates. Haven't looked at it, only spotted the mail.

It seems that the certificate for self-signed.badssl.com expired at "2018-08-08 17:17" and the Travis test expects the certificate to not be expired. This may be a temporary issue, as other Travis tests have run and passed since August 8. So, it seems that self-signed.badssl.com is offered an old certificate at the moment, but at other times has offered a newer one.

drwetter:
Ok, finally looked at it in detail. Thanks and sorry for the delay!

PR #1382, opened 2019-11-19
Adding indent for long(er) http security headers

szycha76:
Longer HTTP security headers are not properly indented.  Please fix.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3048,7 +3048,7 @@ run_security_headers() {
                     first=false
                fi
                pr_svrty_good "$header"
-               outln " $(out_row_aligned_max_width "$HEADERVALUE" "$spaces" $TERM_WIDTH)"
+               outln " $(out_row_aligned_max_width "$HEADERVALUE" "$(echo -n $header|sed 's/./ /g') $spaces" $TERM_WIDTH)"
                fileout "$header" "OK" "$HEADERVALUE"
           fi
      done


drwetter:
hmm, works for me ;-)

You have an example where it does not for you?

szycha76:
When processing a response with longer headers, the first line is not properly wrapped:

001753-before-#1382

With my patch (which is far from ideal) a more readable/pleasant indentation is produced:

001754-after-#1382

drwetter:
Now the (n+1)th line (n>1) appears not properly wrapped to me.

Seriously: I am afraid that for smaller terminals I don't like it.  Admittedly the first line seems not ok. But I'd rather prefer not a double indentation. It becomes rather long otherwise. See github.com and e.g. replace or add ``Content-Security-Policy`` with ``Content-Security-Policy-Report-Only`` and same values.

If the line wrapping is not double indented but just once that'd be fine.  @dcooper16 : didn't you mention a reason years back why the first line formatted by ``out_row_aligned_max_width()`` is not wrapped as expected?

dcooper16:
Hi @drwetter, @szycha76,

Sorry that I wasn't following this PR earlier. I just submitted PR #1405, which I hope you will both accept as solving the problem that this PR was intended to address.

The problem at the moment is that testssl.sh is trying to print:
```
Content-Security-Policy default-src ...
```
in the area that is indented, but only "default-src ..." is passed to `out_row_aligned_max_width()`. So, `out_row_aligned_max_width()` thinks it has `$TERM_WIDTH-${#spaces}` characters available to print the first row, but it really only has `$TERM_WIDTH-${#spaces}-24` characters available. PR #1405 fixes the problem by passing `"$header $HEADERVALUE"` to `out_row_aligned_max_width()` rather than just `$HEADERVALUE`, so that `out_row_aligned_max_width()` can correctly determine how many characters it can print on the first line of output.

I just added a second commit to #1405, which increases the indentation of every line after the first one. It only increases the indentation by two spaces, so it doesn't reduce the printable area as much as this PR does, but indenting by two extra spaces does seem to make the output a bit more readable.

drwetter:
see #1405

szycha76:
Awesome.  I expected some bash hacker would actually *fix* the issue -- that's why I wrote "Please fix" not "Please merge" in the first place ;-)

Thanks - and keep up the good work.

PR #1383, opened 2019-11-19
Quick'n'dirty client certificate support for s_client

szycha76:
I was amazed to see what your script can do without client certificate.  Yet alpha-level support for client certs can be easily added by injecting relevant options via s_client_options() function.

Status: works for me ;-)

Usage:
$ export keyopts="-cert path/to/cert.pem -CAfile path/to/cert.pem"
$ ./testssl.sh [usual options]

cert.pem may be single file containing pem-encoded:
- certificate key (not encrypted)
- client certificate
- any number of intermediate certificates

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2027,7 +2027,7 @@ s_client_options() {
           [[ "$1" =~ secp192r1 ]] && options="${options//secp192r1/prime192v1}"
           [[ "$1" =~ secp256r1 ]] && options="${options//secp256r1/prime256v1}"
      fi
-     tm_out "$options"
+     tm_out "$options $keyopts"
 }
 
 ###### check code starts here ######


drwetter:
Hi @szycha76 ,

thanks! A good + straightforward idea. There seem to be a handful of cases where ``s_client_options()`` is not being used. Probably it doesn't matter much as the places for a client certificate required ~only the HTTP header test.

szycha76:
If `s_client` is used without `s_client_options()` by intention, and client certificate would come handy, one can easily put just `$keyopts` there directly.

Probably you could easily turn it into command line option, since external variables may be regarded as less elegant/readable and harder to debug. However, external variables are not recorded in bash history file nor in process signature, which may be considered beneficial in this context.

dcooper16:
> There seem to be a handful of cases where s_client_options() is not being used.

It has been a long time since I wrote the `s_client_options()` function, but I just looked at all of the places where `s_client_options()` is not used. Many of them are not actually connection attempts -- they include `-connect x`. In a couple of other places, the command line is `$options` and `$options` was created using `s_client_options()`.

For the calls that are connection attempts, one includes the `-ssl3` option and the rest include the `-ssl2` option:
```
> grep '$OPENSSL s_client' testssl.sh | grep -v s_client_options | egrep -v '^ *#' | grep -v 'connect x' | grep -v '$options' | grep -v help | grep -v foo
                    $OPENSSL s_client -cipher "${ciphers_to_test:1}" $STARTTLS $BUGS -connect $NODEIP:$PORT $PROXY -ssl2 >$TMPFILE 2>$ERRFILE </dev/null
          $OPENSSL s_client $STARTTLS $BUGS -connect $NODEIP:$PORT $PROXY -ssl2 >$TMPFILE 2>$ERRFILE </dev/null
               $OPENSSL s_client $STARTTLS $BUGS -connect $NODEIP:$PORT $PROXY -ssl2 >$TMPFILE 2>$ERRFILE </dev/null
                    $OPENSSL s_client -cipher "$1" $BUGS $STARTTLS -connect $NODEIP:$PORT $PROXY -ssl2 2>$ERRFILE >$TMPFILE </dev/null
                    $OPENSSL s_client $STARTTLS $BUGS -ssl2 -cipher $ssl2_sweet32_ciphers -connect $NODEIP:$PORT $PROXY >$TMPFILE 2>$ERRFILE </dev/null
          $OPENSSL s_client -ssl3 $STARTTLS $BUGS -cipher $cbc_ciphers -connect $NODEIP:$PORT $PROXY >$TMPFILE 2>$ERRFILE </dev/null
               $OPENSSL s_client $STARTTLS $BUGS -cipher $exportrsa_cipher_list -connect $NODEIP:$PORT $PROXY -ssl2 >$TMPFILE 2>$ERRFILE </dev/null
          $OPENSSL s_client -cipher "${sslv2_ciphers_ossl:1}" $STARTTLS $BUGS -connect $NODEIP:$PORT $PROXY -ssl2 >$TMPFILE 2>$ERRFILE </dev/null
```

While I would want to do a bit of testing before making any changes, my guess would be that there is no reason that `s_client_options()` could not be added to the above lines. I think the call to `s_client_options()` was just omitted because in these cases `s_client_options()` wouldn't actually do anything to modify the command line, so calling it would be a waste.

drwetter:
This morning (@dcooper16 ) I did a similar grep :-)

I also think it doesn't matter whether we add more instances. The certificate is only needed when we want to transfer something over the encrypted connection or complete the TLS/SSL handshake for other reasons.

See e.g. this site: https://secure.cacert.org/

drwetter:
Thanks @szycha76 ! I merged it. Needs some polish though (see #1389)

PR #1386, opened 2019-11-20
Fix #1385

### LINKED ISSUE #1385: Session Resumption line when testing servers requiring a client certificate
E.g. using: ``./testssl.sh -S https://secure.cacert.org/``

```

 TLS extensions (standard)    "server name/#0" "renegotiation info/#65281" "EC point formats/#11" "session ticket/#35" "status request/#5" "heartbeat/#15"
 Session Ticket RFC 5077 hint no -- no lifetime advertised
 SSL Session ID support       yes
 Session Resumption           Tickets no,  TLS clock skew               Random values, no fingerprinting possible 
 Signature Algorithm          SHA512 with RSA
 Server key size              RSA 4096 bits
 Server key usage             Digital Signature, Key Encipherment, Key Agreement
```

dcooper16:
This PR fixes #1385.

`sub_session_resumption()` returns 3 when `$CLIENT_AUTH` is true. However, the comment at the beginning of the function indicates that 6 will be returned. `run_server_defaults()` is prepared to handle a return value of 6 (to indicate client auth), but is not expecting 3 as a possible return value.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6098,7 +6098,7 @@ sub_session_resumption() {
                return 1
           fi
      fi
-     "$CLIENT_AUTH" && return 3
+     "$CLIENT_AUTH" && return 6
      if "$HAS_NO_SSL2"; then
           addcmd+=" -no_ssl2"
      else


drwetter:
You're right, thank you!

PR #1390, opened 2019-11-21
Fix indentation

dcooper16:
This PR fixes some indentation issues. The PR is a bit long, but it only makes changes to indentation (except for one comment line, where a trailing space character is removed).


drwetter:
Yes, good idea. Thank you!

PR #1397, opened 2019-11-25
Fix Bash error messages on bad command line

dcooper16:
If the command line is bad and either `$do_json` or `$do_pretty_json` is set, but `$JSONFILE` is not set, then this results in a Bash error messages. For example:
```
testssl.sh --json out.json 127.0.0.1

Fatal error: URI comes last

./testssl.sh: line 811: : No such file or directory
./testssl.sh: line 812: : No such file or directory
./testssl.sh: line 798: : No such file or directory
./testssl.sh: line 799: : No such file or directory
./testssl.sh: line 798: : No such file or directory
./testssl.sh: line 799: : No such file or directory
./testssl.sh: line 798: : No such file or directory
./testssl.sh: line 799: : No such file or directory
./testssl.sh: line 798: : No such file or directory
./testssl.sh: line 799: : No such file or directory
./testssl.sh: line 798: : No such file or directory
./testssl.sh: line 821: : No such file or directory
./testssl.sh: line 744: : No such file or directory
```
The problem occurs because `cleanup()` calls `fileout_footer()`, which in turn calls `fileout_json_footer()`. In the case above, `$do_json` has been set, but `$JSONFILE` is empty.

A similar problem would occur if `$do_json` has been set, but `$JSONFILE` is a directory, which could happen with a command line such as:
```
   testssl.sh --jsonfile jsondirectory --html out.html 127.0.0.1
```
If `$do_json` is set and `$JSONFILE` points to a file, then another problem occurs:
```
testssl.sh --jsonfile out.json --html out.html 127.0.0.1
```
results in just a footer being added to out.json, so that the resulting file is not valid JSON.

This PR fixes the problem by having `fileout_json_footer()` check that `$JSONFILE` is not empty and that it is not directory before attempting to write to `$JSONFILE`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -734,6 +734,9 @@ strip_quote() {
 #################### JSON FILE FORMATTING ####################
 
 fileout_json_footer() {
+     if [[ ! -s "$JSONFILE" ]] || [[ -d "$JSONFILE" ]]; then
+          return 0
+     fi
      if "$do_json"; then
           if [[ "$SCAN_TIME" -eq 0 ]]; then
                fileout_json_finding "scanTime" "WARN" "Scan interrupted" "" "" ""


dcooper16:
Note that there are some issues that this PR does not address. If a call is made of the form:
```
testssl.sh --jsonfile out.json --html out.html 127.0.0.1
```
and out.json already existed (and was non-empty), then the following will just be appended to out.json:
```
         {
              "id"           : "scanProblem",
              "ip"           : "/",
              "port"         : "443",
              "severity"     : "FATAL",
              "finding"      : "URI comes last"
          }
,         {
              "id"           : "scanTime",
              "ip"           : "/",
              "port"         : "443",
              "severity"     : "WARN",
              "finding"      : "Scan interrupted"
          }
]
```
This is because `json_header()` has not yet been called to check whether out.json is non-empty.

On the other hand, if the above call is made and out.json does not yet exist (or is empty), then out.json will not be created (i.e., the "URI comes last" message will not be sent to the JSON file). This may not be an issue, though. If there is a problem with the command line, then it may not be a good idea to write to any file name specified on the command line, since the only thing we know for certain is that the command line is incorrect.

Another related issue is that if the command line is:
```
testssl.sh --jsonfile out.json 127.0.0.1
```
and out.json already exists and is non-empty, the program will stop with an error:
```
Fatal error: non-empty "out.json" exists. Either use "--append" or (re)move it
```
and the following will be written to out.json:
```
         {
              "id"           : "scanProblem",
              "ip"           : "/",
              "port"         : "443",
              "severity"     : "FATAL",
              "finding"      : "non-empty 'out.json' exists. Either use '--append' or (re)move it"
          }
,         {
              "id"           : "scanTime",
              "ip"           : "/",
              "port"         : "443",
              "severity"     : "WARN",
              "finding"      : "Scan interrupted"
          }
]
```
Aside from this creating bad JSON, it seems wrong to write to out.json if the goal is to prevent accidentally overwriting an existing file.

So, this PR is a partial fix, but there are some other issues that need to be addressed, even though they are rare edge cases.

PR #1402, opened 2019-12-03
Dockerfile: alpine 3.10 and simplify installation

vitaliytv:


drwetter:
Thank you!

For 3.10: yes. Makes sense (despite we'll have some time: https://wiki.alpinelinux.org/wiki/Alpine_Linux:Releases). I haven't tested 3.10 yet. But did you (and are you confident it works?)

For the other points (simplify installation): you're right. @gandrade was earlier here. But I had a point here which also applies to your suggestion, see https://github.com/drwetter/testssl.sh/pull/1142#issuecomment-561111125

I'll change the run statements hereafter. If you signal that you have tested alpine 3.10 good enough., please open another PR.

Cheers, Dirk

vitaliytv:
> Thank you!
> 
> For 3.10: yes. Makes sense (despite we'll have some time: https://wiki.alpinelinux.org/wiki/Alpine_Linux:Releases). I haven't tested 3.10 yet. But did you (and are you confident it works?)
> 
> For the other points (simplify installation): you're right. @gandrade was earlier here. But I had a point here which also applies to your suggestion, see [#1142 (comment)](https://github.com/drwetter/testssl.sh/pull/1142#issuecomment-561111125)
> 
> I'll change the run statements hereafter. If you signal that you have tested alpine 3.10 good enough., please open another PR.
> 
> Cheers, Dirk

for my tests 3.10 works well

drwetter:
Travis seemed to be fine as well. So maybe you want to open another PR?

vitaliytv:
> Travis seemed to be fine as well. So maybe you want to open another PR?

done

https://github.com/drwetter/testssl.sh/pull/1404

PR #1404, opened 2019-12-03
Dockerfile: alpine 3.10

vitaliytv:


drwetter:
@vitaliytv I wanted to keep the update & upgrade statement

PR #1405, opened 2019-12-03
Fix formatting of HTTP security headers

dcooper16:
When printing out HTTP security headers, `run_security_headers()` uses `out_row_aligned_max_width()`, since some headers are very long and need to be wrapped. At the moment, however, the first line is too long. The problem is that while `"$header $HEADERVALUE"` is printed in the indented area, only `$HEADERVALUE` is passed to `out_row_aligned_max_width()`.

This PR fixes the problem by passing `"$header $HEADERVALUE"` to `out_row_aligned_max_width()` so that the the first line is wrapped at the correct place.


drwetter:
Thanks David, looks good. The additional 2 chars are a good compromise.

PR #1413, opened 2019-12-12
fixed verson number to 3.0rc6

KommX:
This just increases the version number.


drwetter:
Thanks but did that already

PR #1420, opened 2020-01-02
Dockerfile: Alpine 3.11

vitaliytv:


drwetter:
@vitaliytv : Thanks, first.

You need to sell your PR better.
*  I supposed it works and you have thoroughly tested it?
* Also please it would be great for explaining why this project should move to 3.11. 

For the latter I got a feeling now that at least it should (but can't tell as the release is rather young) not have a negative impact after google for the release notes.

vitaliytv:
>> I supposed it works and you have thoroughly tested it?
yes
>> Also please it would be great for explaining why this project should move to 3.11.

v3.10 security fixes only EOL: 2021-05-01
v3.11 bug fixes EOL: 2021-11-01

https://wiki.alpinelinux.org/wiki/Alpine_Linux:Releases

drwetter:
That sounds better, thanks

PR #1423, opened 2020-01-03
cert_validityPeriod is wrong.

fbacchella:
Calculation should be: `$((diffseconds / (3600 * 24)` instead of `$((diffseconds / 3600 * 24 ))`


drwetter:
Thanks

PR #1440, opened 2020-01-13
Fix Safari 13.0 Client Simulation

dcooper16:
The ciphersuites string for Safari 13.0 ends with a colon (':'), which causes OpenSSL to reject the command line when client simulation testing is performed in `--ssl-native` mode. This PR fixes the problem by removing the trailing colon.


drwetter:
Ok thanks, David!

Hmm... that kind of error is not being spotted by Travis

PR #1441, opened 2020-01-13
Fix run_server_preference() in --ssl-native mode

dcooper16:
This PR fixes two problems that occur when testing a server that supports TLSv1.3 using OpenSSL 1.1.1 in `--ssl-native` mode.

First, when testing whether the server has a cipher order, the value of `$sclient_success` is checked after each call to `tls_sockets()`, but `$sclient_success` is not to the result of the `tls_sockets()` call. As the goal is just to verify that the connection was successful (and didn't downgrade), `$?` can be checked rather than `$sclient_success`. [When not in `--ssl-native mode` this problem is masked since `$sclient_success` is set to 0 earlier in the function.]

The second problem is that line 6646 tries to copy "$TEMPDIR/$NODEIP.parse_tls13_serverhello.txt", but this file is currently only created (on line 6287) if `tls_sockets()` is used to determine the negotiated protocol. This PR fixes the problem by also populating "$TEMPDIR/$NODEIP.parse_tls13_serverhello.txt" when OpenSSL is used to determine the negotiated protocol.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6293,6 +6293,7 @@ run_server_preference() {
           if sclient_connect_successful $? $TMPFILE; then
                cipher0=$(get_cipher $TMPFILE)
                debugme tm_out "0 --> $cipher0\n"
+               cp $TMPFILE "$TEMPDIR/$NODEIP.parse_tls13_serverhello.txt"
           else
                # 2 second try with $OPTIMAL_PROTO especially for intolerant IIS6 servers:
                $OPENSSL s_client $(s_client_options "$STARTTLS $OPTIMAL_PROTO $BUGS -connect $NODEIP:$PORT $PROXY $SNI") </dev/null 2>>$ERRFILE >$TMPFILE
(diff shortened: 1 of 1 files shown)


dcooper16:
In developing this PR, a minor issue occurred to me. Currently, `run_server_preferences()` always uses `tls_sockets()` to determine whether the server has a cipher order for TLSv1.3. However, it seems that when in `--ssl-native`, OpenSSL should be used for the testing rather than `tls_sockets()` (at least if OpenSSL supports TLSv1.3).

drwetter:
Thanks!

> In developing this PR, a minor issue occurred to me. Currently, run_server_preferences() always uses tls_sockets() to determine whether the server has a cipher order for TLSv1.3. However, it seems that when in --ssl-native, OpenSSL should be used for the testing rather than tls_sockets() (at least if OpenSSL supports TLSv1.3).

Good point. Strictly speaking, yes: we're doing something out of our previous way performing checks. OTOH it'll concern only people don't use our openssl, either when it comes from the distribution supporting TLS 1.3 (they don't distribute testssl.sh' openssl to my knowledge). Or when used deliberately with ``--ssl-native``.

The effect seems minimal to me. I would leave it up to you.

Issue #1273, opened 2019-05-21
maketempf() assumes /tmp is writable 


danieldjewell:
Hello!

Although this is probably a minority of cases, running testssl.sh on a system that does not permit writing to /tmp (e.g. Termux on Android) causes the entire script to fail because the path /tmp is hardcoded and assumed to be writable:

https://github.com/drwetter/testssl.sh/blob/d6fb23215239cbe33847902589ac7f61a5739a95/testssl.sh#L16676-L16677

A simple fallback to trying to write to the current directory seems like an easy and workable solution? I can write it up and do a PR if you'd like.

drwetter:
Hi @danieldjewell ,

sounds good, we could work on that. Just out of curiosity: Did you get testssl.sh to work on Android? I tried a (longer) while back and I didn't get anywhere fast enough. 

Cheers, Dirk

drwetter:
Daniel: in the case I wasn't answering clear enough: yes, please go ahead if you want to submit a PR.

I am happy to do it as well otherwise.

danieldjewell:
@drwetter Sounds good! I'll work on it!

Actually, no, the script does not run fully ... there are a few other issues on Termux that I'm going to look at ... the next one is quoted below. The error is that one can't write to /dev/stdout on Termux (not rooted)... I'm not sure if there are other things that are breaking it. 

The script *does* start running on Termux -- it detects OpenSSL, it starts some of the scans, and does output *some* results. (I'm not 100% sure but I don't believe the deb package of openssl-1.1.1b-3 from termux-packages enables all of the "old" ciphers for testing -- it seems it might not enable SSLv2/3 by default -- see: https://github.com/termux/termux-packages/blob/master/packages/openssl/build.sh for the build options... but that's not too terribly difficult to fix -- I've compiled openssl before directly inside Termux with additional options.)

That said, I'm not too sure if there are additional things that could break. I'll dig in and find out. I see a few ways to possibly address these issues:

* Write some kind of platform detection into the system and then modify the various things that would break on that platform (e.g. on Termux, $PREFIX and $HOME are set to /data/data/com.termux/files/usr and /data/data/com.termux/files/home, respectively) ... This would be good if there are mutliple platforms that need very custom settings to work (are there others?) ... obviously, this is also somewhat complex to maintain ... 
[...]

PR #1445, opened 2020-01-14
Try temp file creation in a different location

drwetter:
... if the standard directory /tmp is not allowed to write to. As noted in #1273 this might be the case for Termux on Android.

Looking at #1273 this PT amends #1443 and replaces #1277

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -17194,7 +17194,12 @@ EOF
 }
 
 maketempf() {
-     TEMPDIR=$(mktemp -d /tmp/testssl.XXXXXX) || exit $ERR_FCREATE
+     TEMPDIR=$(mktemp -d /tmp/testssl.XXXXXX)
+     if [[ $? -ne 0 ]]; then
+          # for e.g. devices where we can't write to /tmp:
+          TEMPPATH=$PWD
+          TEMPDIR=$(mktemp -d $PWD/testssl.XXXXXX) || exit $ERR_FCREATE
+     fi
      TMPFILE=$TEMPDIR/tempfile.txt || exit $ERR_FCREATE
      if [[ "$DEBUG" -eq 0 ]]; then
           ERRFILE="/dev/null"
--- a/testssl.sh
+++ b/testssl.sh
@@ -17198,10 +17198,10 @@ maketempf() {
      if [[ $? -ne 0 ]]; then
           # For e.g. devices where we can't write to /tmp we chose $PWD but we can't
           # allow every char as we haven't quoted all strings depending on it, see #1445
-          if [[ $PWD =~ ^[A-Za-z0-9\.,-/_]+$ ]]; then
-               fatal "\$PWD contains illegal chars: \"$PWD\"" $ERR_FCREATE
+          if [[ $PWD =~ [^A-Za-z0-9\.,/_-] ]]; then
+               fatal "\$PWD contains illegal chars: \"$BASH_REMATCH\"" $ERR_FCREATE
           fi
-          TEMPDIR=$(mktemp -d "PWD/testssl.XXXXXX") || exit $ERR_FCREATE
+          TEMPDIR=$(mktemp -d "$PWD/testssl.XXXXXX") || exit $ERR_FCREATE
      fi
      TMPFILE=$TEMPDIR/tempfile.txt || exit $ERR_FCREATE
      if [[ "$DEBUG" -eq 0 ]]; then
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -17194,7 +17194,12 @@ EOF
 }
 
 maketempf() {
-     TEMPDIR=$(mktemp -d /tmp/testssl.XXXXXX) || exit $ERR_FCREATE
+     TEMPDIR=$(mktemp -d /tmp/testssl.XXXXXX)
+     if [[ $? -ne 0 ]]; then
+          # for e.g. devices where we can't write to /tmp:
+          TEMPPATH=$PWD
+          TEMPDIR=$(mktemp -d $PWD/testssl.XXXXXX) || exit $ERR_FCREATE
```
dcooper16: Hi Dirk,

I think the `TEMPPATH=$PWD` line can just be removed. As far as I can tell, `TEMPPATH` is not currently defined in testssl.sh, and this PR doesn't make use of it, so there doesn't seem to be a need to define it.

As someone noted in #1277, there is a possibility that the current directory ($PWD) contains spaces in it, which could cause problems since at the moment $TEMPDIR is used in many places without quoting it. We could go through and add quotes wherever `$TEMPDIR` (or any variable such as TMPFILE, ERRFILE, HEADERFILE, SOCK_REPLY_FILE, HOSTCERT, etc.) is used. Barring that, however, it might be a good idea to check whether `$PWD` contains any white space characters, and then issue an error if /tmp is not writable and `$PWD` contains white space.

drwetter: Still thinking on it not sure whether this is enough

``mkdir "ls;" ; cd "ls;" ``. What else should we not allow? Maybe an alphanumerical white list is just better plus "/"

testssl.sh
```diff
@@ -17194,7 +17194,15 @@ EOF
 }
 
 maketempf() {
-     TEMPDIR=$(mktemp -d /tmp/testssl.XXXXXX) || exit $ERR_FCREATE
+     TEMPDIR=$(mktemp -d /tmp/testssl.XXXXXX)
+     if [[ $? -ne 0 ]]; then
+          # For e.g. devices where we can't write to /tmp we chose $PWD but we can't
+          # allow every char as we haven't quoted all strings depending on it, see #1445
+          if [[ $PWD =~ ^[A-Za-z0-9\.,-/_]+$ ]]; then
```
dcooper16: Hi Dirk,

`^[A-Za-z0-9\.,-/_]+$` didn't work for me:
```
> [[ "/home/user" =~ ^[A-Za-z0-9\.,-/_]+$ ]] && echo "illegal charaters: \"$BASH_REMATCH\"" || echo "OK"
illegal charaters: "/home/user"
> [[ "/home/the user" =~ ^[A-Za-z0-9\.,-/_]+$ ]] && echo "illegal charaters: \"$BASH_REMATCH\"" || echo "OK"
OK
> [[ "/home/user/is a semicolon okay;" =~ ^[A-Za-z0-9\.,-/_]+$ ]] && echo "illegal charaters: \"$BASH_REMATCH\"" || echo "OK"
OK
```
`[[ "$PWD" =~ [^A-Za-z0-9\.,/_-] ]]` seems to work better:
```
> [[ "/home/user" =~ [^A-Za-z0-9\.,/_-] ]] && echo "illegal charaters: \"$BASH_REMATCH\"" || echo "OK"
OK
> [[ "/home/the user" =~ [^A-Za-z0-9\.,/_-] ]] && echo "illegal charaters: \"$BASH_REMATCH\"" || echo "OK"
illegal charaters: " "
> [[ "ls;" =~ [^A-Za-z0-9\.,/_-] ]] && echo "illegal charaters: \"$BASH_REMATCH\"" || echo "OK"
illegal charaters: ";"
> [[ "/home/user/this-is_my.path" =~ [^A-Za-z0-9\.,/_-] ]] && echo "illegal charaters: \"$BASH_REMATCH\"" || echo "OK"
OK
```

testssl.sh
```diff
@@ -17194,7 +17194,15 @@ EOF
...
+               fatal "\$PWD contains illegal chars: \"$PWD\"" $ERR_FCREATE
+          fi
+          TEMPDIR=$(mktemp -d "PWD/testssl.XXXXXX") || exit $ERR_FCREATE
```
dcooper16: Missing `$`: should be `mktemp -d "$PWD/testssl.XXXXXX"`

PR #1446, opened 2020-01-14
Add missing variable declarations

dcooper16:
`derive-handshake-traffic-keys()` uses the variables `derived_secret`, `server_write_key`, and `server_write_iv`, but they are not declared as local variables of the function. This PR fixes that.


drwetter:
Thanks

PR #1451, opened 2020-01-15
Add support for HTTP Basic Auth

mkauschi:
Hi Dirk,
I had the problem, when scanning a webserver with HTTP basic auth the test for the security headers is not working correctly. I was looking in the help section of testssl for a switch to provide the basic auth credentials but could not find this feature. So I decided to implement it and open a PR on this issue.

Please let me know if anything is missing.

Cheers

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -204,6 +204,7 @@ SNEAKY=${SNEAKY:-false}                 # is the referer and useragent we leave 
 QUIET=${QUIET:-false}                   # don't output the banner. By doing this you acknowledge usage term appearing in the banner
 SSL_NATIVE=${SSL_NATIVE:-false}         # we do per default bash sockets where possible "true": switch back to "openssl native"
 ASSUME_HTTP=${ASSUME_HTTP:-false}       # in seldom cases (WAF, old servers, grumpy SSL) service detection fails. "True" enforces HTTP checks
+BASICAUTH=${BASICAUTH:-""}              # HTTP basic auth credentials can be set here like user:pass
 BUGS=${BUGS:-""}                        # -bugs option from openssl, needed for some BIG IP F5
 WARNINGS=${WARNINGS:-""}                # can be either off or batch
 DEBUG=${DEBUG:-0}                       # 1: normal output the files in /tmp/ are kept for further debugging purposes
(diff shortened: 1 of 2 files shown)


testssl.sh
```diff
@@ -19783,6 +19788,9 @@ parse_cmd_line() {
                --ssl_native|--ssl-native)
                     SSL_NATIVE=true
                     ;;
+               --basicauth|--basicauth=*)
+                   BASICAUTH="$(parse_opt_equal_sign "$1" "$2")"
+                   ;;
```
drwetter: You would need to add ``[[ $? -eq 0 ]] && shift`` here, otherwise the value of  <user:pass> will be treated as another argument to testssl.sh

testssl.sh
```diff
@@ -18412,10 +18413,14 @@ determine_service() {
           # no STARTTLS.
           determine_optimal_sockets_params
           determine_optimal_proto
+          BASIC_AUTH_HEADER=""
           $SNEAKY && \
                ua="$UA_SNEAKY" || \
                ua="$UA_STD"
-          GET_REQ11="GET $URL_PATH HTTP/1.1\r\nHost: $NODE\r\nUser-Agent: $ua\r\nAccept-Encoding: identity\r\nAccept: text/*\r\nConnection: Close\r\n\r\n"
+          if [[ ! -z "$BASICAUTH" ]]; then
+               BASIC_AUTH_HEADER="Authorization: Basic `echo $BASICAUTH | basenc --base64` \r\n"
```
drwetter: * basenc is not very portable. Could you please exchange that to openssl base64 ?
* instead of backtick please use ``$(somecmd)``

testssl.sh
```diff
@@ -18412,10 +18413,14 @@ determine_service() {
           # no STARTTLS.
           determine_optimal_sockets_params
           determine_optimal_proto
+          BASIC_AUTH_HEADER=""
```
drwetter: I would rather see that in the top please like all other vars, see _Following variables make use of $ENV and can be used like_ , maybe below ``ASSUME_HTTP``

dcooper16: > BASIC_AUTH_HEADER needs to be in the very beginning please (_Initialization part, further global vars just_)

Actually, it seems that BASIC_AUTH_HEADER is only used within `determine_service()`. So, it should be defined as a local variable in that function. By [convention](https://github.com/drwetter/testssl.sh/wiki/Coding-Style), local variables should be lower case. So, the above line can be deleted, and a line should be added  at the beginning of `determine_service()`: `local basic_auth_header=""`

drwetter: ACK

testssl.sh
mkauschi: > Moreover a line in `help()` is missing
The line in `help()` is included in my PR (line 17167). I currently don't get what is missing there. Could you please be more specific on this?

drwetter:
Litmus question: did you test it and did it work?

mkauschi:
> Litmus question: did you test it and did it work?

Yes I have tested it and it works.

drwetter:
Thanks a lot for the PR and the polishing!

PR #1453, opened 2020-01-16
Check for the Cache-Control and Pragma header

mkauschi:
Hi Dirk,
when testing webservers, I would like to see if the Cache-Control and Pragma headers are set. In some cases this is a good hint (see the following post for reference https://isc.sans.edu/forums/diary/The+Security+Impact+of+HTTP+Caching+Headers/17033/)

I have added the two headers in the other_header variable. I'm not sure if it would better fit to security_header? If you like this PR just tell me if I should put it to a different location.

Cheers

Manuel


drwetter:
> I'm not sure if it would better fit to security_header?

 Good question, normally it should be ok to put it where you suggested. It depends what's in there = the value. Value checking isn't implemented anywhere yet. That it also not easy

drwetter:
Thanks!

PR #1454, opened 2020-01-16
More polishing of http basic auth

dcooper16:
* Replace `! -z` with `-n`
* Replace "openssl' with "$OPENSSL"
* Redirect stderr output of $OPENSSL to /dev/null to supress "WARNING: can't open config file: /usr/local/etc/ssl/openssl.cnf" message (see #833)
* Remove unnecessary spaces from `$GET_REQ11` string.


drwetter:
yes, thank you, David

Issue #1444, opened 2020-01-14
Unnecessary warnings for TLS 1.3-only servers


dcooper16:
I was just testing a TLS 1.3-only server and noticed the following output lines:
```
 Secure Renegotiation (RFC 5746)           OpenSSL handshake didn't succeed
 Secure Client-Initiated Renegotiation     not vulnerable (OK)
 CRIME, TLS (CVE-2012-4929)                test failed (couldn't connect)
```
For all three of these tests, the tests are specifically designed to not connect using TLS 1.3. Given that `$TLS13_ONLY` is `true` at the time these tests are run, it seems that the output should indicate that everything is okay rather than using `prln_warning()` to suggest that something went wrong with the test.

The output for "Secure Client-Initiated Renegotiation" seems fine. However, it does seem wasteful to perform a test that involves specifying `-no_tls1_3` on the command line if `$TLS13_ONLY` is true.

drwetter:
For TLS 1.3 only hosts there should be per definition no TLS compression offered. I haven't implemented this. Maybe later (call it a day soon)

drwetter:
see PR above

PR #1463, opened 2020-01-20
Shortcuts for vulnerability tests for TLS 1.3 only servers

### LINKED ISSUE #1462: Unnecessary warnings for TLS 1.3-only servers (II)
#1459  was accidentally reverted. Cleanup

drwetter:
Several vulnerability checks add a time penalty when the server side only support TLS 1.3 as The TLS 1.3 RFC 8446 and implementations known so far don't support the flaws being checked for.

This PR adds "shortcut" checks for all TLS 1.3, assuming that the TLS 1.3 implementation is correct which seems at this time a valid assumpution. That either saves a TCP connect or at least some logic to be executed.  Also in some cases a TLS 1.3 only server emitted unnecessary warnings, see #1444.

If $DEBUG -eq 1 then it outputs information that a shortcut was used. It doesn't do that in other cases because the screen output seems too obtrusive.

It also adds a shortcut for beast when SSL 3 or TLS 1.0 is is known not to be supported.

This commit readds 747fb039edc329e759aa634fed8a256031808324 which was accidentally reverted in 45f28d816652be59f6d41e7b2200f3325bf04f3e.It fixes #1462.

See also #1459.

### DIFF
--- a/t/08_isHTML_valid.t
+++ b/t/08_isHTML_valid.t
@@ -9,12 +9,12 @@ use Data::Dumper;
 
 my $tests = 0;
 my $prg="./testssl.sh";
-my $uri="dev.testssl.sh";
+my $uri="heise.de";
 my $out="";
 my $html="";
 my $debughtml="";
 my $edited_html="";
-my $check2run="--ip=one  --color 0 --htmlfile tmp.html";
+my $check2run="--ip=one --color 0 --htmlfile tmp.html";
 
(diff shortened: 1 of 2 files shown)


drwetter:
Repeating the code ``if "$TLS13_ONLY"; then ... `` seems not very beautiful. That should be addressed in 3.1dev including something similar for ciphers.

drwetter:
The remaining Travis failures are coming from badssl.com itself

PR #1464, opened 2020-01-22
Further handshakes / minor changes

drwetter:
This PR adds the Android 10 handshake simulation to the list.

In addition it modifies the handshake's service from stock Android (i.e. this project, not SSLlabs) to ANY instead of HTTP,FTP,SMTP,POP,IMAP. This adds to other STARTTLS protocols NNTPS and the like also the Android handshakes


drwetter:
see badssl problems @ https://github.com/chromium/badssl.com/issues/429

PR #1465, opened 2020-01-22
Undo copy and paste error

dcooper16:
This PR removes what appears to be a copy and paste error introduced by #1463.


drwetter:
ops, thanks!

PR #1468, opened 2020-01-22
Fix XMPP starttls

nosnilmot:
'-xmpphost' option required in combination with '-starttls xmpp' was missing
in determine_optimal_proto()

Also tweaked a couple of log messages

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -18349,7 +18349,7 @@ determine_optimal_proto() {
                     -ssl2)   "$HAS_SSL2" || continue ;;
                     *) ;;
                esac
-               $OPENSSL s_client $(s_client_options "$STARTTLS_OPTIMAL_PROTO $BUGS -connect "$NODEIP:$PORT" $PROXY -msg $STARTTLS" $SNI) </dev/null >$TMPFILE 2>>$ERRFILE
+               $OPENSSL s_client $(s_client_options "$STARTTLS_OPTIMAL_PROTO $BUGS -connect "$NODEIP:$PORT" $PROXY -msg $STARTTLS $SNI") </dev/null >$TMPFILE 2>>$ERRFILE
                if sclient_auth $? $TMPFILE; then
                     all_failed=false
                     add_tls_offered "${proto/-/}" yes
--- a/testssl.sh
+++ b/testssl.sh
@@ -10218,13 +10218,13 @@ starttls_imap_dialog() {
 }
 
 starttls_xmpp_dialog() {
-     debugme echo "=== starting imap XMPP dialog ==="
+     debugme echo "=== starting xmpp STARTTLS dialog ==="
      [[ -z $XMPP_HOST ]] && XMPP_HOST="$NODE"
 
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -18349,7 +18349,7 @@ determine_optimal_proto() {
                     -ssl2)   "$HAS_SSL2" || continue ;;
                     *) ;;
                esac
-               $OPENSSL s_client $(s_client_options "$STARTTLS_OPTIMAL_PROTO $BUGS -connect "$NODEIP:$PORT" $PROXY -msg -starttls $1" $SNI) </dev/null >$TMPFILE 2>>$ERRFILE
+               $OPENSSL s_client $(s_client_options "$STARTTLS_OPTIMAL_PROTO $BUGS -connect "$NODEIP:$PORT" $PROXY -msg $STARTTLS" $SNI) </dev/null >$TMPFILE 2>>$ERRFILE
```
dcooper16: Hi Dirk,

I just noticed an error on line 18352 that is unrelated to this PR (and probably my fault).

In the call to `s_client_options()` the end quotation mark for the parameters comes before `$SNI` rather than after it. So, if `$SNI` is not empty, its contents are treated as the second and third parameters to `s_client_options()` rather than part of the first. This results in `$SNI` not appearing in the command line sent to $OPENSSL (and may result in `-noservername` being added to the command line).

While you are modifying this line in this PR, could you fix the error by moving the closing quotation mark to after `$SNI` rather than before it?

Thanks,

David

drwetter: Thanks for the heads up, David!

Also not related but it just popped up again while looking at it: any reason here and on other occasions for ``-msg``

Thx, Dirk

drwetter:
Hi @nosnilmot ,

thanks! I PR'd to your branch and will merge your PR when you merge mine.

Cheers, Dirk

testssl.sh

`-msg` is used in 7 places at the moment: 3 in `get_server_certificate()`, 1 in `run_renego()`, 1 in `run_logjam()`, and 2 in `determine_optimal_proto()`. It seems to be needed in 6 of these 7 places, but probably can be removed from one.
* `-msg` seems to be needed here (in `determine_optimal_proto()`), as the output from `$OPENSSL s_client` is sent to `sclient_auth()`, and `sclient_auth()` has a comment line that says:
```
    if grep -q '^<<< .*CertificateRequest' "$2"; then                # CertificateRequest message in -msg
```
* `-msg` is needed in the 3 calls from `get_server_certificate()` since the output from `$OPENSSL s_client` is sent to `extract_stapled_ocsp()`, which extracts the OCSP information from the text printed as a result of `-msg`.
* In `run_logjam()` the text printed as a result of `-msg` is parsed to obtain the server's ephemeral public key.

I can't see any reason for the `-msg` in the call from `run_renego()`. I believe that you recently removed `-msg` from another call to `$OPENSSL s_client` in `run_renego()`, and it seems that it can be removed from the other one as welll.

drwetter: Thx, David

PR #1469, opened 2020-01-22
Simplify code to determine draft TLS 1.3 version

dcooper16:
This PR simplifies the code for determining which draft version of TLS 1.3 a server is offering by making use of a simple regular expression and `$BASH_REMATCH` rather than looping through every possible draft version.


drwetter:
I like that, thanks!

PR #1473, opened 2020-01-24
Full AEAD cipher implementations

dcooper16:
[RFC 8446](https://www.rfc-editor.org/rfc/rfc8446.html) specifies cipher suites that use three symmetric encryption algorithms, all of which are Authenticated Encryption with Associated Data (AEAD) algorithms. In each of these algorithms when data is encrypted an authentication tag is created, which allows the recipient to verify that the data has not been modified. The authentication may also cover some additional data that was not encrypted.

The current implementations of these algorithms in testssl.sh decrypt the ciphertext, but do not check that the authentication tag is correct (which involves the recipient computing the correct tag for the received data and then comparing it to the provided tag). While testssl.sh can get away with not checking authentication tags when receiving data, the ability to compute authentication tags is needed in order to send encrypted data, as TLS servers would reject any encrypted data that did not have a correct authentication tag. Being able to send encrypted data is necessary to be able to complete the TLS 1.3 handshake.

This PR replaces the current implementations of the symmetric encryption algorithms with full implementations of each of the algorithms. These full implementations include the ability to encrypt data for sending, and can also verify the authentication tag when decrypting data. Since the Bash implementations of these algorithms is very slow, the decryption code is designed to only compute and check authentication tags in debug mode.

While the implementation of the code to compute authentication tags for AES-CCM was based on NIST [Special Publication 800-38C](https://doi.org/10.6028/NIST.SP.800-38C), I was not able to implement the code for AES-GCM or Poly1305 from their specifications (NIST [Special Publication 800-38D](https://doi.org/10.6028/NIST.SP.800-38D) and [RFC 8439](https://www.rfc-editor.org/rfc/rfc8439.html), respectively). So, I would very much like to thank the implementers of https://github.com/mko-x/SharedAES-GCM and https://github.com/floodyberry/poly1305-donna. The implementations of AES-GCM and Poly1305 in this PR were developed by translating the C code in https://github.com/mko-x/SharedAES-GCM and https://github.com/floodyberry/poly1305-donna into Bash. I don't understand what that code is doing, but it seems to work. :-)

I have only tested this code on a computer with a 64-bit operating system. While I have not tested it, I believe that the decryption code will work with 32-bit integers if not in debug mode (i.e., if not trying to compute the authentication tags). I also believe that the AES-CCM code for computing authentication tags will work with 32-bit integers. However, AES-GCM and Poly1305 code for computing authentication tags will definitely only work on systems that have 64-bit integers. So, on systems that do not have 64-bit integers, encryption will not work for AES-GCM or ChaCha20-Poly1305, and decryption will not work for these algorithms if in debug mode.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -11097,63 +11097,7 @@ derive-handshake-traffic-keys() {
      return 0
 }
 
-generate-ccm-gcm-keystream() {
-     local icb="$1" icb_msb icb_lsb1
-     local -i i icb_lsb n="$2"
-
-     icb_msb="${icb:0:24}"
-     icb_lsb=0x${icb:24:8}
-
-     for (( i=0; i < n; i=i+1 )); do
-          icb_lsb1="$(printf "%08X" $icb_lsb)"
-          printf "\x${icb_msb:0:2}\x${icb_msb:2:2}\x${icb_msb:4:2}\x${icb_msb:6:2}\x${icb_msb:8:2}\x${icb_msb:10:2}\x${icb_msb:12:2}\x${icb_msb:14:2}\x${icb_msb:16:2}\x${icb_msb:18:2}\x${icb_msb:20:2}\x${icb_msb:22:2}\x${icb_lsb1:0:2}\x${icb_lsb1:2:2}\x${icb_lsb1:4:2}\x${icb_lsb1:6:2}"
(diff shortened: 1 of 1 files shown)


drwetter:
Excellent work, by the way!

PR #1475, opened 2020-01-27
Reorganize TLS 1.3 key derivation

dcooper16:
This PR reorganizes the code for deriving TLS 1.3 symmetric keys in order to facilitate implementing the full key schedule. For example, rather than having a single function to derive the handshake traffic keys, this PR creates one function to derive the handshake secret and a separate function to derive the handshake traffic keys. The second function has been generalized so that it can derive either client or server traffic keys. Separating into two functions also makes the handshake secret available for later use to derive the master secret and then the application traffic secrets and the application traffic keys.

This PR also changes where there message transcript is created, a message transcript will also be needed to derive the application traffic secrets. This PR includes the code to add the messages to the initial message transcript that will be needed for the input to the application traffic secret derivation function.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -10951,18 +10951,87 @@ derive-secret() {
 }
 
 # arg1: hash function
-# arg2: private key file
-# arg3: file containing server's ephemeral public key
-# arg4: ASCII-HEX of messages (ClientHello...ServerHello)
-# See key derivation schedule diagram in Section 7.1 of RFC 8446
-derive-handshake-traffic-secret() {
+# arg2: secret
+# arg3: purpose ("key" or "iv")
+# arg4: length of the key
+# See RFC 8446, Section 7.3
+derive-traffic-key() {
      local hash_fn="$1"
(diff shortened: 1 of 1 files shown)


drwetter:
Good, thanks!

testssl.sh
```diff
@@ -10951,19 +10951,88 @@ derive-secret() {
...
+#arg2: First ClientHello, if response was a HelloRetryRequest
+#arg3: HelloRetryRequest, if one was sent
+#arg4: Final (or only) ClientHello
+#arg5: ServerHello
+create-initial-transcript() {
+     local cipher="$1"
+     local clienthello1="$2" hrr="$3" clienthello2="$4" serverhello="$5"
+     local hash_clienthello1 msg_transcript
+
+     if [[ -n "$hrr" ]] && [[ "${serverhello:8:4}" == "7F12" ]]; then
```
drwetter, line 10980: Please remember that at the RHS of '==' are most of the time not needed.

PR #1476, opened 2020-01-28
TLS 1.3 Finished messages

dcooper16:
This PR adds processing of the Finished messages in TLS 1.3 handshakes. It also addresses some shellcheck issues.

If in debug mode, the HMAC of the transcript hash of the handshake context (`$msg_transcript`) is computed and compared against the Finished message sent by the server.

If the full server response is parsed and the connection with the server is not to be closed when `tls_sockets()` completes, then the TLS 1.3 handshake is completed by creating the client Finished message and sending it to the server.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -10860,6 +10860,31 @@ hmac() {
 }
 
 # arg1: hash function
+# arg2: key
+# arg3: transcript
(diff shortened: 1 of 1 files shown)


drwetter:
looks good, thank you!

PR #1477, opened 2020-01-29
Speedup AES-GCM

dcooper16:
The implementation of AES-GCM in #1473 is much slower than the original version, even when the authentication tag is not being computed. This PR modifies the code in `gcm()` in order to significantly speed up the encryption/decryption time (when authentication tags are not being computed).


drwetter:
:-) 

Thanks!

PR #1481, opened 2020-01-30
Fix HTML generation

dcooper16:
This PR fixes two issues related to the generation of HTML files.

First, text that is to appear in the HTML file is first passed through `html_reserved()` to replace reserved characters with their corresponding entity names (e.g., '>' becomes '\&gt;'). `html_reserved()` seems to work correctly on Ubuntu Linux, but it does not work as expected on MacOS. On MacOS, rather than converting '>' to '\&gt;', it gets converted to '\\\&gt;', and the backslash is rendered by browsers.

This PR appears to fix the problem -- at least it works on both Ubuntu and MacOS. However, given that the original version of `html_reserved()` was not portable, this revised version should be tested on multiple platforms.

I also noticed that in almost every case in which a string is passed to `html_out()`, it is first run through `html_reserved()`, but for some reason that is not the case in `out()` and `outln()`. I can't see any reason why `html_reserved()` is not called first in these two cases, so this PR adds in the calls.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -495,7 +495,7 @@ html_reserved(){
      output="${output//>/&gt;}"
      output="${output//\"/&quot;}"
      output="${output//\'/&apos;}"
-     tm_out "$output"
+     printf -- "%s" "$output"
      return 0
 }
 
--- a/testssl.sh
+++ b/testssl.sh
@@ -490,11 +490,11 @@ html_reserved(){
      local output
      "$do_html" || return 0
      #sed  -e 's/\&/\&amp;/g' -e 's/</\&lt;/g' -e 's/>/\&gt;/g' -e 's/"/\&quot;/g' -e "s/'/\&apos;/g" <<< "$1"
-     output="${1//\&/\&amp;}"
-     output="${output//</\&lt;}"
-     output="${output//>/\&gt;}"
-     output="${output//\"/\&quot;}"
-     output="${output//\'/\&apos;}"
+     output="${1//&/&amp;}"
+     output="${output//</&lt;}"
+     output="${output//>/&gt;}"
+     output="${output//\"/&quot;}"
+     output="${output//\'/&apos;}"
      tm_out "$output"
      return 0
 }
@@ -510,8 +510,8 @@ safe_echo()  { printf -- "%b" "${1//%/%%}"; }
 tm_out()     { printf -- "%b" "${1//%/%%}"; }
 tmln_out()   { printf -- "%b" "${1//%/%%}\n"; }
 
-out()   { printf -- "%b" "${1//%/%%}"; html_out "$1"; }
-outln() { printf -- "%b" "${1//%/%%}\n"; html_out "$1\n"; }
+out()   { printf -- "%b" "${1//%/%%}"; html_out "$(html_reserved "$1")"; }
+outln() { printf -- "%b" "${1//%/%%}\n"; html_out "$(html_reserved "$1")\n"; }
 
 #TODO: Still no shell injection safe but if just run it from the cmd line: that's fine
 


dcooper16:
Travis found a bug in this PR (and in the current code). If `html_reserved()` is called with a string that ends with '\\n', then this newline will not make it into the HTML. The result is that the terminal output will include a few blank lines that do not appear in the HTML output. The source of the problem is that `html_reserved()` uses `tm_out()` to return the modified string and `tm_out()` converts '\\n' to a newline character.

This problem has not occurred so far since `out()` and `outln()` did not call `html_reserved()`, and (apparently) no other print function (e.g., `pr_warning()`) is called with a string that ends with '\\n'.

I can see two ways to fix this problem:
* In every place where `out()` or `outln()` is called with a string that ends with '\\n', remove the '\\n' and replace it with a call to `outln()`.
* Find some way to fix `html_reserved()` so that this problem doesn't happen.

Any thoughts?

dcooper16:
I may have gotten a bit closer to addressing the '\\n' issue. On Ubuntu I seem to be able to get it to work by adding the following to `html_reserved()`:
```
output="${output//\n/\\\n}"
```
On MacOS, however, this line needs to be:
```
output="${output//\\n/\\\n}"
```
Unfortunately, so far I have been unable to find something that works on both operating systems. :-(

drwetter:
Thanks, David. Will probably not get to it today.

Travis seemed anyway to had problem yesterday evening. Some I would guess they were hiccups as e.g. my PRs later went fine.

dcooper16:
> Travis seemed anyway to had problem yesterday evening. Some I would guess they were hiccups as e.g. my PRs later went fine.

The Travis error in this PR is definitely related to the change this code makes. I had looked at #1483, and that was an unrelated error:
```
< #  Ticketbleed (CVE-2016-9244), experiment.  not vulnerable (OK)
---
> #  Ticketbleed (CVE-2016-9244), experiment.  not vulnerable (OK), reply empty
```
I just looked at #1482, and it seems that that one was catching another error on my part. When I renamed a couple of the variables in run_client_simulation(), it didn't occur to me that those variable names also appear in etc/client-simulation.txt. I just submitted #1488, which will hopefully fix that error.

drwetter:
> The Travis error in this PR is definitely related to the change this code makes. I had looked at #1483, and that was an unrelated error:

I did a fix way back and IIRC I was not flushing the server side buffers correctly before. OTOH it could also be a server side issue. I vaguely remember that they're using a loadbalancer (not F5)

dcooper16:
Hi Dirk,

I added a second commit this PR that seems to fix the Travis problem. Basically it changes `html_reserved()` to return its results using

```          printf -- "%s" "$output"```

rather than

```          printf -- "%b" "${output//%/%%}"```

Using `%s` rather than `%b` prevents the escape characters from being interpreted before `html_out()` is called. I tried it on both Ubuntu and MacOS and it seems to work on both. However, if there is something that I am missing, please let me know.

**Comment by dcooper16** [comment]
In trying to fix the problem of printing strings that end with `\n` to HTML I discovered a somewhat unrelated problem. If a string is printed that contains `%` the `%` character is printed twice. This can happen if a URL contains `%` characters (e.g., CRL, OCSP URI, HTTP redirect URI). So,
```
http://invalid.example.com/bad%20example
```
would be printed as
```
http://invalid.example.com/bad%%20example
```
If the URI was for a CRL or OCSP rather than an HTTP redirect, so that the URI was printed using `out()` rather than `pr_uri()`, then the URI would appear as
```
http://invalid.example.com/bad%%%%20example
```
in the HTML.

With the change suggested by the second commit in this PR, the HTML output would only contain two `%` characters (same as the printed output).

It is clear that the `%` is getting doubled as a result of `${1//%/%%}` in functions such as `tm_out()`, `tmln_out()`, `out()`, `outln()`, etc., but the reason for the substitution is not clear. I found an old commit (https://github.com/drwetter/testssl.sh/commit/8a2fe5915a8d4f70ba423f98eff02a6646935aa9) that changed
```
          /usr/bin/printf -- "${1//%/%%}"
```
to
```
          printf -- "%b" "${1//%/%%}"
```
Is there any chance that `${1//%/%%}` was needed with the old version but not the new one?

drwetter:
The %% was there IIRC to prevent printf from e.g. interpreting a single "%s" in the external input as a printf builtin to interpret the input as a string. 

I would need to dig a bit deeper whether that still makes sense.

Strange enough with zsh the double percent output doesn't work with zsh's internal printf.

dcooper16:
> The %% was there IIRC to prevent printf from e.g. interpreting a single "%s" in the external input as a printf builtin to interpret the input as a string.

That's what I figured. Certainly if using `printf -- "<string>"`, any `%` in `<string>` needs to be replaced with `%%`, since `printf` will interpret `%` in `<string>` as the beginning of a formatting command.

However, with `printf -- "%b" "<string>"`, it is not clear that the substitution is needed. It seems that with this form, '\\' escapes in `<string>` are interpreted, but `%` characters are not. With `printf -- "%s" "<string>"`, it would seem that neither '\\' escapes nor `%` characters in `<string>` would be interpreted. That's why I'm wondering whether the `${1//%/%%}` should have been removed when the change in https://github.com/drwetter/testssl.sh/commit/8a2fe5915a8d4f70ba423f98eff02a6646935aa9 was made.

drwetter:
I would like this also to be in 3.0. I can do that as well do that but as it is your patch...

dcooper16:
This commit applies the same changes as #1481, but to the 3.0 branch.

dcooper16:
> I would like this also to be in 3.0. I can do that as well do that but as it is your patch...

Done.

PR #1482, opened 2020-01-30
Suppress shellcheck issue SC2178

dcooper16:
Due to a [bug](https://github.com/koalaman/shellcheck/wiki/SC2178), the shellcheck program will complain if a variable is defined as an array but is later used as an ordinary string, even if the two uses are locally defined variables in different contexts. The error message is:
```
SC2178: Variable was used as an array but is now assigned a string.
```
While the warnings are not highlighting any actual problems in testssl.sh, this PR gets rid of the warnings by renaming a few variables.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4548,10 +4548,10 @@ run_client_simulation() {
      local names=()
      local short=()
      local protos=()
-     local ciphers=()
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks!

PR #1483, opened 2020-01-30
Minor code cleanup

dcooper16:
This PR cleans up the code in a few places by using "case" statements in a few places where the value of a variable is checked multiple times.


drwetter:
thank you!

PR #1484, opened 2020-01-30
TLS 1.3 post-handshake messages

dcooper16:
This PR adds support for post-handshake messages when using sockets with TLS 1.3 connections. If a TLS 1.3 connection is established and the connection is to remain open after `tls_sockets()` finishes, then after the client's Finished message is sent the master secret and the application traffic keys are computed. This PR also adds two new functions to send and receive application data over a TLS 1.3 connection.

This PR includes two proofs-of-concept for the use of the new functions: `send_app_data()` and `receive_app_data()`. `receive_app_data()` is called immediately after the client's Finished message is sent. Some server's will send new session tickets immediately after the handshake is complete. If they do, then the code will decrypt and parse the session ticket messages.

This PR also modifies `service_detection()` to try using sockets if the server only supports TLS 1.3 and $OPENSSL does not support TLS 1.3. After the handshake is complete, this code sends an HTTP GET request and reads the response. The code is fairly slow and it doesn't always work. However, since it is only used in cases in which $OPENSSL cannot work, it can't hurt to try using sockets.

Perhaps in the future someone can try to figure out why the new code in `service_detection()` doesn't work with some servers.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -258,6 +258,7 @@ TLS12_CIPHER_OFFERED=""                 # This contains the hexcode of a cipher 
 CURVES_OFFERED=""                       # This keeps which curves have been detected. Just for error handling
 KNOWN_OSSL_PROB=false                   # We need OpenSSL a few times. This variable is an indicator if we can't connect. Eases handling
 DETECTED_TLS_VERSION=""                 # .. as hex string, e.g. 0300 or 0303
+APP_TRAF_KEY_INFO=""                    # Information about the application traffic keys for a TLS 1.3 connection.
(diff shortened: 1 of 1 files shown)


drwetter:
Awesome

I tried with the container ~/utils/docker-debian10.tls13only.start.sh (uses all common TLS 1.3 ciphers TLS_AES_256_GCM_SHA384 TLS_CHACHA20_POLY1305_SHA256 TLS_AES_128_GCM_SHA256, doesn't work yet). 
Will try other servers later.  

A good step forward!

PR #1488, opened 2020-01-31
Fix client simulation

dcooper16:
When making the changes in #1482, I completely forgot that the variable names in run_client_simulation() also appear in etc/client-simulation.txt. This PR fixes the problem by making the appropriate changes in etc/client-simulation.txt, etc/client-simulation.wiresharked.txt, and utils/update_client_sim_data.pl.


drwetter:
Ok, thanks. Travis-Ci looked good here.

In 3.0 we're fine, right?

dcooper16:
> In 3.0 we're fine, right?

Yes. #1482 didn't change 3.0, so 3.0 didn't get broken.

PR #1492, opened 2020-02-04
Fix typos

dcooper16:
I noticed two places that had a single "&" rather than "&&". I didn't try to do any testing, but these certainly seem like typos.


drwetter:
Thanks, David. That could be also good to have it backported to 3.0. Let me know pls if I rather should take care of that.

PR #1495, opened 2020-02-04
Fix public key length calculation

dcooper16:
This PR fixes a minor bug in `get_pub_key_size()`. If the key size is being determined manually and length encoding requires 4 bytes, then the current code computes the length incorrectly. This is a very insignificant bug, since it does not apply to RSA or ECC keys, and the key would have to be at least 16 megabytes long for it to require 4 bytes to encode.

This PR also cleans up `get_pub_key_size()` a bit by replacing `i=$i+...` with `i+=...` and by enclosing some math in `$(( ... ))`.


drwetter:
Thanks

PR #1496, opened 2020-02-06
Log cookie names without HttpOnly or secure flags

mkauschi:
This PR adds the functionality to print the Cookie names, which do not have the HttpOnly or secure flag set, in the output file.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3022,6 +3022,16 @@ run_cookie_flags() {     # ARG1: Path
           else
                fileout "cookie_httponly" "INFO" "$nr_secure/$nr_cookies at \"$1\" marked as HttpOnly$msg302_"
           fi
+          IFS=$'; ' read -r -a not_secure <<< $(echo $(awk -F"=" '{print$1}' <<< $(awk -F" " '{print$2}' <<< $(grep -iav secure $TMPFILE))))
+          if [[ -v not_secure ]]; then
+               not_secure_cookies=$(printf '%s,' "${not_secure[@]}")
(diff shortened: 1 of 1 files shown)


mkauschi:
I can't figure out why the test is failing. Do you have any suggestions?

drwetter:
> I can't figure out why the test is failing. Do you have any suggestions?

not sure. You can run` /t08_isHTML_valid.t` from the terminal.

``read -r -a not_secure <<< $(awk -F"=" '{print$1}' <<< $(awk -F" " '{print$2}' <<< $(grep -iav secure $TMPFILE)))`` seems rather ugly to me. Preferred solution would be with bash only

mkauschi:
> > I can't figure out why the test is failing. Do you have any suggestions?
> 
> not sure. You can run` /t08_isHTML_valid.t` from the terminal.

I will try that.

> `read -r -a not_secure <<< $(awk -F"=" '{print$1}' <<< $(awk -F" " '{print$2}' <<< $(grep -iav secure $TMPFILE)))` seems rather ugly to me. 

I will improve the code by introducing variables and add some comments. I'm also always open for improvement suggestions.

> Preferred solution would be with bash only

You mean because I used grep to find the correct lines? I tried to be in line with the rest of the "cookie parsing" code and there is grep used as well. But I can see if I find a better solution.

drwetter:
> You mean because I used grep to find the correct lines

The combination of grep and awk. Maybe `read` works here (like `read a b c <<< "Hello you here"`)

Rough coding style: https://github.com/drwetter/testssl.sh/blob/docu_update/Coding_Convention.md

PR #1497, opened 2020-02-06
neat_list() and export

dcooper16:
Currently, the function neat_list() uses the variable "export", but does not define it. The result is that "export" variable in the calling function is used.

This PR fixes that by defining "export" as a local variable in neat_list() and by setting its value via a new parameter to the function.

This PR also removes a "FIXME" from run_rc4() since the problem has already been fixed.


drwetter:
HI David,

thanks!

Dirk

PS: That one probably goes on me, but while looking at it today "export" as a variable seems not a good choice to me as it is close to the bash builtin export (readability).

PR #1499, opened 2020-02-10
Fix printing percent characters

dcooper16:
As noted in #1481, testssl.sh has a problem with printing percent ('%') characters.

At one point, the function out() was implemented as `/usr/bin/printf -- "${1//%/%%}"`. When this was the case, any '%' needed to be replaced with '%%' since '$1' was being used as the format string. This was changed, however, by https://github.com/drwetter/testssl.sh/commit/8a2fe5915a8d4f70ba423f98eff02a6646935aa9. Since the format string is now "%b" rather than '$1', the replacement is not needed anymore. Instead, the replacement now causes any '%' to be printed to be duplicated.

This problem does not happen very often, but does sometimes occur when a '%' character appears in a URI, such as in an HTTP redirect, a certificate revocation list, or an OCSP URI.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -501,17 +501,16 @@ html_reserved(){
 
 html_out() {
      "$do_html" || return 0
-     [[ -n "$HTMLFILE" ]] && [[ ! -d "$HTMLFILE" ]] && printf -- "%b" "${1//%/%%}" >> "$HTMLFILE"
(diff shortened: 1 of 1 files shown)


drwetter:
Hi David,

this conflicts with #1481. Could you have a look please?

dcooper16:
> this conflicts with #1481. Could you have a look please?

Fixed.

drwetter:
Thanks. A backport here would be appreciated, too.

PR #1501, opened 2020-02-12
Wide output for cipher order

dcooper16:
Since, in cases in which the server enforces a cipher order, both `run_cipher_per_proto()` and `run_server_preference()` list every cipher supported by the server for each protocol, there was a discussion at one point about eliminating `run_cipher_per_proto()` and extending `run_server_preference()`.

This PR takes a step in that direction by providing the option to present the "Cipher order" in wide mode.

Note that there is one "limitation" to this PR. According to the --help output:
```
--show-each                   for wide outputs: display all ciphers tested -- not only succeeded ones
```
This PR ignores the --show-each flag in `cipher_pref_check()`. Should that be changed, and if so, how? List all of the unsupported ciphers (ordered by strength) after listing all of the supported ciphers (order by server preference)?

Also, if this PR is accepted, then I may need to follow up by changing some of the documentation for `--wide` mode to indicate that it applies to "Cipher order" as well.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6494,8 +6494,10 @@ run_server_preference() {
           [[ $TLS_NR_CIPHERS == 0 ]] && using_sockets=false
 
           pr_bold " Cipher order"
+          "$WIDE" && outln && neat_header
           while read proto_ossl proto_hex proto_txt; do
                [[ "$proto_ossl" == tls1_3 ]] && ! "$has_tls13_cipher_order" && continue
+               "$WIDE" && pr_underline "$proto_txt"
                cipher_pref_check "$proto_ossl" "$proto_hex" "$proto_txt" "$using_sockets"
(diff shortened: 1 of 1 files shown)


drwetter:
Hi David,

thanks for tackling this! It is actually one of the features we discussed, yes. 

> This PR ignores the --show-each flag in cipher_pref_check(). Should that be changed, and if so, how?

I would say, we just ignore it here. That is something for ``run_allciphers()`` and ``run_cipher_per_proto()``. That's also something we need to discuss: do we still need any of these two in a default run? Do we want rather ``run_server_preference()`` to run per default in wide mode instead?

Also: The information value of  _Negotiated protocol_ and _Negotiated cipher_ maybe seems not clear to the user.

> Also, if this PR is accepted, then I may need to follow up by changing some of the documentation for --wide mode to indicate that it applies to "Cipher order" as well.

Go ahead if you feel to. ;-) I still need some time for other matters.

Cheers, Dirk

PR #1507, opened 2020-02-14
Remove non-printables in rDNS when pretty printing JSON (fixes #1506)

### LINKED ISSUE #1506: Unescaped characters in rDNS key of JSON pretty print output
- **tl;dr: `rDNS` value contains Unicode control characters which causes JSON parsing to fail.**
- **ONLY happens when pretty printing!**
- *~This is probably a bug in `dig`, which can be solved by proper sanitization.~*

---

1. uname -a
`Darwin hampers.local 18.7.0 Darwin Kernel Version 18.7.0: Thu Jan 23 06:52:12 PST 2020; root:xnu-4903.278.25~1/RELEASE_X86_64 x86_64`
   
2. testssl version from the banner: testssl.sh -b 2>/dev/null | head -4 | tail -2
```
testssl.sh       3.0 from https://testssl.sh/
    (567450e 2019-11-19 15:13:14 -- )
```

3. git log | head -1 (if running from git repo)
`b0b084dcda595ae6f4c9a23acf32183042d03816`

4. openssl version used by testssl.sh: testssl.sh -b 2>/dev/null | awk -F':' '/openssl/ { print $2}'
`Using "OpenSSL 1.0.2-chacha (1.0.2k-dev)" [~183 ciphers]`

5. steps to reproduce: testssl.sh or docker command line, if possible incl. host
```bash
# This only happens with some host names, e.g., www.beverlyhills.org
$> testssl.sh -s --jsonfile bh.json www.beverlyhills.org
# Try to parse JSON using jq (any other parser yields same error)
$> jq '.' bh.json
parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line 12, column 422
# Looking for the culprit
$> pcregrep -o "[\x00-\x1F]" bh.json | hexdump
0000000 1a 0a                                          
0000002
```
A closer look reveals that at some point in `rDNS` string a [substitute character](https://en.wikipedia.org/wiki/Substitute_character), i.e., `^Z` appears.
 
6. what exactly was happening, output is needed
See above.

7. what did you expect instead?
Proper data sanitization  before outputting JSON.

yan-foto:
In some rare cases when results are pretty printed in JSON format,
non-printable characters, e.g., control characters, appear in rDNS
output (see #1506). Simple data sanitization solves this.

**NOTE**: this PR does not explain why such error happens in first place, so it is rather a workaround than a proper solution.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1092,7 +1092,7 @@ fileout_json_finding() {
                     \"ip\"              : \"$NODEIP\",
                     \"port\"            : \"$PORT\",
                     \"rDNS\"            : \"$rDNS\",
-                    \"service\"         : \"$finding\"," >> "$JSONFILE"
+                    \"service\"         : \"$finding\"," | tr -dc '[:print:]' >> "$JSONFILE"
                $do_mx_all_ips && echo -e "                    \"hostname\"        : \"$NODE\","  >> "$JSONFILE"
           else
                ("$FIRST_FINDING" && echo -n "                            {" >> "$JSONFILE") || echo -n ",{" >> "$JSONFILE"


drwetter:
Hi @yan-foto,

thanks for the bug report and for the PR. 

The cause in fact seems to be the non-printable characters: ``dig -x 192.5.3.33 +timeout=1 +tries=2 +noall +answer +short`` returns also ``beverlyhillsfiber.net\032.``, so does ``host -t PTR 192.5.3.33``.

However here seems to be a an issue which this PR would be not fix, therefore I am rather closing it. Also ``testssl.sh -s --jsonfile bh.json www.beverlyhills.org`` does produce valid JSON for me. The terminal output is strange though, 

```Start 2020-02-14 19:27:50        -->> 192.5.3.33:443 (www.beverlyhills.org) <<--

 rDNS (192.5.3.33):      bhfiber.net.www.bhalt.net.joinbhpd.org.beverlyhillsfiber.org.bhrent-api.beverlyhills.org.bhfiber.org.beverlyhillsfiber.net▒.www.bhgov.net.rent.beverlyhills.org.beverlyhillsarts.org.beverlyhillsfiber.net.joinbhpd.com.drivenbyinnovation.net.www.bhalt.xyz.www.bhalt.org.data.beverlyhills.org.www.beverlyhills.org.bbhfiber.net.bhgov.org.drivenbyinnovation.org.bbfiber.net.bhfiber.com.
 Service detected:       HTTP
```

Cheers, Dirk
There's are missing spaces here

yan-foto:
Hi Dirk,

thanks for the quick reply and taking time to review the PR.. I was hoping that the PR would be productive (which it obviously wasn't 😅 ).

> However here seems to be a an issue which this PR would be not fix.

Do you mean because it is caused by `dig` and not `testssl.sh` or am I missing something else here?

> Also `testssl.sh -s --jsonfile bh.json www.beverlyhills.org` does produce valid JSON for me.

Yes it does, because it does not print rDNS info at all (at least not for me on `v3.0`).

drwetter:
PS & FYI:The pattern  ``'[:print:]' `` works also for FreeBSD9 (and likley MacOSX)  and older OpenBSDs which is good but I didn't expected that. If you have the chance to test that, just mention it. That would save work on my side. Default is otherwise I try to make sure, especially for changes in tr and sed & Co I'll check it.

drwetter:
> thanks for the quick reply and taking time to review the PR.. I was hoping that the PR would be productive (which it obviously wasn't ).

Sure, I appreciate not only bug reports but even more initiatives fixing problems

yan-foto:
> PS & FYI:The pattern `'[:print:]'` works also for FreeBSD9 (and likley MacOSX) and older OpenBSDs which is good but I didn't expected that. If you have the chance to test that, just mention it. That would save work on my side. Default is otherwise I try to make sure, especially for changes in tr and sed & Co I'll check it.

@drwetter I am on macOS (Mojave) and `[:print:]` with `tr` works just as expected. I haven't tested it on any othe *BSD based system though.

Let me know if you need any further testing on macOS.

drwetter:
>> However here seems to be a an issue which this PR would be not fix.
>Do you mean because it is caused by `dig` and not `testssl.sh` or am I
>missing something else here?

Both dig and host produce a PTR record with \32. You can call this a bug or feature. In the end it seems to lead to this odd output in testssl.sh which need to be fixed.



-- 

yan-foto:
> Both dig and host produce a PTR record with \32. You can call this a bug or feature. In the end it seems to lead to this odd output in testssl.sh which need to be fixed.

I am getting convinced that it's not a bug rather a DNS misconfiguration: I think the PTR RRs are not set correctly as

```bash
$> drill -x ptr  192.5.3.33 2>/dev/null
```
would return exactly the same result and yet again `\032` is only present by `beverlyhillsfiber.net`. I don't suppose that it is just by chance since `dig` and `drill` don't use the same code base. Does that make sense to you?

teward:
I ran a straight PyDNS check on these and spat out the resulting PTR entries.  I can confirm that one of the PTR entries for `beverlyhillsfiber.net` has an errant byte code in it.  This said, there IS a valid DNS entry without the bytes matching the straight domain and another 21 PTR entries as well, it's possible this is just one PTR that got misentered or miscopied when pushing into DNS.

This isnt a bug in testssl.sh nor necessarily dig or drill, but someone doing badness in the DNS entry (either intentionally or by accident).

yan-foto:
@teward this confirms my assumption and raises the question all JSON values should be sanitized to avoid faulty DNS, etc. values.

drwetter:
let's continue the discussion please @ #1506

PR #1516, opened 2020-02-20
Fix use of HSTS_MIN

dcooper16:
This PR fixes two minor issues related to HSTS_MIN:

* If there is a misconfiguration the recommended max-age should be based on $HSTS_MIN rather than being hardcoded to 15552000 seconds = 180 days.

* If max-age is exactly $HSTS_MIN, testssl.sh shouldn't say that max-age is too short while also saying that >= $HSTS_MIN seconds is recommended.


drwetter:
good point, thanks

PR #1518, opened 2020-02-27
Make colors in HTML output more accessible

dcooper16:
I recently discovered the [W3C's Web Content Accessibility Guidelines (WCAG).](https://www.w3.org/WAI/WCAG21/quickref/?versions=2.0) One of the requirements is that when using colors there must be sufficient contrast between the font color and the background color. For a white background, this basically means the colors need to be dark enough.

According to the [Accessible Color Palette Builder](https://toolness.github.io/accessible-color-matrix/), almost all of the colors that testssl.sh uses in its HTML output fail to meet the WCAG's minimum requirements for contrast. :-(

- [pr_svrty_best, pr_svrty_good](https://toolness.github.io/accessible-color-matrix/?n=white&n=best&n=good&n=black&v=FFFFFF&v=00FF00&v=00CD00&v=000000)
- [pr_svrty_low, pr_svrty_medium, prln_svrty_high, pr_svrty_critical](https://toolness.github.io/accessible-color-matrix/?n=white&n=black&n=low&n=medium&n=high&n=critical&v=FFFFFF&v=000000&v=CDCD00&v=CD8000&v=CD0000&v=FF0000)
- [pr_warning, pr_magenta, pr_litecyan, pr_cyan, html_yellow (olive)](https://toolness.github.io/accessible-color-matrix/?n=white&n=warning&n=magenta&n=litecyan&n=cyan&n=olive&v=FFFFFF&v=CD00CD&v=FF00FF&v=00CDCD&v=00FFFF&v=808000)
- [pr_grey, pr_deemphasize](https://toolness.github.io/accessible-color-matrix/?n=white&n=grey&n=dark%20grey&v=FFFFFF&v=7F7F7F&v=A9A9A9)

This PR addresses this issue by changing the colors used in HTML output to ones that meet the WCAG's minimum requirements for contrast. For new colors, I selected colors from https://designsystem.digital.gov/design-tokens/color/system-tokens. This web page assigns a grade to each color, and as explained in https://designsystem.digital.gov/design-tokens/color/overview, any color with a grade 50 or higher meets the WCAG's minimum requirements for contrast when used on a white background.

The picture below shows the colors that testssl.sh currently and the colors that this PR proposes to use. In some cases testssl.sh's color came very close to meeting the requirements, and so the proposed color is almost the same as the current color. In other cases, the current color is very light and so the proposed color differs substantially.

colors2

In developing this PR, I noticed that some print functions are not used at all: *_blue, *_liteblue, *_litegrey, *_squoted, *_dquoted. This PR deletes these. The *_strikethru functions are also not used, but I didn't delete these in case they may be needed in the future (since it may not be obvious how to do strike through in the terminal).

The choice of colors for each of these functions is, of course, subjective. So, I'd welcome any suggestions for changes to the proposed colors.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -530,15 +530,6 @@ outln() { printf -- "%b" "$1\n"; html_out "$(html_reserved "$1")\n"; }
 #TODO: Still no shell injection safe but if just run it from the cmd line: that's fine
 
 # Color print functions, see also https://www.tldp.org/HOWTO/Bash-Prompt-HOWTO/x329.html
-tm_liteblue()   { [[ "$COLOR" -ge 2 ]] && ( "$COLORBLIND" && tm_out "\033[0;32m$1" || tm_out "\033[0;34m$1" ) || tm_out "$1"; tm_off; }    # not yet used
-pr_liteblue()   { tm_liteblue "$1"; [[ "$COLOR" -ge 2 ]] && ( "$COLORBLIND" && html_out "<span style=\"color:#00cd00;\">$(html_reserved "$1")</span>" || html_out "<span style=\"color:#0000ee;\">$(html_reserved "$1")</span>" ) || html_out "$(html_reserved "$1")"; }
-tmln_liteblue() { tm_liteblue "$1"; tmln_out; }
-prln_liteblue() { pr_liteblue "$1"; outln; }
-
-tm_blue()       { [[ "$COLOR" -ge 2 ]] && ( "$COLORBLIND" && tm_out "\033[1;32m$1" || tm_out "\033[1;34m$1" ) || tm_out "$1"; tm_off; }    # used for head lines of single tests
-pr_blue()       { tm_blue "$1"; [[ "$COLOR" -ge 2 ]] && ( "$COLORBLIND" && html_out "<span style=\"color:lime;font-weight:bold;\">$(html_reserved "$1")</span>" || html_out "<span style=\"color:#5c5cff;font-weight:bold;\">$(html_reserved "$1")</span>" ) || html_out "$(html_reserved "$1")"; }
-tmln_blue()     { tm_blue "$1"; tmln_out; }
-prln_blue()     { pr_blue "$1"; outln; }
 
 # we should be able to use aliases here
 tm_warning()    { [[ "$COLOR" -ge 2 ]] && tm_out "\033[0;35m$1" || tm_underline "$1"; tm_off; }                   # some local problem: one test cannot be done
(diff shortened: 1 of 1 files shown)


dcooper16:
If anyone wants to experiment with the color choices outside of testssl.sh, here is the HTML file I used to create the picture in the previous comment.
```
<?xml version="1.0" encoding="UTF-8" ?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
<meta http-equiv="Content-Type" content="application/xml+xhtml; charset=UTF-8" />
<title>testssl.sh</title>
</head>
<body>
<pre>

<table>
  <tr>
    <th>Current Color</th> 
    <th>Proposed Color</th>
  </tr>
  <tr>
    <td><span style="color:lime;font-weight:bold;">prln_svrty_best</span></td>
    <td><span style="color:#008817;font-weight:bold;">prln_svrty_best (green-cool-50v bold)</span></td>
  </tr>
  <tr>
    <td><span style="color:#00cd00;">prln_svrty_good</span></td>
    <td><span style="color:#008817;">prln_svrty_good (green-cool-50v)</span></td>
  </tr>
  <tr>
    <td>outln</td>
    <td>outln (no change)</td>
  </tr>
  <tr>
    <td><span style="color:#cdcd00;font-weight:bold;">prln_svrty_low</span></td>
    <td><span style="color:#a86437;">prln_svrty_low (orange-50)</span></td>
  </tr>
  <tr>
    <td><span style="color:#cd8000;">prln_svrty_medium</span></td>
    <td><span style="color:#c05600;">prln_svrty_medium (orange-50v)</span></td>
  </tr>
  <tr>
    <td><span style="color:#cd0000;">prln_svrty_high</span></td>
    <td><span style="color:#e52207;">prln_svrty_high (red-50v)</span></td>
  </tr>
  <tr>
    <td><span style="color:red;font-weight:bold;">prln_svrty_critical</span></td>
    <td><span style="color:#e52207;font-weight:bold;">prln_svrty_critical (red-50v bold)</span></td>
  </tr>
  <tr>
    <td><span style="color:#cd00cd;">prln_warning</span></td>
    <td><span style="color:#cd00cd;">prln_warning (no change)</span></td>
  </tr>
  <tr>
    <td><span style="color:magenta;font-weight:bold;">prln_magenta</span></td>
    <td><span style="color:#be32d0;font-weight:bold;">prln_magenta (violet-warm-50v bold)</span></td>
  </tr>
  <tr>
    <td><span style="color:#00cdcd;">prln_litecyan</span></td>
    <td><span style="color:#168092;">prln_litecyan (cyan-50)</span></td>
  </tr>
  <tr>
    <td><span style="color:cyan;font-weight:bold;">prln_cyan</span></td>
    <td><span style="color:#0d7ea2;font-weight:bold;">prln_cyan (blue-cool-50v bold)</span></td>
  </tr>
  <tr>
    <td><span style="color:#7f7f7f;font-weight:bold;">prln_grey</span></td>
    <td><span style="color:#757575;font-weight:bold;">prln_grey (gray-50 bold)</span></td>
  </tr>
  <tr>
    <td><span style="color:darkgray;">prln_deemphasize</span></td>
    <td><span style="color:#71767a;">prln_deemphasize (gray-cool-50)</span></td>
  </tr>
  <tr>
    <td><span style="color:olive;">html_brown</span></td>
    <td><span style="color:#8a7237;">html_brown (yellow-50)</span></td>
  </tr>
  <tr>
    <td><span style="color:olive;font-weight:bold;">html_yellow</span></td>
    <td><span style="color:#8a7237;font-weight:bold;">html_yellow (yellow-50 bold)</span></td>
  </tr>
</table>
</pre>
</body>
</html>

```

drwetter:
Hi David,

interesting, I didn't know about that W3C standard either! 

Any step further with colors is much appreciated! First thing which comes to my mind: Should we only change that in HTML? What about the terminal? 

> In developing this PR, I noticed that some print functions are not used at all: *_blue, *_liteblue, *_litegrey, *_squoted, *_dquoted. 

To my concern ;-) this is true. At some point of time I thought to look in general at the colors again as I feel on the terminal we didn't use some colors we could make more use of. E.g. cyan.

> This PR deletes these.

Hmm...

My first take is that it's good to start with improving the colors. Maybe it's better to also think where we want to go with the colors on the terminal, at this moment maybe we don't have to implement that but we should have a goal here too 

What you think about the terminal colors.?

Thanks, Dirk

dcooper16:
> > In developing this PR, I noticed that some print functions are not used at all: *_blue, *_liteblue, *_litegrey, *_squoted, *_dquoted.
> 
> To my concern ;-) this is true. At some point of time I thought to look in general at the colors again as I feel on the terminal we didn't use some colors we could make more use of. E.g. cyan.

I don't think the functions that I removed are much of an issue. The colors created by *_blue and *_liteblue are found in *_svrty_good and *_svrty_best. The functions just reverse which is used when $COLORBLIND is true versus when it is false.

*_litegrey is very similar to *_deemphasize.

_squoted and *_dquoted don't use color, they just place quotes around the text, and it is just easy to include the quotes in the text provided to the print function.

Perhaps testssl.sh could make use of more colors, but I thought going forward the preference was to use names like pr_svrty_best rather than pr_darkgreen.

> 
> My first take is that it's good to start with improving the colors. Maybe it's better to also think where we want to go with the colors on the terminal, at this moment maybe we don't have to implement that but we should have a goal here too
> 
> What you think about the terminal colors.?

Working on the HTML colors was relatively easy. As https://htmlcolorcodes.com notes: "With modern browsers supporting the full spectrum of 24-bit color, there are 16,777,216 different color possibilities." Not only are there so many colors to choose from:
-  the [Accessible Color Palette Builder](https://toolness.github.io/accessible-color-matrix/) was available to tell me whether any given color was accessible on a white background.
- https://designsystem.digital.gov/design-tokens/color/system-tokens provided examples of many different colors that are known to be accessible.
- Most importantly, the colors seem to look about the same on different browsers and on different platforms.

The situation with the terminal is entirely different. If I understand correctly from https://en.wikipedia.org/wiki/ANSI_escape_code#Colors, testssl.sh currently uses the 3/4 bit color codes. As the table on the web page shows, these colors are _not_ displayed the same on different platforms. In my own experience, when using Konsole on a desktop running Kubuntu, *_svrty_low prints as yellow, and I cannot read it. On the MacBook, *_svrty_low prints as a yellowish-brown that is fairly easy to read.

https://en.wikipedia.org/wiki/ANSI_escape_code#Colors mentions 8- and 24-bit colors, but I thought in the past you had said that these were not very portable, as they were not supported on all platforms. If they are supported, and their colors are displayed consistently on different platforms, that would help, but I can only test on a couple of different platforms.

Another complication for the terminal that you mentioned in https://github.com/drwetter/testssl.sh/pull/600#issuecomment-276129876 is that not everyone uses the "black on white" terminal scheme. For HTML, we just need to ensure that the text looks okay on a white background. Just within Konsole, the terminal output looks very different depending on which terminal color scheme I choose. Whereas I cannot read the *_svrty_low output with the "black on white" color scheme, it is very easy to read with most of the color schemes that use a black background. Some of the color schemes change the colors in ways the seem to destroy the information we are trying to use colors to convey.

So, while it would be nice to do something with the terminal colors, that would be a much more difficult task.

drwetter:
Hi David,

thanks for the detailed answer.

My question was more simple: Are we now caring about HTML output only and worry about the terminal maybe later. Or should we made little adjustments or do we have a picture for the terminal?

I personally don't have a clear picture wrt the terminal output but ideas. I would love to use more colors at some point and probably try 16bit colors where supported. For the current branch 16bit is probably out of scope for me.

Another detail to the terminal colors you mentioned: yes, the background color is at least is something which is hard to deal with. (Don't know whether Kubuntu and the Mac terminal have the same background or whether that's another point).


Coming to my original point: If I understand you correctly you see HTML decoupled from the terminal?

Cheers, Dirk

dcooper16:
> If I understand you correctly you see HTML decoupled from the terminal?

To some degree. At least when comparing the HTML to with a "black on white" color scheme, I think the HTML and the terminal should follow the same color theme. Good is indicated by green, with best being in bold. Bad outputs are colored from yellow (yellow-red) to red, with redder colors indicating worse.

Perhaps if we move away from the current 3/4 bit colors, we can make move towards using the same colors on the terminal and in the HTML. https://en.wikipedia.org/wiki/ANSI_escape_code#Colors seems to suggest that, unlike with the 3/4 bit colors, the 8-bit colors specify specific RGB values, and so there is a chance that an 8-bit color code will result in the same color being printed no matter what terminal is being used. With 256 colors to choose from, it might also be reasonable to choose colors that are dark enough to be readable and that convey the meaning we want.

At the moment, however, given that the choices for the terminal are so limited, I think it is okay to make some changes to the HTML output without making corresponding changes to the terminal output. This PR just tries to change the HTML colors just enough to make them accessible without diverging too much from the colors that might appear on the screen.

As for the idea of using a new color in the HTML without also using a new color on the terminal, we already do this with `pr_deemphasize()`. We couldn't find a way to "deemphasize" text on the terminal that didn't make it unreadable on some terminals (see https://github.com/drwetter/testssl.sh/pull/600#issuecomment-276129876). So, `pr_deemphasize()` just calls `tm_out()` to print to the terminal, but specifies a color of `darkgray`. The result being that deemphasized text can be distinguished from regular text in the HTML, but not on the terminal.

Do that with `pr_deemphasize()` wasn't ideal, but I don't think we should entirely rule out doing it again. If there is something we can do to make the HTML output a little better that we can't also do with the terminal output, it should be considered on a case-by-case basis.

drwetter:
still need to look into that, sorry!

dcooper16:
> still need to look into that, sorry!

No problem. This is just something to think about. It doesn't need to be changed anytime soon.

drwetter:
Hi David,

before I come up with a big solution which then also changes the output on the screen I think we should go forward with this.

I would love however to keep the ones which you deleted, not only the ones which use the ``--colorblind`` option.

drwetter:
Thanks a lot (and sorry). 

Waiting until Travis finishes...

PR #1529, opened 2020-03-06
Fix printing percent characters

dcooper16:
This commit makes the same change as #1499, but in the 3.0 branch.


drwetter:
Thanks!

PR #1531, opened 2020-03-06
Fix typo in emphasize_stuff_in_headers()

dcooper16:
This commit fixes a typo in emphasize_stuff_in_headers() where ${yellow} was used rather than ${html_yellow} in the creation of the HTML output.


drwetter:
good catch!

PR #1532, opened 2020-03-06
Fix typo in emphasize_stuff_in_headers()

dcooper16:
This commit fixes the same typo as #1531, but in the 3.0 branch.


drwetter:
thanks

PR #1538, opened 2020-03-17
Fix basicauth bug where a newline is added to the user:password string before encoding

mkauschi:
I have found a bug in the implementation of the basicauth feature which was probably introduced in commit https://github.com/drwetter/testssl.sh/pull/1452/commits/4603d924be47959a8734b5245ed8a4a1b3fd38df. 

The problem here is that the command "openssl base64 <<< $BASICAUTH" adds a newline at the end of the string which results in a differen base64 encoding.

Example:

openssl base64 <<< "user:password" -> dXNlcjpwYXNzd29yZAo=

However, it should rather look like the following.

echo -n "user:password" | openssl base64 -> dXNlcjpwYXNzd29yZA==

I don't see a better solution without using the "echo" command here. If you have better suggestions, let me know.

Best regards

Manuel

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -19606,7 +19606,7 @@ determine_service() {
                ua="$UA_SNEAKY" || \
                ua="$UA_STD"
           if [[ -n "$BASICAUTH" ]]; then
-               basicauth_header="Authorization: Basic $($OPENSSL base64 <<< "$BASICAUTH" 2>/dev/null)\r\n"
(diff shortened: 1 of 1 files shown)


drwetter:
good point :-) Thanks!

Can we settle on the builtin ``safe_echo``? A backport patch for 3.0 would be much appreciated, too

mkauschi:
> 
> Can we settle on the builtin `safe_echo`? 

Thanks, I have changed this.

> A backport patch for 3.0 would be much appreciated, too

I will do that ;-)

PR #1540, opened 2020-03-18
Redirect printf error output to /dev/null

mkauschi:
When scanning a URL which includes a percentage character,  one of the printf calls generates an error.  I have fixed this issue by redirecting stderr to /dev/null.

Example of the bug:
When calling testssl.sh with: 

./testssl.sh https://bruzzzla.de/index.php?test=% 

the following error occures:
/testssl.sh: line 2074: printf: `H': invalid format character

This PR fixes the issue.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2071,7 +2071,7 @@ service_detection() {
                fi
           else
                # SNI is not standardized for !HTTPS but fortunately for other protocols s_client doesn't seem to care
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks for the detection! However your fix would just sweep the erroneous handling under the carpet. Please issue a new PR.

Could you change that to ``tmln_out`` please? And there seem to be more than one instance, I counted 4.

PR #1541, opened 2020-03-19
Better Printing

NetSpida:
Fix printing of special characters, especially NULL-bytes in server banner.
Analog to #695, enhance printing of long lines at various places.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1412,6 +1412,25 @@ out_row_aligned() {
      done <<< "$1"
 }
 
+# prints out a text, trying to align it to the right border if it is too long.
+print_aligned_right_max_width() {
+     local text="$1"
+     local spaces="$2"
+     local -i max_width="$3"
+     local -i i len
+     local print_function="$4"
+     len=${#text}+${#spaces}
+     if [[ $len -lt $max_width ]]; then
+          # If the text to print is shorter than $max_width,
+          # then just print it using print_function.
+          out "$spaces"
+          $print_function "$text"
+     else
+          # try to right-align url
+          printf "%${max_width}s" "$text"
+     fi
+}
+
 # prints text over multiple lines, trying to make no line longer than $max_width.
 # Each line is indented with $spaces.
 out_row_aligned_max_width() {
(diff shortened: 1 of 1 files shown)


dcooper16:
Hello @NetSpida,

I took a look at your PR, and unfortunately, the code that you have written to enhance line wrapping will not work. The code you wrote may look as you want in the terminal. However, if the user asks for an HTML file to be created, the HTML file will not be correct.

testssl.sh generally uses two types of print functions. Functions such as `out`, `outln`, `pr_italic`, `pr_svrty_best`, etc. print to both the terminal and the HTML file. Functions such as `tm_out`, `tm_italic`, `tm_svrty_best`, etc. only print to the terminal. testssl.sh is designed to print some information only to the terminal (e.g., debugging output), but in general the HTML output should look exactly the same as the terminal output.

Given this, there are some easy to fix and some very difficult to address problems in this PR.

A simple issue is that `print_aligned_right_max_width()` calls `printf "%${max_width}s" "$text"` if the line to be printed is too long. By calling `printf` rather than `$print_function` (which is `out` or `pr_url`), the `$text` gets printed to the terminal, but not to the HTML file.

The part that is difficult to address is places in which the PR tries to capture formatted text for later printing. For example
```
cntext="$(pr_italic "$cn")"
```
This code results in `$cn` being printed to the HTML file immediately, but not to the terminal. Later, when you call
```
out_row_aligned_max_width "$cntext" "$spaces" $TERM_WIDTH
```
`$cntext` is printed to the terminal, but not to the HTML file. Perhaps that could be made to work, but other places where the code just does `cntext="($cn)"`, `$cn` will eventually get printed to the terminal, but it will not be printed to the HTML file.

At one point, testssl.sh did have code of the form `cntext="$(pr_italic "$cn")"`, where `$cntext` was printed later. However, I had to get rid of all such calls when I wrote the code to generate HTML output.

The failure that Travis is reporting is from a check that the HTML output is the same as the terminal output. At the very least, that needs to be fixed. However, the Travis check will not notice if text is in italics in the terminal, but not in the HTML. So, it would  be a good idea when you are testing your code to have testssl.sh create HTML output and manually compare the resulting HTML to the terminal output.

drwetter:
Hi @NetSpida ,

thanks! Just feel free to amend the PR.. (I like your programming style)

Cheers, Dirk

drwetter:
clsosing due to the lack of feedback

PR #1542, opened 2020-03-23
Handle printf errors for URLs which include a percent character

mkauschi:
When scanning a URL which includes a "%" character testssl.sh throws an error and returns with that error in stderr. In order to avoid this behaviour the error is now logged out to the console.

Some words about the implementation:
It was suggested to use the function tmln_out in order to print the error to the console. This was not possible because stderr is a stream that could have more lines of errors. Therefore I implemented the function `stderr_out` to print those errors. 

The solution uses process substitution to redirect the stderr to the new function `stderr_out` and redirect the stdout to the intended command.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -852,6 +852,13 @@ is_ipv6addr() {
      return 0
 }
 
+stderr_out() { 
+while read line
+    do
+        printf -- "%b" "$line\n"
+    done
+ }
+
 ###### END universal helper function definitions ######
 
 ###### START ServerHello/OpenSSL/F5 function definitions ######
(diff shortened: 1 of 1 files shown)


dcooper16:
Hi @mkauschi,

I believe that you misunderstood Dirk's suggestion. He wasn't suggesting to use `tmln_out` to print the error. He was suggesting to replace

```printf "$GET_REQ11"```

with something like

```tmln_out "$GET_REQ11"```

so that testssl.sh works correctly with URLs that include percent characters, and there is no error to either print or hide.

@drwetter: Is there a reason you suggested using `tmln_out` rather than `tm_out`?

mkauschi:
Oops, yes I think I misunderstood the comment. Using `tmln_out` instead of `printf` makes perfect sense. Sorry!

I will quickly change that and issue a new PR.

PR #1543, opened 2020-03-23
replace printf with built-in print function

mkauschi:
Replace printf with a built-in function


drwetter:
Hate to say it, David is right (not bc of David but bc of this PR): The line feed is not needed in ``tmln_out`` as we already have line feeds in ``$GET_REQ11``. Better is ``tm_out``.

Could you please issue a new PR pls?

PR #1544, opened 2020-03-23
Replace printf with tm_out

mkauschi:
Replace printf with tm_out in order to avoid errors with URLs which include a percent character.


drwetter:
Thanks!

Issue #1157, opened 2018-11-20
Weak 128 Bit ciphers


exploide:
I think the current rating regarding "Weak 128 Bit ciphers (SEED, IDEA, RC[2,4])" is a bit confusing and unclear to me.

Why is SEED considered a weak 128 bit cipher? I tried to research but found no known weaknesses that make this algorithm bad. The relevant point for the rating is that in TLS, SEED got only specified in CBC mode and in combination with SHA-1 KDF/MAC. And then it got removed from the standard because less researched algorithms do not bring a huge benefit.

(IDEA, on the other hand, appears to has some more attacks but not of real practical relevance. But is also operates on 64 bit blocks, so maybe it is susceptible to SWEET32 like Triple-DES.)

So, is there a reason why SEED (and maybe IDEA) is considered a weak 128 bit cipher? If not I propose to change the wording or restructure the whole cipher categories section.

A possible change would be:

- change "LOW" to  "weak algorithms", put RC2/4 into this category
- change "Weak" to "Deprecated 128 bit ciphers" for IDEA and SEED

Maybe it could be structured even better.

But at first, I'm curious about what you think and why the structure is as it currently is.

exploide:
Yeah, it's complicated to structure all this stuff...

Maybe IDEA falls into the same category as 3DES, due to the cipher with 64-bit block length thing.

SEED, on the other hand, could be omitted, since it isn't that bad to be criticized but it is also not as good as AES, especially because there is no cipher suite in TLS that combines SEED with a strong block mode or a strong hash function.

If the Weak category is rather empty then, maybe shift RC2/4 to the LOW category. They aren't much stronger than DES, aren't they?!

So in the end, maybe we have categories like

- Low: DES, RC2, RC4
- Medium: 3DES, SEED
- High: AES, Camellia
- AEAD ciphers

(Besides I'm wondering whether weakened hash functions like MD5 and SHA-1 for KDF/MAC use shoule be penalized with a yellow marking, e.g. in the cipher table. But that's a different topic and not really urgent since attack MAC algorithms is pretty hard even with bad hash functions.

drwetter:
Moved RC2, RC4 to low first. Please note that I change also for  the file output 'DES+64Bit' --> 'LOW'

drwetter:
before:

![screenshot_20190117_120957](https://user-images.githubusercontent.com/8036727/51317846-f9172a80-1a58-11e9-9852-5e6b61a751d6.png)

after:

[...]

PR #1545, opened 2020-03-23
Don't include SEED in PFS

dcooper16:
`run_pfs()` is only supposed to test for "robust" (P)FS ciphers. For this reason, ciphers that use 3DES are not included. Given that `pr_cipher_quality()` rates SEED the same as 3DES (`pr_svrty_medium`), it would seem that SEED ciphers should not be included either.

`pr_cipher_quality()` also rates IDEA ciphers as `pr_svrty_medium`, but there are no (P)FS cipher suites that use IDEA.

`run_pfs()` still includes cipher suites that are rated by `pr_cipher_quality()` as `pr_svrty_low` -- the ciphers that use CBC padding.


dcooper16:
On a somewhat related note, there is a [discussion](https://mailarchive.ietf.org/arch/msg/saag/ob6Hn6AlD8RW7RpXeBHF1fzCed4/) in the IETF about whether the property should be called "perfect forward secrecy" or just "forward secrecy."

It seems the people generally think it's okay to use either term, but think that "forward secrecy" is a bit better. Eric Rescorla noted that there was an explicit decision to use "forward secret" rather than "perfect forward secret" in RFC 8446.

drwetter:
yeah, FS is better.

Jon Callas: "Forward secrecy is a much better term. Saying "perfect" forward secrecy is kinda like "military-grade" encryption in the way it's misused, and often a shibboleth showing trying to puff things up." 

:-)

With the quantum computing more or less beyond the horizon the attribute "perfect" sounds strange.

drwetter:
Hi David,

thanks.

But maybe you can describe your (cryptography wise) problem with SEED or: where should we draw the line for ciphers to be included?

My idea of the FS section was to focus on the key exchange, then to weed out any weak encryption. I am with you regarding also throwing out the medium ciphers as well. But our stances seems different wrt medium. See also discussion @ #1157.

Cheers, Dirk

dcooper16:
> But maybe you can describe your (cryptography wise) problem with SEED or: where should we draw the line for ciphers to be included?
> 
> My idea of the FS section was to focus on the key exchange, then to weed out any weak encryption. I am with you regarding also throwing out the medium ciphers as well. But our stances seems different wrt medium. See also discussion @ #1157.

Okay, it seems that I submitted the wrong PR. I just ran a test with COLOR=3 and saw that the SEED cipher was listed in a warning color. The color is determined by `pr_cipher_quality`, and when I saw that `pr_cipher_quality` rates 3DES, IDEA, and SEED ciphers the same, I thought that since 3DES ciphers are not considered "robust," SEED ciphers shouldn't be considered "robust" either.

According to #1157, however, it seems that the problem is with `pr_cipher_quality` rather than `run_pfs`. I'll close this PR and work on a different one that tries to better align the ratings in `pr_cipher_quality` with those in `run_cipherlists`.

drwetter:
> I'll close this PR and work on a different one that tries to better align the ratings in pr_cipher_quality(0 with those in run_cipherlists()

It seems in fact I missed SEED last year when I last tried to adjust both

PR #1546, opened 2020-03-25
Fix bug in setting DISPLAY_CIPHERNAMES

dcooper16:
The permitted values for `$DISPLAY_CIPHERNAMES` are "rfc-only", "openssl-only", "openssl", and "rfc". However, `get_install_dir()` incorrectly sets `$DISPLAY_CIPHERNAMES` to "no-rfc" if it cannot find the `$CIPHERS_BY_STRENGTH_FILE`. ("no-rfc" is the string users would specify at the command line for the `--mapping` option, but not the value that `$DISPLAY_CIPHERNAMES` is set to internally).

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -17931,7 +17931,7 @@ get_install_dir() {
      fi
 
      if [[ ! -r "$CIPHERS_BY_STRENGTH_FILE" ]]; then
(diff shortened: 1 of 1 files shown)


drwetter:
:+1:

PR #1548, opened 2020-03-25
Adjust pr_cipher_quality ratings

dcooper16:
This PR makes several changes to the way that ciphers are rated by `pr_cipher_quality`:

* It upgrades SEED ciphers to be considered as strong as the corresponding AES ciphers.

* It downgrades ciphers that use AEAD, but that use a non-FS key exchange (TLS_DH_\*, TLS_ECDH_\*, TLS_PSK_WITH_\*) from best to good, thus giving them the same rating as AEAD ciphers that use static RSA (TLS_RSA_\*).

* It downgrades some CBC ciphers to low (4) that are currently rated as neither good nor bad (5).

* It modifies the ratings created using OpenSSL names to provide the same ratings as those created using RFC names.

One issue that is not addressed by the PR is the GOST ciphers. `pr_cipher_quality` doesn't really handle the GOST ciphers at the moment and so their ratings are somewhat arbitrary.

This PR also does not make any changes to `run_cipherlists`. A next step could be to change `run_cipherlists` so that its ratings are the same as `pr_cipher_quality`. At the least, this would involve separating the AEAD ciphers into those that offer FS and those that do not. I don't know whether any other changes would be required.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -5948,7 +5948,7 @@ pr_cipher_quality() {
                # We have an OpenSSL name and can't convert it to the RFC name which is rarely
                # the case, see "prepare_arrays()" and "./etc/cipher-mapping.txt"
                case "$cipher" in
-                    *NULL*|EXP*|ADH*|AECDH*)
+                    *NULL*|EXP*|ADH*|AECDH*|*anon*)
                          pr_svrty_critical "$text"
                          return 1
                          ;;
@@ -5983,7 +5983,7 @@ pr_cipher_quality() {
                          pr_svrty_best "$text"
                          return 7
                          ;; #best ones
-                    *AES*SHA*|*CAMELLIA*SHA*|*SEED*SHA*)
+                    *AES*SHA*|*CAMELLIA*SHA*|*SEED*SHA*|*CBC*)
                          pr_svrty_low "$text"
                          return 4
                          ;;
--- a/testssl.sh
+++ b/testssl.sh
@@ -5948,7 +5948,7 @@ pr_cipher_quality() {
                # We have an OpenSSL name and can't convert it to the RFC name which is rarely
                # the case, see "prepare_arrays()" and "./etc/cipher-mapping.txt"
                case "$cipher" in
-                    *NULL*|EXP*|ADH*)
+                    *NULL*|EXP*|ADH*|AECDH*)
                          pr_svrty_critical "$text"
                          return 1
                          ;;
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, David! 

Looks good but would like to test it first.
 
Cheers, Dirk

drwetter:
Looks definitely better, thanks!

> doesn't really handle the GOST ciphers at the moment and so their ratings are somewhat arbitrary.

For the better ones I personally haven't really looked at them good enough (and probably there's little known). This one smells fishy: https://en.wikipedia.org/wiki/GOST_(block_cipher). Probably this is used in 0xFF00 and 0xFF01. The first one gets a pr_svrty_high-rating. NULL ciphers should also be covered, right?

For reference those are the ones we have:

```grep GOST cipher-mapping.txt                                                                                                                                                                   3.1dev
      0x00,0x80 - GOST94-GOST89-GOST89           TLS_GOSTR341094_WITH_28147_CNT_IMIT                TLSv1      Kx=GOST        Au=GOST94  Enc=GOST(256)                  Mac=GOST89IMIT         
      0x00,0x81 - GOST2001-GOST89-GOST89         TLS_GOSTR341001_WITH_28147_CNT_IMIT                SSLv3      Kx=GOST        Au=GOST01  Enc=GOST(256)                  Mac=GOST89IMIT         
      0xFF,0x00 - GOST-MD5                       TLS_GOSTR341094_RSA_WITH_28147_CNT_MD5             TLSv1      Kx=RSA         Au=RSA     Enc=GOST(256)                  Mac=MD5                
      0xFF,0x01 - GOST-GOST94                    TLS_RSA_WITH_28147_CNT_GOST94                      TLSv1      Kx=RSA         Au=RSA     Enc=GOST(256)                  Mac=GOST94             
      0xFF,0x02 - GOST-GOST89MAC                 -                                                  TLSv1      Kx=RSA         Au=RSA     Enc=GOST(256)                  Mac=GOST89IMIT         
      0xFF,0x03 - GOST-GOST89STREAM              -                                                  TLSv1      Kx=RSA         Au=RSA     Enc=GOST(256)                  Mac=GOST89IMIT         
      0xFF,0x85 - GOST2012256-GOST89-GOST89      -                                                  SSLv3      Kx=GOST        Au=GOST01  Enc=GOST(256)                  Mac=GOST89IMIT         
      0x00,0x82 - GOST94-NULL-GOST94             TLS_GOSTR341094_WITH_NULL_GOSTR3411                TLSv1      Kx=GOST        Au=GOST94  Enc=None                       Mac=GOSTR3411          
      0x00,0x83 - GOST2001-NULL-GOST94           TLS_GOSTR341001_WITH_NULL_GOSTR3411                SSLv3      Kx=GOST        Au=GOST01  Enc=None                       Mac=GOST94             
      0xFF,0x87 - GOST2012256-NULL-STREEBOG256   -                                                  SSLv3      Kx=GOST        Au=GOST01  Enc=None                       Mac=STREEBOG256        
```

dcooper16:
> NULL ciphers should also be covered, right?

Not entirely. If the etc/cipher-mapping.txt file is available, then ciphers that do not have an RFC name (including xFF87 - GOST2012256-NULL-STREEBOG256) are rated as neither good nor bad (5).

I believe that of the GOST ciphers:
* xFF01 is rated as tls_svrty_good, since its RFC name begins with TLS_RSA_;
* xFF00 is rated pr_svrty_high, since its RFC name ends with MD5;
* x0082 and x0083 are rated pr_svrty_critical, since their RFC names include NULL; and
* the rest as rated as neither good nor bad (5), since their RFC names don't match any of the listed cases.

We could deal with xFF87 by modifying the code to use the OpenSSL name if there is no RFC name, but that would still leave several GOST rated as neither good nor bad (5) just by default.

I just looked as `run_cipherlists()` and it includes all of the non-NULL GOST ciphers in `$average_ciphers`, which results in them being rated `pr_svrty_low` if they are present. Should we do the same in `pr_cipher_quality` (except with xFF00, which should remain as `pr_svrty_high`)?

drwetter:
Hi David,

no sure how we should treat FF01 and friends (x0081 and x0081). For FF87 .. it just seems this (and other GOST) cipher which doesn't have a RFC counterpart.

What if we add for FF87 something like ``TLS_GOST2012256-PLACEHOLDER_NULL-STREEBOG256``?

dcooper16:
> What if we add for FF87 something like TLS_GOST2012256-PLACEHOLDER_NULL-STREEBOG256?

For the moment I handled the ciphers that don't have an RFC name (FF02, FF03, FF85, FF87) in #1553 as I suggested above:

```
          ossl_cipher="$cipher"
          cipher="$(openssl2rfc "$cipher")"
          [[ -z "$cipher" ]] && cipher="$ossl_cipher"
```
With this, the case statement for RFC names catches the "NULL" in GOST2012256-NULL-STREEBOG256 and the "GOST" in GOST-GOST89MAC, GOST-GOST89STREAM, and GOST2012256-GOST89-GOST89.

However, I may never have run testssl.sh against a server that supports one of the ciphers that either doesn't have an RFC name or doesn't have an OpenSSL name. So, I can't be sure how testssl.sh handles such ciphers in general.

Running `testssl.sh --show-each -e 127.0.0.1` seems to work fine; it just leaves a blank space when the cipher does have an (RFC or OpenSSL) name. But, what about the lists of ciphers in `run_pfs()`, run_server_preference()`, `run_beast()`, etc. What if the cipher with the missing name is the "Negotiated cipher"? If one name is missing, then the other one should be used, but I can't guarantee that happens in every case without testing.

It should be possible to do some testing by modifying etc/cipher-mapping,txt to remove names for some ciphers that are commonly supported. I'll have to try that at some point when I have the time.

PR #1553, opened 2020-04-01
Handle GOST ciphers in pr_cipher_quality()

dcooper16:
This PR modifies `pr_cipher_quality()` as proposed in #1548 so that GOST ciphers are handled correctly. It changes `pr_cipher_quality()` so that the OpenSSL name is used in cases in which no RFC name is defined. It also adds a case statement for GOST so that GOST ciphers (that do not use MD5 or Null encryption) are marked as `pr_svrty_low` (as they are in `run_cipherlists`) rather than just being assigned the default rating (5).

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -5937,6 +5937,7 @@ pr_ecdh_curve_quality() {
 pr_cipher_quality() {
      local cipher="$1"
      local text="$2"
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks!

PR #1554, opened 2020-04-01
Align run_cipherlists() with pr_cipher_quality()

dcooper16:
This PR modifies `run_cipherlists()` to align with `pr_cipher_quality()`.

The biggest change made by this PR is that it breaks the current list of STRONG ciphers into two lists: one for AEAD ciphers that offer forward secrecy (STRONG) and one for AEAD ciphers that do not offer forward secrecy (GOOD).

The remaining changes are just minor tweaks:

* A few ciphers that use MD5 are moved from AVERAGE and 3DES to LOW.

* '!AECDH' was added to the OpenSSL description for LOW to catch one cipher in OpenSSL 1.0.2-chacha that offers no authentication that was being included in the LOW list.

This PR also changes sub_cipherlists() to change the output when a cipherlist with a rating of 6 is not present. There was a "FIXME" associated with this output, but it didn't matter before since there were no cipherlists with a rating of 6.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -5663,9 +5663,10 @@ sub_cipherlists() {
                               pr_svrty_good "offered (OK)"
                               fileout "$jsonID" "OK" "offered" "$cve" "$cwe"
                          else
-                              # FIXME: we penalize the absence of high but don't know the result of strong encryption yet (next)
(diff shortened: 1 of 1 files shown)


drwetter:
Thank you!

PR #1560, opened 2020-04-09
FIxed links

multiflexi:
Links in comments with http:// changed to https://. Some non working links fixed.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -13,7 +13,7 @@
 # Project lead and initiator: Dirk Wetter, copyleft: 2007-today, contributions so far see CREDITS.md
 # Main contributions from David Cooper
 #
-# License: GPLv2, see http://www.fsf.org/licensing/licenses/info/GPLv2.html
+# License: GPLv2, see https://spdx.org/licenses/GPL-2.0-only.html
 # and accompanying license "LICENSE.txt". Redistribution + modification under this
 # license permitted.
 # If you enclose this script or parts of it in your software, it has to
@@ -142,7 +142,7 @@ HNAME="${HNAME%%.*}"
 declare CMDLINE
 CMDLINE_PARSED=""                                 # This makes sure we don't let early fatal() write into files when files aren't created yet
 declare -r -a CMDLINE_ARRAY=("$@")                # When performing mass testing, the child processes need to be sent the
-declare -a MASS_TESTING_CMDLINE                   # command line in the form of an array (see #702 and http://mywiki.wooledge.org/BashFAQ/050).
+declare -a MASS_TESTING_CMDLINE                   # command line in the form of an array (see #702 and https://mywiki.wooledge.org/BashFAQ/050).
 
 
 ########### Defining (and presetting) variables which can be changed
@@ -514,7 +514,7 @@ outln() { printf -- "%b" "$1\n"; html_out "$(html_reserved "$1")\n"; }
 
 #TODO: Still no shell injection safe but if just run it from the cmd line: that's fine
 
-# Color print functions, see also http://www.tldp.org/HOWTO/Bash-Prompt-HOWTO/x329.html
+# Color print functions, see also https://www.tldp.org/HOWTO/Bash-Prompt-HOWTO/x329.html
 tm_liteblue()   { [[ "$COLOR" -ge 2 ]] && ( "$COLORBLIND" && tm_out "\033[0;32m$1" || tm_out "\033[0;34m$1" ) || tm_out "$1"; tm_off; }    # not yet used
 pr_liteblue()   { tm_liteblue "$1"; [[ "$COLOR" -ge 2 ]] && ( "$COLORBLIND" && html_out "<span style=\"color:#00cd00;\">$(html_reserved "$1")</span>" || html_out "<span style=\"color:#0000ee;\">$(html_reserved "$1")</span>" ) || html_out "$(html_reserved "$1")"; }
 tmln_liteblue() { tm_liteblue "$1"; tmln_out; }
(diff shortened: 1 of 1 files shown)


drwetter:
Hi @multiflexi,

thanks!

For the bare change of http --> https: of course.

For the ones which don't work (mozilla dxr vs. mxr): yes, thanks.

But why did you change others which work if you would have just using "https"?

testssl.sh
```diff
@@ -8723,7 +8723,7 @@ certificate_info() {
           determine_trust "$jsonID" "$json_postfix" || ((ret++))
      fi
 
-     # http://events.ccc.de/congress/2010/Fahrplan/attachments/1777_is-the-SSLiverse-a-safe-place.pdf, see page 40pp
+     # https://www.eff.org/files/ccc2010.pdf, see page 40pp
```
drwetter, line 8726: https://events.ccc.de/congress/2010/Fahrplan/attachments/1777_is-the-SSLiverse-a-safe-place.pdf works too

testssl.sh
```diff
@@ -8352,10 +8352,10 @@ certificate_info() {
                                         let ret++ ;;
           esac
           out "$short_keyAlgo "
-          # https://tools.ietf.org/html/rfc4492,  http://www.keylength.com/en/compare/
-          # http://infoscience.epfl.ch/record/164526/files/NPDF-22.pdf
-          # see http://csrc.nist.gov/publications/nistpubs/800-57/sp800-57_part1_rev3_general.pdf
-          # Table 2 @ chapter 5.6.1 (~ p64)
+          # https://tools.ietf.org/html/rfc4492,  https://www.keylength.com/en/compare/
+          # https://doi.org/10.1007/s00145-001-0009-4
+          # see http://dx.doi.org/10.6028/NIST.SP.800-57pt1r4
```
drwetter, line 8357: https://csrc.nist.gov/publications/detail/sp/800-57-part-1/rev-4/final

testssl.sh
```diff
@@ -8352,10 +8352,10 @@ certificate_info() {
                                         let ret++ ;;
           esac
           out "$short_keyAlgo "
-          # https://tools.ietf.org/html/rfc4492,  http://www.keylength.com/en/compare/
-          # http://infoscience.epfl.ch/record/164526/files/NPDF-22.pdf
-          # see http://csrc.nist.gov/publications/nistpubs/800-57/sp800-57_part1_rev3_general.pdf
-          # Table 2 @ chapter 5.6.1 (~ p64)
+          # https://tools.ietf.org/html/rfc4492,  https://www.keylength.com/en/compare/
+          # https://doi.org/10.1007/s00145-001-0009-4
```
drwetter, line 8356: https://infoscience.epfl.ch/record/164526/files/NPDF-22.pdf

testssl.sh
```diff
@@ -13,7 +13,7 @@
 # Project lead and initiator: Dirk Wetter, copyleft: 2007-today, contributions so far see CREDITS.md
 # Main contributions from David Cooper
 #
-# License: GPLv2, see http://www.fsf.org/licensing/licenses/info/GPLv2.html
+# License: GPLv2, see https://spdx.org/licenses/GPL-2.0-only.html
```
drwetter, line 16: maybe that would make sense -- can't tell whether I want the LF here -- but https://www.gnu.org/licenses/old-licenses/gpl-2.0.html would work too.

testssl.sh
```diff
@@ -10484,7 +10484,7 @@ code2network() {
      NW_STR=$(sed -e 's/,/\\\x/g' <<< "$1" | sed -e 's/# .*$//g' -e 's/ //g' -e '/^$/d' | tr -d '\n' | tr -d '\t')
 }
 
-# sockets inspired by http://blog.chris007.de/?p=238
+# sockets inspired by https://blog.chris007.de/using-bash-for-network-socket-operation/
```
drwetter, line 10487: good catch!

testssl.sh
```diff
@@ -1334,7 +1334,7 @@ html_header() {
           html_out "<?xml version=\"1.0\" encoding=\"UTF-8\" ?>\n"
           html_out "<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Strict//EN\" \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd\">\n"
           html_out "<!-- This file was created with testssl.sh. https://testssl.sh -->\n"
-          html_out "<html xmlns=\"http://www.w3.org/1999/xhtml\">\n"
+          html_out "<html xmlns=\"http://www.w3.org/1999/xhtml/\">\n"
```
drwetter, line 1337: Hmm... for some reason the w3c even lists without trailing slash: https://www.w3.org/TR/1999/WD-xhtml1-19991124/ .  Not probably that it would matter much but how come you stumbled over this (reference)?

testssl.sh
multiflexi: Works for me now, did not before.

testssl.sh
multiflexi: I thought DOI will be better because it shows the current version of the document so it is not necessary to keep track of changing URL.

testssl.sh
multiflexi: Again I perceive DOI as more stable URL. Just an idea.

testssl.sh
multiflexi: I have seen that this is used currently in many FOSS projects for license reference. I guess either will work.

testssl.sh
multiflexi: I do not remember messing with this, just a mistake probably. Sorry.

testssl.sh
```diff
@@ -18597,7 +18600,7 @@ mybanner() {
      bb1=$(cat <<EOF
 
 ###########################################################
-    $PROG_NAME       $VERSION from 
+    $PROG_NAME       $VERSION from
```
drwetter, line 18603: This blank is needed

testssl.sh
```diff
@@ -18606,7 +18609,7 @@ EOF
              modification under GPLv2 permitted.
       USAGE w/o ANY WARRANTY. USE IT AT YOUR OWN RISK!
 
-       Please file bugs @ 
+       Please file bugs @
```
drwetter, line 18612: This one too !

testssl.sh
multiflexi: I did not delete those, Atom just removes stuff I guess.

testssl.sh
multiflexi: Same

drwetter:
Ok, thanks for the feedback. Then let me work on finishing this during the next days.

testssl.sh
drwetter: You're not the first. :-)  I think I need a workaround for atom and vim users who automatically squash those blanks.

Issue #100, opened 2015-05-09
Rating


drwetter:
can be done with ~~return values of functions~~, global vars

drwetter:
c&p'ed from #31:

For the rating I guess the best is a global variable. It starts with hundred (equals 100%) and then each step comes a penalty. So far so good. The difficult part is a fair rating. If you look at SSLlabs: they have a 10 page PDF for this. So one needs to put "some" thoughts into it. And: it's changing over time as new bugs hit the streets or it becomes public e.g. that EC curve ABC is compromised or whatever.

ajbouh:
@drwetter, 

Have you seen sslrate over at https://github.com/einaros/sslrate/blob/master/sslrate.py? They claim to have implemented something like SSLlabs ratings for sslyze xml output.

Beyond just a rating, they include a check against known bad certificates in openssl-blacklist. I did the (admittedly minimal) work to unpack openssl-blacklist to be usable without debian/ubuntu package tools. If it'd be useful, I can contribute that unpacking logic back to this project somehow.

Ideally I'd like to have a strict set of requirements that *must* be met, as a part of our in-house continuous integration tests. Most of testssl.sh's tests are perfect for this, except for the OCSP ones, as they appear to assume that an external/third party server would be accessible during testing.

drwetter:
Hi @ajbouh

Am 07/22/2015 um 08:59 PM schrieb ajbouh:
> @drwetter <https://github.com/drwetter>,
> 
> Have you seen sslrate over at https://github.com/einaros/sslrate/blob/master/sslrate.py? They
> claim to have implemented something like SSLlabs ratings for sslyze xml output.

ok, thx. Reads interesting.

> Beyond just a rating, they include a check against known bad certificates in openssl-blacklist.
> I did the (admittedly minimal) work to unpack openssl-blacklist to be usable without
> debian/ubuntu package tools. If it'd be useful, I can contribute that unpacking logic back to
> this project somehow.

Thx for the reminder. Comment:

yeah, well. It's certainly a "nice to have", so I opened an issue (#146).
But as that was ages ago and I think the likeliness of still detecting broken
keys/certificates is incredibly small, it's just a matter of completing this project.

If you feel like doing this in bash, pls go ahead.

> Ideally I'd like to have a strict set of requirements that /must/ be met, as a part of our
> in-house continuous integration tests. Most of testssl.sh's tests are perfect for this, except
> for the OCSP ones, as they appear to assume that an external/third party server would be
[...]

PR #1569, opened 2020-04-15
Added grading based on ssllabs

magnuslarsen:
Brings a score + grade, based on the [open document from SSLLabs](https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide).
Checks that aren't implemented yet, is documented in the `testssl.1.md` file, under `GRADING`.

Some of those looks quite easy to implement, and with a some help, could be implemented.
I would really like help with implementing [DH param](https://www.openssl.org/docs/man1.0.2/man1/dhparam.html) bit length, which is the proper method for looking at "Key Exchange"-scores.
It does also expose more information in general for TestSSL. That should be its own issue/PR?

It currently looks something like this:
image
It can also output basic json/csv (just grade and specification).

This can always be ported to an "include file" in the future, if another grading spec or need, arises. But for now, it's a simple solution to scoring/grading, which is very similar to what Qualys brings.

It has been tested against various local sites, and relevant sites from [badssl.com](https://badssl.com/), and brings the same score as Qualys for >90% of the time (based on my small testing group (~10 local sites).

The only discrepancies I've seen, is when TestSSL detects something that Qualys don't, and vice versa (Forward Secrecy being prominent).
Sometimes Qualys scores "Key Exchange" weird (70), which is not a possible answer (as seen in Table 4), which potentially can lead to a different score too:
image

All kinds of feedback welcome :-)

### DIFF
--- a/doc/testssl.1.md
+++ b/doc/testssl.1.md
@@ -56,6 +56,8 @@ linked OpenSSL binaries for major operating systems are supplied in `./bin/`.
 
 9) client simulation
 
+10) Result of script in form of a grade
+
 
 ## OPTIONS AND PARAMETERS
 
@@ -142,8 +144,7 @@ in `/etc/hosts`.  The use of the switch is only useful if you either can't or ar
 
 `--phone-out` Checking for revoked certificates via CRL and OCSP is not done per default. This switch instructs testssl.sh to query external -- in a sense of the current run -- URIs. By using this switch you acknowledge that the check might have privacy issues, a download of several megabytes (CRL file) may happen and there may be network connectivity problems while contacting the endpoint which testssl.sh doesn't handle. PHONE_OUT is the environment variable for this which needs to be set to true if you want this.
 
-`--add-ca <cafile>` enables you to add your own CA(s) for trust chain checks. `cafile` can be a single path or multiple paths as a comma separated list of root CA files. Internally they will be added during runtime to all CA stores. This is (only) useful for internal hosts whose certificates is issued by internal CAs. Alternatively 
-ADDITIONAL_CA_FILES is the environment variable for this.
+`--add-ca <cafile>` enables you to add your own CA(s) for trust chain checks. `cafile` can be a single path or multiple paths as a comma separated list of root CA files. Internally they will be added during runtime to all CA stores. This is (only) useful for internal hosts whose certificates is issued by internal CAs. Alternatively ADDITIONAL_CA_FILES is the environment variable for this.
 
 
 ### SINGLE CHECK OPTIONS
@@ -286,6 +287,8 @@ Please note that in testssl.sh 3,0 you can still use `rfc` instead of `iana` and
 5. display bytes received via sockets
 6. whole 9 yards
 
+`--disable-grading` disables grading explicitly.
+Grading automatically gets disabled, to not give a wrong or misleading grade, when not all required functions are executed (e.g when checking for a single vulnerabilities). `DISABLE_GRADING` is the according environment variable which you can use.
 
 
 ### FILE OUTPUT OPTIONS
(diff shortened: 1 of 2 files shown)


drwetter:
Hi @magnuslarsen,

pleasant surprise, thank you! 

It's a great idea. It puts some burden on maintaining the rating though and I expect discussions like "why is this penalized" and "why is that" etc.. I probably don't have much time for that if it'll have a frequency like on SSLlabs. So using SSLlabs rating would shift the the burden a bit. But I am not sure whether I would like to follow SSLlabs rating for 100%. :=) Don't know which pill I want to take, red or blue here. So a pluggable include file would be the best for medium terms, i.e. sometime after this is merged into 3.1dev.

I read quickly through it. Code looks good for a first contribution! I'll merge it first to temporary branch to polish it a bit. If you want then to amend your work, pls PR against that branch.

Thanks & cheers, Dirk


See also #100.

magnuslarsen:
I agree with you. I don't think everyone will like a particular grading system, regardless of which is used, and how.

I don't particularly like grading in general, since most people begin to look into grading more so than the actual security.
But I need a way to convey a "state of the situation" to my higher-ups, where grading/scoring is so helpful. "Two major websites are now an A grade, from last weeks F grades" is something anyone can understand :-)

I'll stay around, and see if I can chime in somehow.

Thank you!

testssl.sh
```diff
@@ -2705,11 +2822,13 @@ run_hpkp() {
                "$has_backup_spki" && out "$spaces"       # we had a few lines with backup SPKIs already
                prln_svrty_high " No matching key for SPKI found "
                fileout "HPKP_SPKImatch" "HIGH" "None of the SPKI match your host certificate, intermediate CA or known root CAs. Bricked site?"
+               set_grade_cap "A" "Problems with HTTP Public Key Pinning (HPKP)"
```
drwetter, line 2825: For a bricked site an A?

testssl.sh
```diff
@@ -5042,6 +5167,7 @@ run_protocols() {
                     latest_supported_string="SSLv3"
                fi
                add_tls_offered ssl3 yes
+               set_grade_cap "B" "SSLv3 is offered"
```
drwetter, line 5170: Just a comment: I thought this would be an error. But it isn't (https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide#protocol-support)

drwetter: Likely we could adjust that -- given the B rating TLS 1.0/1.1 protocol offering has

doc/testssl.1.md
```diff
@@ -383,13 +386,56 @@ Except the environment variables mentioned above which can replace command line
...
 
+### GRADING
+This script has a near-complete implementation of SSLLabs's '[SSL Server Rating Guide](https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide)'.
+
+This is *not* a reimplementation of the [SSLLab's SSL Server Test](https://www.ssllabs.com/ssltest/analyze.html), but a implementation of the above grading specification, slight discrepancies might occur!
+
+Disclaimer: Having a good grade does **NOT** necessary equal to having good security! Never rely solely on a good grade!
+
+As of writing, these checks are missing:
+* Authenticated encryption (AEAD) - should be graded **B** if not supported
```
drwetter, line 397: That could also be retrieved, right?

doc/testssl.1.md
```diff
@@ -383,13 +386,56 @@ Except the environment variables mentioned above which can replace command line
...
+The `set_grading_state()` automatically disables grading, if all the required checks are *not* enabled.
+This is to prevent giving out a misleading or wrong grade.
+
+#### Implementing a new revision
+When a new revision of the grading specification comes around, the following has to be done:
+* New grade caps has to be either:
+  1. Added to the script wherever relevant, or
+  2. Added to the above list of missing checks (if *i.* is not possible)
+* New grade warnings has to be added wherever relevant
+* The revision output in `run_grading()` function has to updated
```
drwetter, line 432: Documentation here is really helpful! :-) Thank you!

doc/testssl.1.md
magnuslarsen: I think it could yes. I couldn't really figure out where it would be sensible, without introducing big changes in the code.
You could get all ciphers that support AEAD, and see if any of them is offered. But I feel like there is a better way

testssl.sh
magnuslarsen: It's an old rule yes: https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide#changes-in-2009h-30-october-2014

I also think we could adjust it. B is way too high IMO

testssl.sh
magnuslarsen: According to the rules, yes: https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide#changes-in-2009l-4-march-2016

Again, i think we can adjust accordingly (F).

drwetter:
https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide#protocol-support doesn't mention TLS 1.3 :-(   Following the discussion on github, I sensed that different though. The typical complaint was a 128Bit TLS 1.3 cipher in RFC leads to downgrade

drwetter:
> https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide#protocol-support doesn't mention TLS 1.3 :-( Following the discussion on github, I sensed that different though. The typical complaint was a 128Bit TLS 1.3 cipher in RFC leads to downgrade

Just ran it against a TLS 1.3-only host:

``Grade cap reasons             Grade capped to C. TLS1.2 is not offered``
``Grade capped to A. Does not support TLS_FALLBACK_SCSV``

testssl.sh
drwetter: The problem I see: if we touch the rating it's hard to finger point at SSLlabs

testssl.sh
```diff
@@ -15860,6 +16022,7 @@ run_sweet32() {
                sclient_connect_successful $? $TMPFILE
                sclient_success=$?
                [[ $DEBUG -ge 2 ]] && grep -Eq "error|failure" $ERRFILE | grep -Eav "unable to get local|verify error"
+               [[ $proto == -tls1_1 && $sclient_success -eq 0 ]] && tls1_1_vulnable=true
```
drwetter, line 16025: Why did you check against the protocol here?

magnuslarsen:
https://discussions.qualys.com/thread/20751-servers-that-only-support-tls-13-shouldnt-be-downgraded
^ TLS1.3 is being finalized.

I did see a comment, which I can't find anymore, arguing that TLS1.2 properly configured is just as secure as TLS1.3, and has support for older client, which leads to a better reachability; so having both on would lead to a "better" website.
But I am in for waiting till they finalize TLS1.3, and properly score it here.

testssl.sh
```diff
@@ -15879,9 +16042,11 @@ run_sweet32() {
      if [[ $sclient_success -eq 0 ]] && "$ssl2_sweet" ; then
           pr_svrty_low "VULNERABLE"; out ", uses 64 bit block ciphers for SSLv2 and above"
           fileout "SWEET32" "LOW" "uses 64 bit block ciphers for SSLv2 and above" "$cve" "$cwe" "$hint"
+          "$tls1_1_vulnable" && set_grade_cap "C" "Uses 64 bit block ciphers with TLS1.1+ (vulnerable to SWEET32)"
      elif [[ $sclient_success -eq 0 ]]; then
           pr_svrty_low "VULNERABLE"; out ", uses 64 bit block ciphers"
           fileout "SWEET32" "LOW" "uses 64 bit block ciphers" "$cve" "$cwe" "$hint"
+          "$tls1_1_vulnable" && set_grade_cap "C" "Uses 64 bit block ciphers with TLS1.1+ (vulnerable to SWEET32)"
```
drwetter, line 16049: Shouldn't this be just TLS1.1?  ``$tls1_1_vulnable` is only set to true if TLS 1.1 has a 64 bit cipher. TLS 1.2 shouldn't have any 64 bit cipher.

Second remark: why do you limit this to TLS 1.1 in the first place?

drwetter (commented):
I am just in the process going roughly through you commit and push the result into a new branch. There we can work on further details.

testssl.sh
magnuslarsen: Second point in this: https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide#changes-in-2009n-19-january-2017---3-april-2017

Should be only `TLS1.1` yes, and not `TLS1.1+` (since TLS1.2 doesn't have)

drwetter: 2nd point: ok, read that. Thanks. However it's hard to believe they don't penalize it for earlier versions. This needs to be checked maybe with a real life example. I add that as an open point

drwetter:
> It does also expose more information in general for TestSSL. That should be its own issue/PR?

what do you mean by that?

drwetter:
See #1573. I'll close this but if you want to comment on _old_ points which we started here you can do that here as well.

Thanks

PR #1487, opened 2020-01-30
Remove (harmless) AESGCM dup in the strong ciphers list

simondeziel:
Signed-off-by: Simon Deziel <simon@sdeziel.info>

PR #1572, opened 2020-04-16
Fix typo in Strong grade Ciphers AEAD docs

rhardy613:
Thanks for an awesome tool.
<li><code>Strong grade Ciphers</code> (AEAD): 'AESGCM:CHACHA20:AESGCM:CamelliaGCM:AESCCM8:AESCCM'</li>
The Strong AEAD ciphers on line 243 above don't make sense. I suspect a copy paste bug. The 2nd AESGCM should be ARIAGCM no?
Sorry I haven't checked the code to see if the same issue is in the code or if this was just a documentation glitch.

### DIFF
--- a/doc/testssl.1.html
+++ b/doc/testssl.1.html
@@ -240,7 +240,7 @@ ADDITIONAL_CA_FILES is the environment variable for this.</p>
 <li><code>LOW</code> (64 Bit + DES ciphers, without EXPORT ciphers): 'LOW:DES:RC2:RC4:!ADH:!EXP:!NULL:!eNULL'</li>
 <li><code>3DES + IDEA Ciphers</code>: '3DES:IDEA:!aNULL:!ADH'</li>
 <li><code>Average grade Ciphers</code>: 'HIGH:MEDIUM:AES:CAMELLIA:ARIA:!IDEA:!CHACHA20:!3DES:!RC2:!RC4:!AESCCM8:!AESCCM:!AESGCM:!ARIAGCM:!aNULL'</li>
-<li><code>Strong grade Ciphers</code> (AEAD): 'AESGCM:CHACHA20:AESGCM:CamelliaGCM:AESCCM8:AESCCM'</li>
+<li><code>Strong grade Ciphers</code> (AEAD): 'AESGCM:CHACHA20:ARIAGCM:CamelliaGCM:AESCCM8:AESCCM'</li>
 </ul>
 
 


drwetter:
Hi @rhardy,

thanks!

You're right. That was a c&p error from a former version of the code (https://github.com/drwetter/testssl.sh/pull/1487)

However the html file is (normally) being generated from the md file. Would you mind amending your PR with a patched md and roff (~/doc/testssl.1) file?

dcooper16:
> You're right. That was a c&p error from a former version of the code (#1487)
> 
> However the html file is (normally) being generated from the md file. Would you mind amending your PR with a patched md and roff (~/doc/testssl.1) file?

This is also a mistake of mine. This information was changed by #1554, but I didn't change the documentation, since I didn't know that the documentation included this level of detail about these tests. The lists that are currently being tested in the 3.1dev branch are:
```
NULL ciphers (no encryption): 'NULL:eNULL'
Anonymous NULL Ciphers (no authentication): 'aNULL:ADH'
Export ciphers (w/o ADH+NULL): 'EXPORT:!ADH:!NULL'
LOW: 64 Bit + DES, RC[2,4], MD5 (w/o export): 'LOW:DES:RC2:RC4:MD5:!ADH:!EXP:!NULL:!eNULL:!AECDH'
Triple DES Ciphers / IDEA: '3DES:IDEA:!aNULL:!ADH:!MD5'
Obsolete: SEED + 128+256 Bit CBC cipher: 'HIGH:MEDIUM:AES:CAMELLIA:ARIA:!IDEA:!CHACHA20:!3DES:!RC2:!RC4:!AESCCM8:!AESCCM:!AESGCM:!ARIAGCM:!aNULL:!MD5'
non-FS Strong encryption (AEAD ciphers): 'AESGCM:CHACHA20:CamelliaGCM:AESCCM:ARIAGCM:!kEECDH:!kEDH:!kDHE:!kDHEPSK:!kECDHEPSK:!aNULL'
Forward Secure Strong encryption (AEAD ciphers): 'AESGCM:CHACHA20:CamelliaGCM:AESCCM:ARIAGCM:!kPSK:!kRSAPSK:!kRSA:!kDH:!kECDH:!aNULL'
```

drwetter:
> This is also a mistake of mine

David, you're doing so much. The word mistake is feeling a bit exaggerated for this. :-)

rhardy613:
Not sure I speak md and roff but I can take a whirl this weekend.
Much of my testing with this tool was trying to find a strong cipher string for Tomcat 9 which would pass also an SSL Labs scan. I finally ended up finding that but it didn't follow your strong string.
My FS strong ciphers string for Tomcat 9 that passed all the scans was:
"CHACHA20:AESGCM:ARIAGCM:CamelliaGCM:AESCCM8:AESCCM:-PSK:-DSS:-aNULL:-RSA"

rhardy613:
It's interesting that when I test a server with SSL Labs your FS strong ciphers string it results in the exact same ciphers in Tomcat 9 as mine but when I expand both strings with openssl your string gives several additional ciphers which I assume my Tomcat 9 build doesn't implement or allow. It would be nice to get CHACHA20 with TLS 1.3 but Tomcat 9.0.16 seems to ignore cipher order for TLS 1.3. It always wants to use TLS_AES_256_GCM_SHA384 with TLS 1.3.

drwetter:
> Not sure I speak md and roff but I can take a whirl this weekend

It's always good to know in life what one can do and what not. :-) But I am sure you don't need extended knowledge here. They are just text files with a similar logic. Only the M in HTML is different and it's not hypertext.

\<OT\> 

> It's interesting that when I test a server with SSL Labs your FS strong ciphers string it results in the exact same ciphers in Tomcat 9 as mine but when I expand both strings with openssl your string gives several additional ciphers which I assume my Tomcat 9 build doesn't implement or allow

It depends on openssl (version used, config), Java and as you figured Tomcat. 

> but Tomcat 9.0.16 seems to ignore cipher order for TLS 1.3.

Don't know about Tomcat but with Apache or Nginx it depends on openssl IIRC. Google is your friend. Honestly I wouldn't not care that much. I would rather not use Tomcat without an Apache or Nginx in front of it as they easier to configure and provide better means of control.

\<\/OT\>

rhardy613:
Tomcat 9 can either use openssl or the JSSE implementation. I'm forcing openssl use. I would normally like a full web server as a bastion host in front but that is incompatible with other security controls in this case.

drwetter:
Hi @rhardy,

thanks for raising your hand! 

To get this off the table I patched that myself in both branches.

Cheers, Dirk

Issue #1502, opened 2020-02-12
Feature request: --ignore-tests <test id>


GottemHams:
Perhaps I'm overlooking something but I couldn't find anything in the manual nor the issues on here. Currently there doesn't seem to be a way to exclude a specific test when running with the default options, e.g. if you're aware of a certain risk so you want to ignore warnings about it.

__Example__
When you're using the default options it also runs the BREACH test. Now, you could just specify all other options yourself and simply exclude -B, but as tests get added/changed you'd continually have to keep the flags up to date. In the specific example of BREACH: I'm aware of the possible risk but personally I can live with this.

__Proposal__
Add an argument like `--ignore-tests <test id,test id 2,...>`, where the `test id` can be stuff like `BREACH,xxx,yyy,zzz`. These flags would correspond to the test IDs as they're shown in the output JSON: `"id": "BREACH"`. I'm assuming these don't usually change.

drwetter:
> Perhaps I'm overlooking something 

Nope ;-) 

In general you can always pick the tests you like with the all switches and just omit what you don't like. That might be not as handy though as the suggestion above.

I was thinking about skipping tests too but what I had in mind then was just to use e.g. ``+B`` instead of ``-B`` - albeit the logic of ``+`` seems a bit upside down. Of course also your suggestion would work. (anybody else has a good proposal?)

Thx, Dirk

GottemHams:
I only started using testssl a couple days ago so I'm not 100% familiar with all arguments yet. But I assume the following is what makes it test for "outdated" TLS versions (1.0, 1.1):
`-p, --protocols               checks TLS/SSL protocols (including SPDY/HTTP2)`

Now, while your suggestion of using `+B` to omit the BREACH test would also work, you can't skip __just__ the test for TLS 1.0. Like in case you __have__ to use 1.1 to support some older clients, but you don't want or need 1.0. However, by using exclusions based on test ID, you __can__ do that. Just exclude `TLS1` and you should be good, since `TLS1_1` is the ID for 1.1 and I'm assuming there are also `TLS1_2` and `TLS1_3`. The same would likely hold true for other "grouped" tests.

So `--ignore-tests` would be much more flexible and would have my preference.

drwetter:
> .. you can't skip just the test for TLS 1.0. Like in case you have to use 1.1 to support some older clients, but you don't want or need 1.0. 

[...]

PR #1573, opened 2020-04-17
WIP: Add grading based on ssllabs

drwetter:
Merged and polished PR from @magnuslarsen, see  #1569

Brings a score + grade, based on the [open document from SSLLabs](https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide).
Checks that aren't implemented yet, is documented in the `testssl.1.md` file, under `GRADING`.


drwetter:
Copy and paste the todos from the commits:

* Review whether it is rated as intended ==> more testing
* Do we want to modify SSL Lab's rating? (SSLv3 e.g., T for SHA1 certificate?)
* <s>Does JSON output work?</s>: yes
* <s>TLS 1.3 only server are not rated properly --> wait for SSLlabs?</s>
* SWEET32: rating refers to TLS 1.1 atm. SSLlabs docu doesn't give a hint
  (is their docu incomplete?)
* <s>Rating for STARTTLS at all?</s> see 127cf95e2266384c505d541a79b727b701f160f6 . But documentation needes to be amended for this
* Implementing DH param bit length???

magnuslarsen:
>> [in regards to DH_PARAMs] It does also expose more information in general for TestSSL. That should be its own issue/PR?

> what do you mean by that?

I just mean that we can output more information (looking besides grading), what I'm thinking:
```
Server DH Params: 4096 bits (OK)
Server DH Params: 2048 bits (OK)
Server DH Params: 0 bits, no DH key exchange (OK)
Server DH Params: 0 bits, using DH key exchange (not OK)
...
```

drwetter:
> I just mean that we can output more information (looking besides grading), what I'm thinking

ah, ok. 

> That should be its own issue/PR?

Yes. In principle we have that, e.g. try ``./testssl.sh --nsa testssl.sh``

magnuslarsen:
Hmm, couldn't we use `$DH_GROUP_LEN_P` for EC keys, since the DH params (or group) is what is taken into account, when rating key exchange score (and not if the actual group is known or not):
https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide#key-exchange
(especially the note below the table)
```
if [[ $cert_key_algo =~ ecdsa ]] || [[ $cert_key_algo =~ ecPublicKey ]]; then
...
fi

set_key_str_score "$short_keyAlgo" "$DH_GROUP_LEN_P"
```

Niko78:
hi
just tried grading
it seems there is an error about TLS_FALLBACK_SCSV
I got that on my servers :
Grade capped to A. Does not support TLS_FALLBACK_SCSV
while I have  TLS_FALLBACK_SCSV (RFC 7507)              No fallback possible (OK), no protocol below TLS 1.2 offered
in ssllabs I get well A+

Nicolas

magnuslarsen:
@Niko78 , I've been too aggressive in that grading cap. I tihnk we should leave out any "not possible" checks. "Not possible" (due to no available protocols) != not supported. The only check we should leave in, is this:
```
if [[ -z "$POODLE" ]]; then
    pr_warning "Rerun including POODLE SSL check. "
    pr_svrty_medium "Downgrade attack prevention NOT supported"
    fileout "$jsonID" "WARN" "NOT supported. Pls rerun with POODLE SSL check"
    ret=1
elif [[ "$POODLE" -eq 0 ]]; then
    pr_svrty_high "Downgrade attack prevention NOT supported and vulnerable to POODLE SSL"
    fileout "$jsonID" "HIGH" "NOT supported and vulnerable to POODLE SSL"
    set_grade_cap "C" "Vulnerable to POODLE"
else
    pr_svrty_medium "Downgrade attack prevention NOT supported"
    fileout "$jsonID" "MEDIUM" "NOT supported"
fi
set_grade_cap "A" "Does not support TLS_FALLBACK_SCSV"
```

drwetter:
> in ssllabs I get well A+

Thx but with this bare info we cannot reproduce it, @Niko78

Niko78:
> 
> 
> > in ssllabs I get well A+
> 
> Thx but with this bare info we cannot reproduce it, @Niko78

@drwetter I sent you 2 links to check ;-)

dcooper16:
> > I just mean that we can output more information (looking besides grading), what I'm thinking
> 
> ah, ok.
> 
> > That should be its own issue/PR?
> 
> Yes. In principle we have that, e.g. try `./testssl.sh --nsa testssl.sh`

For finite-field DH, one should also look at LOGJAM: `./testssl.sh --logjam --nsa testssl.sh`

`run_fs()` (`--nsa`) only checks for support for strong forward-secure cipher suites, so it will not catch if ephemeral DH keys are being used with a weak cipher suite. `run_logjam()` (`--logjam`) checks all non-TLSv1.3 cipher suites that use ephemeral DH keys.

`run_logjam()` won't catch if ephemeral DH keys are being used with TLSv1.3. However, only the groups from RFC 7919 can be used with TLSv1.3, and these are all strong groups with primes that are at least 2048 bits.

dcooper16:
> Hmm, couldn't we use `$DH_GROUP_LEN_P` for EC keys, since the DH params (or group) is what is taken into account, when rating key exchange score (and not if the actual group is known or not):
> https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide#key-exchange
> (especially the note below the table)
> 
> ```
> if [[ $cert_key_algo =~ ecdsa ]] || [[ $cert_key_algo =~ ecPublicKey ]]; then
> ...
> fi
> 
> set_key_str_score "$short_keyAlgo" "$DH_GROUP_LEN_P"
> ```

This is mixing two different things. `$short_keyAlgo`, which is defined in `certificate_info()` indicates the type of public key in the server's certificate. `$DH_GROUP_LEN_P` is about the length of a DH key used in an ephemeral key exchange.

So, `$short_keyAlgo` were 'DH', this certificate could be used with cipher suites such as TLS_DH_DSS_WITH_AES_256_GCM_SHA384 or TLS_DH_RSA_WITH_AES_256_CBC_SHA256. The "_DH_" indicates that a static DH key is used, and that the certificate that contains the server's DH key is signed with DSS or RSA, as indicated by what follows "_DH_" in the name.

`$DH_GROUP_LEN_P` would be populated if the server supported a cipher suite such as  TLS_DHE_RSA_WITH_AES_256_GCM_SHA384. This cipher suite indicates that key exchange is performed using ephemeral DH, and that the server's public key (`$short_keyAlgo`) is an RSA key.

`certificate_info()` only looks at the contents of the server's certificate()s. As noted in the previous comment, `run_fs()` and `run_logjam()` are the functions that collect information about any ephemeral DH or ECDH keys the server uses for key exchange.

Niko78:
I ran some tests again with internal servers and I have a question about results like that :
Protocol Support (weighted)  95 (28)
 Key Exchange     (weighted)  90 (27)
 Cipher Strength  (weighted)  90 (36)
 Final Score                  91
 Grade                        B
 Grade cap reasons            Grade capped to B. Uses weak DH key exchange parameters (vulnerable to LOGJAM)
                              Grade capped to B. TLS 1.1 offered
                              Grade capped to B. TLS 1.0 offered
                              Grade capped to B. RC4 ciphers offered
                              Grade capped to A. HSTS is not offered

first : HSTS not offered = A ?
second : this server has 
Negotiated cipher            ECDHE-RSA-RC4-SHA, 521 bit ECDH (P-521) -- inconclusive test, matching cipher in list missing, better see below
 RC4 (CVE-2013-2566, CVE-2015-2808)        VULNERABLE (NOT ok): ECDHE-RSA-RC4-SHA RC4-SHA RC4-MD5

so are you sure "B" is right ?

magnuslarsen:
Regarding the latest commit (127cf95e2266384c505d541a79b727b701f160f6) and JSON/CSV output, I think we should be wary of including reason in the same output as the grade.
JSON, at least, is primarily meant for machine readable output. When expecting just a grade, and suddenly having more than just the grade; it leads to hassle coding around it.

If it should be included, I'd suggest having in a different field `grade_cap_reasons`, and populating that with reasons (to keep consistent)

drwetter:
> field grade_cap_reasons, and populating that with reasons (to keep consistent)

yes, makes sense. That is difficult if there's more than one reason.

drwetter:
BTW: While for each test / group of category "grade" sounds ok, otherwise I would rather go with rating. SSL Labs does that similar. 

IANANS though (NS=native speaker)

magnuslarsen:
> BTW: While for each test / group of category "grade" sounds ok, otherwise I would rather go with rating. SSL Labs does that similar.

As long as it is consistent, it shouldn't matter. I'm fine with either.

> That is difficult if there's more than one reason.

Couldn't it be done by just joining them with something like a semicomma?
```json
{ "finding": "Grade capped to B. TLS1.1 offered; Grade capped to B. TLS1.0 offered" }
```

By the way, I found a annoying bug. `--disable-grading` no longer works. This is because `$do_grading` is set to false (when calling `--disable-grading`), BEFORE `count_do_variables()` is called (in `parse_cmd_line()`).
`count_do_variables()` checks for variables that are `true` - which `$do_grading` isn't, so `set_scanning_defaults()` is called, which sets it to true again :/

drwetter:
> By the way, I found a annoying bug. --disable-grading no longer works

Actually I was thinking removing that, latest when #1502 is in place

drwetter:
> By the way, I found a annoying bug

Sigh. "Disabling" is a different logic which doesn't fit yet. Probably time either to remove it or to prepare for #1502

drwetter:
The grade_cap_reasons were now committed. Actually I liked the idea from you.

> Couldn't it be done by just joining them with something like a semicomma?

Or just multiple grade_cap_reasons. E.g cloudflare.com lists multiple certificates, so  ``testssl.sh -S --json cloudflare.com``  has multiple entries like _cert_keySize <cert#1>_, _cert_keyUsage <cert#1>_ and _cert_keySize <cert#2>_, _cert_keyUsage <cert#2>_

drwetter:
> ``run_fs()`` [..]  will not catch if ephemeral DH keys are being used with a weak cipher suite. ``run_logjam()`` [..] checks all non-TLSv1.3 cipher suites that use ephemeral DH keys

Currently scratching my head whether this can be done better, don't know whether we discussed that before. Either combine the results of the two functions and store the results. Or, off the top of my head and not looking at any details,  just one function which determines all keys and stores them for ``run_logjam()`` and ``run_fs()``.

dcooper16:
>   >  `run_fs()` [..] will not catch if ephemeral DH keys are being used with a weak cipher suite. `run_logjam()` [..] checks all non-TLSv1.3 cipher suites that use ephemeral DH keys
> 
> Currently scratching my head whether this can be done better, don't know whether we discussed that before. Either combine the results of the two functions and store the results. Or, off the top of my head and not looking at any details, just one function which determines all keys and stores them for run_logjam() and run_fs().

At the moment `run_fs()` does set `$DH_GROUP_OFFERED` and `$DH_GROUP_LEN_P` when possible. If they are set, then `run_logam()` uses the contents of these variables (although it still tests whether the server supports any DH export ciphers). `run_fs()` will only set `$DH_GROUP_OFFERED` and `$DH_GROUP_LEN_P` if it finds something using TLS 1.2 or below. So, it most likely will not set these variables if the server supports TLS 1.3. Also, at noted before, it won't set these variables if the server only supports weak DH ciphers.

drwetter:
> By the way, I found a annoying bug

this is fixed now

magnuslarsen:
In commit 32eab3ead9c988e4574776987b09cd0df27176ed, I think the removal of `do_rating=true` in `set_scanning_defaults()` was unintended? It at least borked the rating, and kept it disabled at all times.
In my simple testing, by just adding `do_rating=true` in again, made it work, and `--disable-rating` works again (by disabling the output - nice!)

drwetter:
> In commit 32eab3e, I think the removal of do_rating=true in set_scanning_defaults() was unintended?

No, that was intended but the effect was expected to be not that rating is completely disabled :-)  

Fixed now.

drwetter:
> Or just multiple grade_cap_reasons

done

drwetter:
I'll merge this as is during the next days. Key sizes can be added later. Also I probably leave the rating for now. We'll see how that goes.

@magnuslarsen : It would be great if you can continue listening for any related rating issues. And if maybe you can keep an eye on any new release on SSL Labs rating template.


@magnuslarsen :

magnuslarsen:
I'll stay around, see where I can help! Thanks for the merge :heart:

drwetter:
FYI: rating is now in 3.1dev.

drwetter:
Comments crossed ;-) Well, thank you!

PR #1580, opened 2020-04-22
WIP: Extended run_server_preference()

dcooper16:
This PR extends `run_server_preference()` to list every cipher supported by each protocol even in cases in which the server does not enforce a preference order.

For protocols where the server enforces a cipher order the list of supported ciphers is ordered by server preference (as now). For protocols where the server does not enforce a cipher order, the ciphers are listed by encryption strength (as `run_cipher_per_proto()` does).

In order to implement this, `ciphers_by_strength()` was extended to offer a non-wide mode.

This PR is currently marked as a work-in-progress since the output needs improvement:

- The section is labelled "Cipher order" even if the server does not enforce a cipher order.
- In wide mode, the text explaining for which protocols ciphers are ordered by server order versus strength may not be clear.
- In non-wide mode, there is no indicator at all of which ordering is being used for each protocol.
- I am not sure whether the JSON/CSV output (particularly the labels) are the best for automated processing (different labels for the list of supported ciphers depending on which way they are ordered).

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3970,7 +3970,7 @@ run_allciphers() {
 # are good or bad) and list them in order to encryption strength.
 ciphers_by_strength() {
      local proto="$1" proto_hex="$2" proto_text="$3"
-     local using_sockets="$4"
+     local using_sockets="$4" wide="$5"
      local ossl_ciphers_proto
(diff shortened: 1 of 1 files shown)


drwetter:
I merged this into #1583 to get this finalized.

drwetter:
Thaanks!

PR #1583, opened 2020-04-23
 WIP: Extended run_server_preference()

drwetter:
This is a working branch for @dcooper16's PR #1580. 

Aim is eventually to ``run server_preference()`` in wide mode. This PR is supposed to help to get this finalized an merged into 3.1dev

### Details from #1580:

This PR extends `run_server_preference()` to list every cipher supported by each protocol even in cases in which the server does not enforce a preference order.

For protocols where the server enforces a cipher order the list of supported ciphers is ordered by server preference (as now). For protocols where the server does not enforce a cipher order, the ciphers are listed by encryption strength (as `run_cipher_per_proto()` does).

In order to implement this, `ciphers_by_strength()` was extended to offer a non-wide mode.

This PR is currently marked as a work-in-progress since the output needs improvement:

- The section is labelled "Cipher order" even if the server does not enforce a cipher order.
- In wide mode, the text explaining for which protocols ciphers are ordered by server order versus strength may not be clear.
- In non-wide mode, there is no indicator at all of which ordering is being used for each protocol.
- I am not sure whether the JSON/CSV output (particularly the labels) are the best for automated processing (different labels for the list of supported ciphers depending on which way they are ordered).

[Details from #1580  (done)]

### DIFF
--- a/doc/testssl.1
+++ b/doc/testssl.1
@@ -46,10 +46,10 @@ Any OpenSSL or LibreSSL version is needed as a helper\. Unlike previous versions
 2) standard cipher categories to give you upfront an idea for the ciphers supported
 .
 .P
-3) checks forward secrecy: ciphers and elliptical curves
+3) server's cipher preferences (server order)
 .
 .P
-4) server preferences (server order)
+4) forward secrecy: ciphers and elliptical curves
 .
 .P
 5) server defaults (certificate info, TLS extensions, session information)
@@ -61,10 +61,7 @@ Any OpenSSL or LibreSSL version is needed as a helper\. Unlike previous versions
 7) vulnerabilities
 .
 .P
-8) testing each of 370 preconfigured ciphers
-.
-.P
-9) client simulation
+8) client simulation
 .
 .SH "OPTIONS AND PARAMETERS"
 Options are either short or long options\. Any long or short option requiring a value can be called with or without an equal sign\. E\.g\. \fBtestssl\.sh \-t=smtp \-\-wide \-\-openssl=/usr/bin/openssl <URI>\fR (short options with equal sign) is equivalent to \fBtestssl\.sh \-\-starttls smtp \-\-wide \-\-openssl /usr/bin/openssl <URI>\fR (long option without equal sign)\. Some command line options can also be preset via ENV variables\. \fBWIDE=true OPENSSL=/usr/bin/openssl testssl\.sh \-\-starttls=smtp <URI>\fR would be the equivalent to the aforementioned examples\. Preference has the command line over any environment variables\.
--- a/doc/testssl.1.html
+++ b/doc/testssl.1.html
@@ -123,9 +123,9 @@ linked OpenSSL binaries for major operating systems are supplied in <code>./bin/
 
 <p>2) standard cipher categories to give you upfront an idea for the ciphers supported</p>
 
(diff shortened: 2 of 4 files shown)


drwetter:
> In non-wide mode, there is no indicator at all of which ordering is being used for each protocol.

I wonder whether we should offer non-wide mode here. To me it would make more sense to not run ``run_allciphers()`` instead.

drwetter:
> The section is labelled "Cipher order" even if the server does not enforce a cipher order

I phrased that different. But at least in wide mode I am not as happy with those four lines before the cipher per proto listing comes

drwetter:
New open points:

- cipher_pref_check() doesn't save to PROTOS_OFFERED (was there before)
  (just stumbled over this but how about we also use get_protocol() / parse_tls_serverhello()
- do we want run_allciphers() to be started by default?
- $WIDE per default for run_cipher_per_proto() ?
- probably better not to display text in round square brackets when there's no cipher:

```
Hexcode  Cipher Suite Name (OpenSSL)       KeyExch.   Encryption  Bits     Cipher Suite Name (IANA/RFC)
-----------------------------------------------------------------------------------------------------------------------------
SSLv2 (listed by strength)
SSLv3 (server order)
TLSv1 (server order)
TLSv1.1 (server order)
TLSv1.2 (server order)
 xc02c   ECDHE-ECDSA-AES256-GCM-SHA384     ECDH 256   AESGCM      256      TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
[..]
```
- when a server has no preference at all it shows in wide mode (e.g. dev.testssl.sh as of now)
```
Has server cipher order?     no (NOT ok) -- only for TLS 1.3
 Negotiated protocol          TLSv1.3
 Negotiated cipher            TLS_AES_256_GCM_SHA384, 253 bit ECDH (X25519)
 Cipher listing per protocol

Hexcode  Cipher Suite Name (OpenSSL)       KeyExch.   Encryption  Bits     Cipher Suite Name (IANA/RFC)
-----------------------------------------------------------------------------------------------------------------------------
SSLv2
 -
SSLv3
 -
TLSv1 (no server order, thus listed by strength)
 xc014   ECDHE-RSA-AES256-SHA              ECDH 521   AES         256      TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA
[..]

```

drwetter:
See pic  from a default run, and log of last commit. Comments?


Screenshot_20200424_134431

drwetter:
closed:
- <s>do we want run_allciphers() to be started by default?</s> no
- <s>$WIDE per default for run_cipher_per_proto() ?</s>
- <s>when a server has no preference at all it shows in wide mode (e.g. dev.testssl.sh as of now)</s> my configuration mistake on the server side

open:
- cipher_pref_check() doesn't always save to PROTOS_OFFERED (``if [[ -n "$order" ]]``) <sup>1</sup>
- probably better not to display text in round square brackets when there's no cipher --\> more difficult one
- no ``\n - `` when protocol was not known not to be supported before



<sup>1</sup> General thing: how about we also use get_protocol() / parse_tls_serverhello()

drwetter:
the point _probably better not to display text in round square brackets when there's no cipher --> more difficult one_ is a nice to have but not a thing which MUST be there.

So to me all is fine and I'll merged this soon unless travis complains

PR #1585, opened 2020-04-23
Fix run_logjam() in --ssl-native mode (3.0 branch)

dcooper16:
This commit fixes the same issue as #1584, but in the 3.0 branch.


drwetter:
Thanks again

PR #1586, opened 2020-04-23

dcooper16:
This PR fixes a couple of issues related to the use of testssl.sh with OpenSSL 3.0.0-alpha1.

First, when the command line includes an unknown option (e.g., -ssl2), OpenSSL 3.0.0-alpha responds with "Unknown option: -ssl2" rather than "Option unknown option -ssl2". This PR addresses this by making the check for "unknown option" case insensitve.

Second, when printing a DH key, OpenSSL 3.0.0-alpha1 labels the prime and the generator using "prime P:" and "generator G:" rather than just "prime:" and "generator:". This PR addresses this by changing testssl.sh to match on either string.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4794,7 +4794,7 @@ run_client_simulation() {
 #
 locally_supported() {
      [[ -n "$2" ]] && out "$2 "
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, David!

People use old versions of testssl.sh with newer distributions. Thus I think it would be useful also for the 3.0 branch, if you don't mind

PR #1594, opened 2020-04-29

dcooper16:
This commit makes the same changes to the 3.0 branch as #1586 makes to the 3.1dev branch.


drwetter:
Thank you, David!

PR #1595, opened 2020-04-29
Ticketbleed and TLS 1.3

dcooper16:
`run_ticketbleed()` and `sub_session_ticket_tls()` each include one call to `$OPENSSL s_client`. For each of these calls the expected response is a TLS 1.2 or earlier ServerHello. However, if $OPENSSL supports TLS 1.3, then a TLS 1.3 ClientHello will be sent.

This PR fixes this problem in two ways. For the call in `run_ticketbleed()`, "-no_tls1_3" is added to the command line if $OPENSSL supports TLS 1.3. For the call in `sub_session_ticket_tls()`, this PR changes the function so that the same ClientHello version is sent as will sent by run_ticketbleed() via sockets.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -15134,13 +15134,14 @@ run_ccs_injection(){
 }
 
 sub_session_ticket_tls() {
+     local tls_proto="$1"
      local sessticket_tls=""
      #FIXME: we likely have done this already before (either @ run_server_defaults() or at least the output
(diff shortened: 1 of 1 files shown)


drwetter:
Thank you, David! 

This would be great for 3.0 too. ( I tried by myself to merge your PR also into 3.0  using git commands, so that the attribution who did the patch is still available. Unless just using ``patch`` there were a lot of merge conflicts which had nothing to do with your changes. One was even in Readme.md)

PR #1597, opened 2020-04-30
Use $HAS_X25519 and $HAS_X448

dcooper16:
`generate_key_share_extension()` and `prepare_tls_clienthello()` currently check the $OPENSSL version number to determine whether X25519 and X448 are supported. The commit changes these functions to use $HAS_X25519 and $HAS_X448.


drwetter:
thanks!

PR #1598, opened 2020-04-30
Improve compatibility with LibreSSL 3.0.2 and earlier

dcooper16:
This commit addresses two compatibility issues with LibreSSL.

First, with LibreSSL, `$OPENSSL s_client` does not support the "-curves" option, so the "-groups" option needs to be used instead. Note that with LibreSSL, the command line `$OPENSSL s_client -groups $curve -connect invalid.` will not work, as it will complain "no port defined," but will not indicate whether the specified curve is supported. Adding a port number fixes that problem. (There does not seem to be a need to include a port number for other tests, such as whether the "-curves" option itself is supported.)

Second, including "-out -" in the command line for `$OPENSSL genpkey` causes LibreSSL to create a file with the name "-" if the algorithm is supported. This is not an issue at the moment, since LibreSSL's genpkey does not support X25519 or X448. However, genpkey with both OpenSSL and LibreSSL uses stdout as the default output if no "-out" is specified, so the "-out -" is not necessary.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -297,6 +297,7 @@ OSSL_VER_MINOR=0
 OSSL_VER_APPENDIX="none"
 CLIENT_PROB_NO=1
 HAS_DH_BITS=${HAS_DH_BITS:-false}       # initialize openssl variables
+HAS_CURVES=false
 OSSL_SUPPORTED_CURVES=""
 HAS_SSL2=false
 HAS_SSL3=false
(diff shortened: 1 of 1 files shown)


drwetter:
Good point, David!

As this is somewhat recent, it would make sense to backport this to 3.0 too. I could do that with git diff, then patch. But then the git log misses your name. (Don't know whether similar would be possible just using git commands)

PR #1602, opened 2020-05-01
Improve compatibility with LibreSSL 3.0.2 and earlier

dcooper16:
This commit makes the same changes as #1598, but for the 3.0 branch.


drwetter:
Thanks!

PR #1609, opened 2020-05-04
Check server's response

dcooper16:
There is code at the beginning of `parse_tls_serverhello()` that checks whether the server's response appears to consist of a sequence of messages of the form \<protocol\>\<content type\>\<content\>. However, at the moment the check is only performed if `$do_starttls` is false. This commit changes `parse_tls_serverhello()` so that the check is always performed.


drwetter:
Ok, thanks!

drwetter:
Care for a 3.0 backport? (I would be doing that otherwise)

PR #1616, opened 2020-05-06
tls_sockets() return value

dcooper16:
There is one place in `parse_tls_serverhello()` that returns 8 if the server's response is not well-formed TLS. However, there is no code in testssl.sh that is prepared to handle this return value. Every function except `run_protocols()` only distinguishes between 0, 2, and everything else. `run_protocols()`, however, gets confused if `tls_sockets()` returns a value that it is not expecting. So, this commit changes `parse_tls_serverhello()` to return 1 whenever the server's response can not be parsed.


drwetter:
Thanks!

PR #1618, opened 2020-05-06
Fix #1433

### LINKED ISSUE #1433: FYI: OpenSSL SECLEVEL
OpenSSL 1.1.0 introduced the ability to specify a security level: https://github.com/openssl/openssl/blob/master/doc/man3/SSL_CTX_set_security_level.pod. By default the security level is set to 1 unless a compile-time option is used to set the default a different value. The security level may also be set by the command line, e.g.,
```
openssl s_client -cipher @SECLEVEL=0:ALL -connect 127.0.0.1:443
```
At the moment, testssl.sh does not use the `@SECLEVEL=n` directive, but we may want to look into using `s_client_options()` to add it in some circumstances. The particular reason for suggesting this is https://github.com/openssl/openssl/pull/10786 and https://github.com/openssl/openssl/pull/10787. https://github.com/openssl/openssl/pull/10787 notes that:

> This has as effect that SHA1 and MD5+SHA1 are no longer supported at security level 1. This has as effect that TLS < 1.2 is no longer supported at the default security level of 1, and that you need to set the security level to 0 to use TLS < 1.2.

However, even without https://github.com/openssl/openssl/pull/10786 and https://github.com/openssl/openssl/pull/10787, OpenSSL 1.1.0 and 1.1.1 will test for more weak ciphers if the security level is set to 0 rather than 1.

dcooper16:
This commit fixes #1433 by adding "@SECLEVEL=0" to the `$OPENSSL s_client` and `$OPENSSL ciphers` command lines if that option is supported. Adding this option configures OpenSSL to support some weak ciphers that it would not use in the default configuration.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -312,6 +312,7 @@ HAS_PKEY=false
 HAS_NO_SSL2=false
 HAS_NOSERVERNAME=false
 HAS_CIPHERSUITES=false
+HAS_SECLEVEL=false
 HAS_COMP=false
 HAS_NO_COMP=false
 HAS_ALPN=false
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, David, looks good!

<s>Maybe distributions using 3.0 with OpenSSL 1.1.1 (which won't use our deliberately rott^Wold openssl) would benefit from it too. Care for a backport?</s>

PR #1619, opened 2020-05-06
Fix #1576

### LINKED ISSUE #1576: Proposed feature: More information on certificate public keys
The discussion in #1573 about DH groups in `run_fs()` and `run_logjam()` led me to think about the subject public key in `certificate_info()`. If the server supports ephemeral (EC)DH key exchange, we report the elliptic curves and DH groups used in `run_fs()` and `run_logjam()`, but when the server's certificate contains a EC or DH key, we just report the key's length.

Perhaps we could extend `certificate_info()` to similarly report additional information. For an EC subject public key, report which curve. In the unlikely event of a DH subject public key, report the DH group (is it a common prime?).

I checked SSL Labs, and it just reports the length for an EC subject public key. However, for an RSA key it reports the value of the exponent (e.g., "e 65537"), which we don't.

Would this be a useful feature?

dcooper16:
This commit adds additional information to the "Server key size" line for a certificate if the subject public key is RSA, ECDSA, or DH.

For RSA it show the public exponent. For ECDSA, it shows the curve. For DH, it shows the group used, if it is a common prime.

Here is an example of the "Server key size" for 4 different certificates:
```
   Server key size              RSA 2048 bits (exponent is 65537)
   Server key size              EC 256 bits (curve P-256)
   Server key size              DSA 2048 bits
   Server key size              DH 4096 bits (RFC7919/ffdhe4096)
```

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -8547,6 +8547,7 @@ certificate_info() {
                  if [[ -n "$cert_spki_info" ]]; then
                       out " (exponent is $cert_spki_info)"
                       json_msg+=" (exponent is $cert_spki_info)"
+                      [[ $cert_spki_info -eq 1 ]] && set_grade_cap "F" "RSA certificate uses exponent of 1"
                  fi
                  ;;
           "EC")  cert_spki_info="${cert_txt##*Subject Public Key Info:}"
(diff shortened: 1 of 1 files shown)


drwetter:
Cool, thanks!

PR #1620, opened 2020-05-06
Improve key exchange grading

dcooper16:
This PR makes a couple of improvements to `set_key_str_score()`.

It rates (finite-field) DH keys the same as RSA and DSA keys.

Second, in the case of a server that has more than one certificate, the current code sets $KEY_EXCH_SCORE based on the length of the public key in the last certificate that is parsed. This PR changes `set_key_str_score()` so that $KEY_EXCH_SCORE is set based on the weakest public key.

Note that there is still the issue that the key exchange score does not take into account any ephemeral keys used. However, that needs to be addressed by calling `set_key_str_score()` from `run_fs()` and `run_logjam()`, as `certificate_info()` cannot provide information about ephemeral keys.


drwetter:
Thank you!

PR #1623, opened 2020-05-07
Fix copy and paste error

dcooper16:
This PR fixes a bug that was introduced by an accidental copy and paste.


drwetter:
Ouch, thanks

PR #1624, opened 2020-05-07
Fix copy and paste error

dcooper16:
This PR fixes a bug in the 3.0 branch that was introduced by an accidental copy and paste.


drwetter:
Thanks again! That one was older than a year

PR #1629, opened 2020-05-08
Rate RSA e=1 as CRITICAL

dcooper16:
PR #1619 set the grade cap to 'F' if the server has a certificate with an RSA public key with e=1. However, it did not change the rating in the JSON/CSV output. This commit changes the cert_keySize rating to CRITICAL for an RSA key with e=1, regardless of the size of the modulus. It also uses `pr_svrty_critical()` to print the exponent in this case.


drwetter:
Thanks, David

PR #1631, opened 2020-05-08
Implement $HAS_SIGALGS

dcooper16:
The "-sigalgs" option is used in `get_server_certificate()` to obtain certificates the server uses with TLS 1.3. `get_server_certificate()` is currently designed to use $OPENSSL, if $OPENSSL supports TLS 1.3.

LibreSSL 3.1.{0,1} has added client support for TLS 1.3, but does not support the "-sigalgs" option. So, this commit determines whether the "-sigalgs" option is supported, and if it isn't, then uses tls_sockets().

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -307,6 +307,7 @@ HAS_SSL3=false
 HAS_TLS13=false
 HAS_X448=false
 HAS_X25519=false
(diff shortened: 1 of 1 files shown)


drwetter:
Thaanks!

PR #1637, opened 2020-05-11
[Rating] simple DH group length

magnuslarsen:
A simple implementation, that rates the DH group length in `run_logjam()`.
This could probably be implemented in the `pr_dh_quality()` function too, however that seems excessive.

I did remove the lowest key size check, for readability. When capped at an F, the score isn't calculated anyway, so the additional check is unnecessary.

The previous grade_caps where guesses. The rating spec does not differ from known or unknown groups.. only the group size matters.

### DIFF
--- a/doc/testssl.1.md
+++ b/doc/testssl.1.md
@@ -145,7 +145,7 @@ in `/etc/hosts`.  The use of the switch is only useful if you either can't or ar
 
 `--phone-out` Checking for revoked certificates via CRL and OCSP is not done per default. This switch instructs testssl.sh to query external -- in a sense of the current run -- URIs. By using this switch you acknowledge that the check might have privacy issues, a download of several megabytes (CRL file) may happen and there may be network connectivity problems while contacting the endpoint which testssl.sh doesn't handle. PHONE_OUT is the environment variable for this which needs to be set to true if you want this.
 
-`--add-ca <cafile>` enables you to add your own CA(s) for trust chain checks. `cafile` can be a single path or multiple paths as a comma separated list of root CA files. Internally they will be added during runtime to all CA stores. This is (only) useful for internal hosts whose certificates is issued by internal CAs. Alternatively 
+`--add-ca <cafile>` enables you to add your own CA(s) for trust chain checks. `cafile` can be a single path or multiple paths as a comma separated list of root CA files. Internally they will be added during runtime to all CA stores. This is (only) useful for internal hosts whose certificates is issued by internal CAs. Alternatively
 ADDTL_CA_FILES is the environment variable for this.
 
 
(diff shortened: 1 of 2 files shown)


drwetter:
Sorry about my lack of response. I looked at it a few days back but I figured I need to look closer into it. Bear with me pls.

magnuslarsen:
No rush! This was just a very simple start. Better than just rating the key size :)

magnuslarsen:
I pushed some changes:
* merged 3.1dev (resolve merge conflicts)
* I missed a TODO comment; now removed
* The output is now differentiated, which makes it way more readable
* I allowed multiple key exchanges grade caps (if the score is equal) - even if there are multiple certificates, we won't get duplicate grade caps, since the grade caps are uniquely sorted. This does although allow us to report on both weak cert key and ephemeral keys:
```
 Grade cap reasons            Grade capped to B. Using a weak ephemeral DH key (DH parameters)
                              Grade capped to B. Using a weak key
```

drwetter:
Sorry, I thought I did comment on this but obviously I did not.

> I did remove the lowest key size check, for readability. When capped at an F, the score isn't calculated anyway, so the additional check is unnecessary.

I thought that maybe we should just keep the check (and KEY_EXCH_SCORE=20) but don't set the grade_cap as later on in ``let c2_score=$KEY_EXCH_SCORE`` and subsequently it'll be more clear (20\*30\/100 vs. 40\*30\/100).

magnuslarsen:
In rough code; this is what you would like?

```bash
# set_key_str_score()
if [[ $size -lt 512 ]] && [[ $KEY_EXCH_SCORE -ge 20 ]]; then
      let KEY_EXCH_SCORE=20
elif [[ $size -lt 1024 ]] && [[ $KEY_EXCH_SCORE -ge 40 ]]; then
      let KEY_EXCH_SCORE=40
elif [[ $size -lt 2048 ]] && [[ $KEY_EXCH_SCORE -ge 80 ]]; then
      let KEY_EXCH_SCORE=80
elif [[ $size -lt 4096 ]] && [[ $KEY_EXCH_SCORE -ge 90 ]]; then
      let KEY_EXCH_SCORE=90
fi


# run_rating()
## Category 2
if [[ $KEY_EXCH_SCORE -le 40 ]]
     set_grade_cap "F" "Using insecure $output_type"
elif [[ $KEY_EXCH_SCORE -le 80 ]]
     set_grade_cap "B" "Using weak $output_type"
fi

let c2_score=$KEY_EXCH_SCORE
let c2_wscore=$c2_score*30/100

```

drwetter:
Thanks @magnuslarsen . That would be even better I think.

magnuslarsen:
Hi @drwetter, hereby changed :-) And I agree. This makes the set_key_str_score() function much more simple

drwetter:
Thank you @magnuslarsen  and sorry for the delay

PR #1638, opened 2020-05-11
Fix $SHOW_SIGALGO

dcooper16:
This PR fixes two issues related to $SHOW_SIGALGO.

First, `cipher_pref_check()` does not show the signature algorithm if any of the ciphers were found using `tls_sockets()`, since the call to `tls_sockets()` does not specify that the server's certificate should be extracted.

Second, in `run_beast()` the call to `tls_sockets()` indicates that the server's certificate should be extracted if "$SHOW_SIGALGO" is true, even if "$WIDE" is false. While this does not cause any problems, extracting the certificate is a waste of effort if "$WIDE" is false, since the signature algorithm is not shown in that case.


drwetter:
Thanks,D avid!

Issue #1519, opened 2020-02-27
Show peer signing digest / signed_params hash function


djcater:
OS: `Ubuntu 19.10`

testssl.sh version:
```
testssl.sh       3.0 from https://testssl.sh/
(4dbd9a9 2020-02-15 15:23:07 -- )
commit 4dbd9a98ba503f624b24d00629da0e98d655021c
```
OpenSSL used:
```
./bin/openssl.Linux.x86_64
```
Chrome is looking to deprecate the RSA SHA-1 signing digest: https://bugs.chromium.org/p/chromium/issues/detail?id=658905

This is not the signature algorithm of the certificate, nor the HMAC of the cipher suite, but (I think) the algorithm used to sign the server parameters during the handshake.

An example site (taken from this Mozilla ticket: https://bugzilla.mozilla.org/show_bug.cgi?id=1600449) is https://www.daserste.de/.

In the Firefox developer tools, you can see in the "Network" tab of a request, and then in the "Security" sub-tab, it says: "Signature Scheme": `RSA-PKCS1-SHA1`.

This can also be seen in the OpenSSL output:
```
openssl s_client -connect www.daserste.de:443 -servername www.daserste.de
```
```
Peer signing digest: SHA1
Peer signature type: RSA
```
There is a ticket for adding this to SSL Labs here: https://github.com/ssllabs/ssllabs-scan/issues/465

It would be really useful to add to testssl.sh. Thank you.

dcooper16:
This seems like something that would be a useful addition, but some thought would have to be given as to how to present this information. As you noted, this is the hash algorithm used when signing the ephemeral public key in the ServerKeyExchange message for non-anonymous PFS ciphers. So, it seems that the information should appear in the PFS section, along with the set of "robust" PFS ciphers supported, the set of ECDH curves supported, and the set of DH groups supported.

Some thought would be need to be given about what information to provide and how to provide it.

In TLS 1.3, the signature appears in the CertificateVerify message rather than the ServerKeyExchange. Should testssl.sh show both of these. For TLS 1.1 and below, there isn't a choice of hash algorithm. For RSA the hash is SHA1+MD5 and for (EC)DSA it is SHA-1. Should the hash algorithms used with TLS 1.1 and earlier be shown.

[...]

PR #1642, opened 2020-05-14
Extract server's signature algorithm

dcooper16:
PR #1519 requested that testssl.sh show the signature algorithm that the server uses during the TLS handshake. In TLS 1.3, this appears in the CertificateVerify message. In TLS 1.2 it appears in the ServerKeyExchange message when the chosen cipher suite uses an ephemeral (DH or ECDH) key, except in the case of cipher suites that provide no authentication. This information is not present in TLS 1.1 and earlier, as the hash algorithm to use in these earlier versions of the protocol is hard coded into the specification.

This PR takes a first step towards being able to show the signature algorithm by extending `parse_tls_serverhello()` to extract the signature algorithm when it is present. Matching the output produced by OpenSSL, it outputs two separate lines, the "Peer signature type" (RSA, RSA-PSS, DSA, ECDSA, Ed25519, or Ed448) and the "Peer signing digest" (MD5, SHA1, SHA224, SHA256, SHA384, or SHA512). This will allow the same function to extract the signature algorithm and digest, whether the handshake was performed using `$OPENSSL s_client` or `tls_sockets()`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -12800,6 +12800,7 @@ parse_tls_serverhello() {
      local len1 len2 len3 key_bitstring="" pem_certificate
      local dh_p dh_param ephemeral_param rfc7919_param
      local -i dh_p_len dh_param_len
+     local peering_signing_digest=0 peer_signature_type=0
 
      DETECTED_TLS_VERSION=""
      [[ $DEBUG -ge 1 ]] && echo > $TMPFILE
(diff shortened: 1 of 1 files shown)


drwetter:
cool, thank you! 

Should weak signature algorithms be penalized according to the SSL Labs rating guide?

dcooper16:
I couldn't find anything at https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide. It only seems to mention certificate signatures.

The next step for testssl.sh should be to change `run_fs()` in some way to report information about the signature algorithm(s) used by the server in ServerKeyExhange (TLS 1.2) and CertificateVerify (TLS 1.3) messages. If this should affect the grade at all, that will probably be the place to make the change.

PR #1649, opened 2020-06-01
Fix Shellcheck SC2034 issues

dcooper16:
This commit fixes several issues related to Shellcheck issue SC2034: unused variables.

In most cases variables are declared in a function, but are referenced later. The exceptions are:

* `SESS_RESUMPTION` is declared and values are assigned to it, but it us never used. (Same applies for `not_new_reused` in `sub_seession_resumption()`.)

* In `run_cipherlists()`, there is a typo in the declaration of sslv2_tdes_ciphers.

* In `get_caa_rr_record()`, `hash`, `len`, and `line` are used but not declared.


drwetter:
Thanks for the cleanup!

PR #1651, opened 2020-06-08
Fix missing spaces

dcooper16:
In some cases when the Trust finding is printed, there is no space between the results when SNI is used and the results without SNI (which appear in paraenthesis). This commit adds the missing space.


drwetter:
Thanks!

Issue #1555, opened 2020-04-01
Partial wildcard certificates


Pkuutn:
Should "partial wildcard certificates" be trusted?

I have a cert with Common Name "*-test.my.domain", and while connecting host "en-host1-test.my.domain" got the message:

 Trust (hostname)             certificate does not match supplied URI (same w/o SNI)

dcooper16:
It seems that this should be recognized as a match, so more information would be needed to try to find out why the match is not being recognized. The wildcard matching rules that we implemented are those from [RFC 6125, Section 6.4.3](https://www.rfc-editor.org/rfc/rfc6125.html#section-6.4.3). We support the "MAY" part, but do not allow things described as "SHOULD NOT".

Perhaps you can read the rules in that section to see if there is something in your common name or domain name that would cause the match not to work. If not, then we would need the actual domain name in order to be able to run tests to try to find out why it isn't matching.

Pkuutn:
Sorry for late reply.

As I've done some investigation, partial wildcards are recognized Ok in SAN section of certificate via "DNS:*-test.local" statement i.e. but not in CN part.
I'm not sure if RFC states this clear...

Thank you anyways!

dcooper16:
> As I've done some investigation, partial wildcards are recognized Ok in SAN section of certificate via "DNS:*-test.local" statement i.e. but not in CN part.

I've looked at the source code and testssl.sh checks both the SAN and the CN for a wildcard match. So, if testssl.sh is finding a match when the wildcard appears in the SAN but not when it appears in the CN, then I have only one other guess at the moment. Before matching the server's DNS name against the CN in the certificate, testssl.sh tries to determine whether the CN contains a DNS name:
```
     # If the CN contains any characters that are not valid for a DNS name,
     # then assume it does not contain a DNS name.
     [[ -n $(sed 's/^[_\.a-zA-Z0-9*\-]*//' <<< "$cn") ]] && return $subret
```
If the CN string contains any characters other than a letter, number, underscore ('_'), hyphen ('-'), period ('.'), or the wildcard character ('*'), then testssl.sh assumes the CN does not contain a DNS name and so doesn't try to match it against the server's DNS name. DNS names in the SAN are not checked in this way since the issuer of the certificate has specifically indicated that those names are DNS names.

> I'm not sure if RFC states this clear...

[...]

PR #1652, opened 2020-06-08
Fix and enhance CN matching

dcooper16:
PR #1373 changed `get_cn_from_cert()` to handle certificate subject names that include more than one CN attribute. It did this by converting newline characters to spaces. It seems that this resulted in a space character being added to the end of the string returned by `get_cn_from_cert()` even in the case that the subject name only included one CN attribute. The presence of the space character in the returned value caused `compare_server_name_to_cert()` to determine that the CN attribute did not contain a DNS name (since DNS names cannot include spaces), and so `compare_server_name_to_cert()` reports that the server name does not match against the CN in the subject. This may be the reason for the problem noted in #1555.

This commit fixes the above problem and also enhances the matching of the CN in the subject name against the server's name. Currently, `compare_server_name_to_cert()` assumes that the subject field contains at most one CN attribute. However, as noted in #1373, some certificates include subject names with more than one CN attribute, and [RFC 6125](https://www.rfc-editor.org/rfc/rfc6125.html#section-6.2.2) (Section 6.2.2) indicates that the certificate subject name may include more than one CN, with each specifying a different DNS name.

So, in addition to fixing the problem with the space character, this commit also enhances the CN matching to work even if the certificate includes more than one CN attribute in the subject name.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7804,9 +7804,9 @@ wildcard_match()
 
 compare_server_name_to_cert() {
      local cert="$1"
-     local servername cn dns_sans ip_sans san dercert tag
+     local servername cns cn dns_sans ip_sans san dercert tag
      local srv_id="" xmppaddr=""
-     local -i i len len1
+     local -i i len len1 cn_match=0
      local -i subret=0             # no error condition, passing results
 
      HAS_DNS_SANS=false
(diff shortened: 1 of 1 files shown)


drwetter:
Thank you, David!

PR #1656, opened 2020-06-18
[Rating] Clearer grade cap reasons

magnuslarsen:
fixes #1655 
It now looks like this, which makes clears up some confusion:
```
 Rating (experimental) 

 Rating specs (not complete)  SSL Labs's 'SSL Server Rating Guide' (version 2009q from 2020-01-30)
 Specification documentation  https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide
 Protocol Support (weighted)  0 (0)
 Key Exchange     (weighted)  0 (0)
 Cipher Strength  (weighted)  0 (0)
 Final Score                  0
 Overall Grade                T
 Grade cap reasons            Grade capped to T. Issues with the chain of trust (expired)
                              Grade capped to T. Certificate expired
                              Grade capped to B. TLS 1.1 offered
                              Grade capped to B. TLS 1.0 offered
                              Grade capped to A. Problems with HTTP Public Key Pinning (HPKP)
                              Grade capped to A. HSTS max-age is too short

 Done 2020-06-18 21:11:48 [  70s] -->> 81.169.166.184:443 (borken.testssl.sh) <<--
```

I want to extend this PR, to make the rating output consistent with STARTTLS too. I have a few options in mind:
1. Change the current check to `set_grade_cap "t" "STARTTLS encryption is opportunistic. Rating it would be misleading"`
This would still list other grade cap reasons, however, no calculations would be performed.
2. Like above, but clear the `$GRADE_CAP_REASONS` first, so no other grade caps would be shown.
3. Change the current check to `grade_cap_warning "STARTTLS encryption is opportunistic. The grade is worth very little"`
This _would_ calculate an actual grade, with the above text yellow printed. A proper rating can help a admin configure proper TLS. We should just warn very clearly that STARTTLS is opportunistic.
(would also make the best grade possible an A-). Here is an example:
image

For all 3, this would clear out a chunk of duplicate code; and the "UI" would be consistent with non-STARTTLS ratings :)

PR #1657, opened 2020-06-19
[Rating] STARTTLS output styling

magnuslarsen:
As shortly discussed in #1656, rating STARTTLS connections should be consistent with non-STARTTLS connections, in both UI and fileout (csv/json) output.

This lists all the reasons for grade caps and warnings, and calculates an actual grade. This does however limit the score to `A-` (meaning that `A+` is impossible to get, using STARTTLS)

BEFORE:
image
AFTER:
image

### DIFF
--- a/doc/testssl.1.md
+++ b/doc/testssl.1.md
@@ -407,6 +407,10 @@ As of writing, these checks are missing:
 * Exportable key exchange - should give **40** points in `set_key_str_score()`
 * Weak key (Debian OpenSSL Flaw) - should give **0** points in `set_key_str_score()`
 
+#### STARTTLS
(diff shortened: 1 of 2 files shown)


drwetter:
Thanks, @magnuslarsen . Excuse my delay.

Maybe here's a slight misunderstanding. Actually I still think it is wrong to tell people using STARTTLS even when all is superb to give anything better than T. (Normally) there's no trust bc of a potential  MiTM scenario (e.g. https://github.com/tintinweb/striptls. Moxie had a similar tool. At least he demo'd that at some LV conference IIRC).

We can display the rest though.

So what I would suggest to leave overall grade always @ T.

Then: I am thinking on phrasing the grade warning better but atm I lack fantasy. How about: _Encryption via STARTTLS is not mandatory (=opportunistic). A grade better than T would lead to a false sense of security_  (maybe too long)

magnuslarsen:
How about this then:
image

The text is not longer than other parts of a normal output

drwetter:
Thanks @magnuslarsen !

PR #1659, opened 2020-06-23
Add wiresharked Android 7.0 (native)

csett86:
After being bitten by https://stackoverflow.com/questions/39133437/sslhandshakeexception-handshake-failed-on-android-n-7-0
I add a wiresharked Android 7.0 to reflect that bug in Android 7.0.


typingArtist:
See also #540.

drwetter:
Ok, thanks. What app / did you use?

And: you're sure you want to leave the server name in there?

csett86:
I used RiotX (which uses Android system libraries and thus was no longer working just on Android 7.0 once our hoster switched to EC P-384 certificates) and the server name is fine in there, its a public service operated by myself.

drwetter:
Ok, thank you!

Do you mind handing in a PR for the 3.0 branch as well?

csett86:
PR for 3.0 here: #1661

PR #1668, opened 2020-06-25
Polish STARTTLS rating output

drwetter:
Moved the sentence ~i "A grade better than T would lead to a false sense of security"  to the documentation. No reason for excuses in the output. ;-) Explanation fits better in the doc.

Also the doc was amended with corresponding patches of the [md file](069c5ae917709f2fd1540cc4ce6bd4334e0db47b) in the roff and html doc.
    
See also #1657


drwetter:
This is how it looks like

Screenshot_20200626_103746

Issue #1311, opened 2019-08-30
Relax recommendation on cipher order


exploide:
Currently, testssl prints a big red warning when a server has no server preferred cipher order.

Mozilla recently relaxed their recommendations regarding cipher order. If only strong cipher suites are supported anyway, why not deciding according to the client's preferences. Maybe it's a phone that wants to optimize for performance on low hardware. Additionally, often browsers are better maintained than web servers updated. So the client's preferences are probably not bad.

So Mozilla decided to prefer server order only for the "old" compatibility configuration.

There were several discussions about this, but this was the outcome so far.
https://github.com/mozilla/ssl-config-generator

I propose that testssl also relaxes its recommendation regarding cipher order.

To be more precise: Either print the warning just in yellow instead of red, or make it dependent on the supported TLS versions and algorithms. E.g. when no cipher concerns are present, make the order setting also no concern.

drwetter:
> Mozilla recently relaxed their recommendations regarding cipher order.

you have a pointer?

Please keep in mind that testssl.sh tests not only HTTP services but almost everything. Realistically nowdays browsers are not a big concern, but I can tell you that there are really bad clients out there.

> If only strong cipher suites are supported anyway, why not deciding according to the client's preferences. 

yes, **if** . Means, it would need a test for every cipher upfront.

exploide:
Thanks for your reply.

> you have a pointer?

See https://ssl-config.mozilla.org/ for the outcome of the popular Mozilla config generator. Note that server cipher order is off for modern and intermediate and only on for old.

Discussion is hard to follow because spread over several issues. E.g. https://github.com/mozilla/ssl-config-generator/issues/48, https://github.com/mozilla/ssl-config-generator/issues/34 and starting at https://github.com/mozilla/server-side-tls/issues/178#issuecomment-506026018

> but I can tell you that there are really bad clients out there

I agree. I do not propose to remove the cipher order rating altogether. But maybe if it would be only written in yellow. Or is a bit more explanatory in which cases it is important.

> Means, it would need a test for every cipher upfront.

Not necessarily. It might be sufficient to utilize the already existing cipher "category" rating from the top. I.e. only warn about cipher order when already warned about NULL, ANON, EXPORT, or LOW.

drwetter:
Thanks for the pointers.

From a server side when you already know what you configure, things might seem easier. Albeit still I think to set server side preference should be best practice in TLS <= 1.2.

[...]

PR #1669, opened 2020-07-06
Separate pr_cipher_quality() into two functions

dcooper16:
This PR separates `pr_cipher_quality()` into two functions, one that returns the quality of a cipher as a numeric rating (`get_cipher_quality()`) and one that prints a cipher based on its quality (`pr_cipher_quality()`). This separation allows `get_cipher_quality()` to be used to determine how good a cipher is without having to print anything. Having this ability would be helpful in implementing the changes suggested in #1311.


drwetter:
Thanks!

PR #1673, opened 2020-07-07
Fix printing of unrecognized option in 3.0 branch

dcooper16:
When testssl.sh is called with an unknown option it prints something like:

     0: unrecognized option "--option"

It should be printing the name of the program rather than "0". This PR fixes that.


drwetter:
Thanks!

PR #1674, opened 2020-07-07
Include cipher quality in JSON and CSV

dcooper16:
run_cipherlists() checks for support for different groups of ciphers, but does not indicate which ciphers in each group are supported. So, for example, if the JSON file indicates that there is a problem with severity level "HIGH" because the "LOW" ciphers are available, there is no clear indication of which of these ciphers are supported by the server.

If run_server_preference() is run with "--color 3", then there will be a visual indication (via color) of the ciphers the server supports that are considered bad, but this information does not appear in the JSON (or CSV) output. The JSON (or CSV) output will include information about every cipher that is supported, but the severity level is always "INFO".

This PR addresses this problem by changing the fileout() calls in ciphers_by_strength() and cipher_pref_check() that output each supported cipher individually so that the "severity" argument is an indication of the quality of the cipher. With this, information about which bad ciphers are supported can easily be found in the JSON/CSV output.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4378,7 +4378,7 @@ ciphers_by_strength() {
                outln "${sigalg[i]}"
                id="cipher$proto"
                id+="_${normalized_hexcode[i]}"
-               fileout "$id" "INFO" "$proto_text  $(neat_list "${normalized_hexcode[i]}" "${ciph[i]}" "${kx[i]}" "${enc[i]}" "${export2[i]}") $available"
+               fileout "$id" "$(get_cipher_quality_severity "${ciph[i]}")" "$proto_text  $(neat_list "${normalized_hexcode[i]}" "${ciph[i]}" "${kx[i]}" "${enc[i]}" "${export2[i]}") $available"
           fi
      done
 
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, David! Do you mind to amend the documentation @ ~/doc/ ?

dcooper16:
> Thanks, David! Do you mind to amend the documentation @ ~/doc/ ?

Hi Dirk,

I took a look at the documentation and I see that the description of `run_server_preference()` is a bit out of date. However, I don't see anywhere in the documentation that really describes what information is included in the JSON/CSV output. Do you have some thought about where in the documentation the change made by the PR could be described?

drwetter:
Hi David,

> Do you have some thought about where in the documentation the change made by the PR could be described?

Not really.  I was just looking at your PR and did a search on  https://github.com/drwetter/testssl.sh/blob/3.1dev/doc/testssl.1.md. and figured that your change is missing. At that point I only had a glance at it and I though we must have had that documented somewhere.

Looking at it again now, yes. Good point :-)

See also #1675 . I think a new paragraph with a heading is needed here. I'll add later an issue for that so it doesn't fall off the table

PR #1685, opened 2020-07-15
Check for bad OCSP intermediate certificates

dcooper16:
This PR checks whether any intermediate certificates provided by the server include an extended key usage extension that asserts the OCSP Signing key purpose.

This PR replaces #1680, which checks for such certificates by comparing the server's intermediate certificates against a fixed list of known bad certificates.


drwetter:
Cool, thank you! Works e.g. for www.dhl.de .

Also great for saving the intermediate certificates in an array

PR #1692, opened 2020-08-03
Update GREASE reference

dcooper16:
The GEASE Internet Draft is now RFC 8701. This PR updates the references.


dcooper16:
This PR does not change any code, just comments and documentation. There is one code change, however, that could be considered.

The code in `run_grease()` was based on draft-ietf-tls-grease-01.txt, which says that any ALPN identifier with a prefix of "ignore/" is reserved as a "GREASE" value. If the server supports ALPN and selects "foo", then a test ClientHello is sent that only offers "ignore/foo". RFC 8701 instead says that the reserved ALPN values are "{0x0A,0x0A}, {0x1A,0x1A}, ..., {0xFA,0xFA}".

So, we can either leave our ALPN GREASE test unchanged, or change it to conform to RFC 8701 by using one of the reserved ALPN values rather than "ignore/foo".

drwetter:
Thanks, David!

> So, we can either leave our ALPN GREASE test unchanged, or change it to conform to RFC 8701 by using one of the reserved ALPN values rather than "ignore/foo".

I would rather recommend to stick to standards, if you don't mind.

Around line 493 in the markup file I started collecting the RFCs which we use. In general I think it's probably not complete but this one should be mentioned too. I will do that tomorrow, or if you like fee free to amend this PR.

drwetter:
Thanks again!

PR #1694, opened 2020-08-04
Align ALPN GREASE test with RFC 8701

dcooper16:
The ALPN GREASE test was written based on draft-ietf-tls-grease-01, which reserved all ALPN identifiers beginning with the prefix "ignore/". This PR changes the test to align with RFC 8701, which instead reserves {0x0A,0x0A}, {0x1A,0x1A}, ..., {0xFA,0xFA}.


drwetter:
Thanks

PR #1695, opened 2020-08-04
Update ETSI ETS test

dcooper16:
The PR updates the test for the "Visibility Information" certificate extension used in the ETSI Enterprise Transport Security protocol.

The main change is to support OpenSSL 3.0.0, which prints more information about subject alternative names that are encoded as otherName. For otherName types for the OpenSSL has no information, it prints "otherName: \<OID\>::\<unsupported\>", where "\<OID\>" is the actual OID for the name type, rather than just "otherName: \<unsupported\>". So, testssl.sh needs to account for the possibility that the OID for the visibility information certificate extension will be printed.

This PR also updates the reference for this extension and changes the name of the function from etsi_etls_visibility_info() to etls_ets_visibility_info() since the name of the protocol was changed from Enterprise TLS (eTLS) to Enterprise Transport Security (ETS).

This commit does not change the output to the terminal or to JSON/CSV, even though those outputs use the previous name of eTLS rather than ETS.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -8008,8 +8008,8 @@ compare_server_name_to_cert() {
 
 # This function determines whether the certificate (arg3) contains "visibility
 # information" (see Section 4.3.3 of
-# https://www.etsi.org/deliver/etsi_ts/103500_103599/10352303/01.01.01_60/ts_10352303v010101p.pdf .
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks!

PR #1696, opened 2020-08-06
Use fewer external function calls

dcooper16:
This PR modifies a few functions to use fewer external function calls. In most cases this involves replacing external function calls with Bash internal functions, but in one case it involves replacing multiple external function calls with one call to `awk`.

This PR makes a few changes to the way that some functions work.

`is_ipv4addr()` and `is_ipv6addr()` will now strictly only accept a string that is an IPv4 (or IPv6) address and nothing else.

A couple of changes were also made to `match_ipv4_httpheader()`. First, lines that match `$excluded_header` (formerly `$whitelisted_header`) are not processed in the while loop. This prevents the excluded header from being output in the case that `$HEADERFILE` includes a non-excluded header with an IPv4 address and an excluded header with a string that looks like an IPv4 address.

The list of excluded headers was also modified to exclude any line that begins "Server: " rather than just lines that begin "Server: PRTG". According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server, the "Server" header describes the software used by the server, so it seems reasonable to expect that this header line will never contain an IPv4 address. Also, looking at some old test results I found cases in which Oracle software version numbers in the Server header were mistakenly matched as IPv4 addresses.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -124,7 +124,7 @@ trap "child_error" USR1
 #
 declare -r VERSION="3.1dev"
 declare -r SWCONTACT="dirk aet testssl dot sh"
-grep -E -q "dev|rc|beta" <<< "$VERSION" && \
+[[ "$VERSION" =~ dev|rc|beta ]] && \
      SWURL="https://testssl.sh/dev/" ||
      SWURL="https://testssl.sh/"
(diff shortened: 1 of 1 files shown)


drwetter:
Very good, David. Thank you!

PR #1705, opened 2020-08-20
Fix missing bind-tools in container

aussieklutz:
Name resolution not working in docker image, due to base image not including bind-tools package.
Updates the Dockerfile to install bind-tools to re-enable name resolution


drwetter:
There's drill.

Besides it is tested and worked. Also dockerhub uses this file. 

Let me know if I missed something.

PR #1707, opened 2020-08-26
`hostname` is not defined by POSIX, use portable `uname -n` instead.

ghen2:


drwetter:
Thanks. First, after reading, I was afraid that some OS don't support this https://en.wikipedia.org/wiki/Portable_Operating_System_Interface#Mostly_POSIX-compliant . But I was wrong

ghen2:
Could you merge this into 3.0 as well?

drwetter:
If you send me a PR ;-)

PR #1714, opened 2020-09-03
Print information about certificate-based client authentication

### LINKED ISSUE #1709: Display whether server requests/requires a Client Certificate
One of the big challenges in debugging TLS/SSL connections is determining whether the server is requesting/requiring a client certificate.  With **openSSL** and with **testssl.sh**, it is unclear. Both tools output nice information about the negotiated protocol, but neglect to complete the handshake sequence afterwards. openssl outputs a cryptic message (if any). 

It would be great if **testssl.sh** could CLEARLY indicate whether: (a) server has requested a client certificate; and (b) whether the connection fully succeeded (i.e., did server connect anyway if no client cert was provided / specified)?  Having spent hours troubleshooting TLS connections with "handshake_failure" messages (a Java 'catch-all'), having this detail would be great.

This topic mentions the issue with openssl. It is possble to detect that an CertificateRequest has been sent from the server.
https://security.stackexchange.com/questions/101511/determine-if-a-server-is-asking-for-a-client-certificate-using-openssl-s-client

IMPLEMENTATION CONSIDERATION:
By using s_client in openssl, the CA list can be viewed and checked. However some servers only request client authentication after a specific URL is requested. To obtain the list in this case it is necessary to use the -prexit command and send an HTTP request for an appropriate page.

dcooper16:
This PR is a first step towards addressing #1709. It attempts to determime whether certificate-based client authentication is (1) not requested, (2) optional, or (3) required. If it is either optional or required, then it extracts the list of CA names (DNs) that the server sends in its CertificateRequest message.

The code for extracting the CA list from the CertificateRequest message seems to be working correctly. However, this commit is incomplete for a couple of reasons. First, it does not produce any new output, it just collects the information. Second, sclient_auth() needs some work.

The current sclient_auth() simply returns 0 if $OPENSSL returned 0. This may be okay if only trying to determine whether certificate-based client authentication is required. However, if it is optional, then the output will include "CertificateRequest", but $OPENSSL will return 0, since the connection was successful even though the client did not provide a certificates.

If $OPENSSL does not return 0, then sclient_auth() checks whether Master-Key is present. This works for TLS 1.2 and earlier, but not for TLS 1.3. So, sclient_auth() needs to be updated to work correctly with TLS 1.3.

The modified version of sclient_auth() will set CLIENT_AUTH and CLIENT_AUTH_CA_LIST for any version of TLS, but the remaining part of the code needs work. As I am not clear on the reason for this code, I need some help with it. Why does the code only look for "CertificateRequest" if "Master-Key" is present? Why is there a check for Session-ID in a function that is supposed to just be checking for client authentication. Why is CLIENT_AUTH set to false if SESSION-ID is absent (this is a no-op since CLIENT_AUTH would already have been false)?

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -277,7 +277,8 @@ declare -r ALPN_PROTOs="h2 spdy/3.1 http/1.1 grpc-exp h2-fb spdy/1 spdy/2 spdy/3
 TEMPDIR=""
 TMPFILE=""
 ERRFILE=""
-CLIENT_AUTH=false
+CLIENT_AUTH="none"
+CLIENT_AUTH_CA_LIST=""
 TLS_TICKETS=false
 NO_SSL_SESSIONID=false
 CERT_COMPRESSION=${CERT_COMPRESSION:-false}  # secret flag to set in addition to --devel for certificate compression
@@ -2223,7 +2224,7 @@ s_client_options() {
 service_detection() {
      local -i was_killed
 
-     if ! "$CLIENT_AUTH"; then
+     if [[ "$CLIENT_AUTH" != require ]]; then
           if ! "$HAS_TLS13" && "$TLS13_ONLY"; then
                # Using sockets is a lot slower than using OpenSSL, and it is
                # not as reliable, but if OpenSSL can't connect to the server,
@@ -2272,7 +2273,7 @@ service_detection() {
                out " $SERVICE, thus skipping HTTP specific checks"
                fileout "${jsonID}" "INFO" "$SERVICE, thus skipping HTTP specific checks"
                ;;
-          *)   if "$CLIENT_AUTH"; then
+          *)   if [[ "$CLIENT_AUTH" == require ]]; then
                     out " certificate-based authentication => skipping all HTTP checks"
                     echo "certificate-based authentication => skipping all HTTP checks" >$TMPFILE
                     fileout "${jsonID}" "INFO" "certificate-based authentication => skipping all HTTP checks"
@@ -2494,7 +2495,7 @@ run_http_date() {
      local spaces="                              "
      jsonID="HTTP_clock_skew"
 
-     if [[ $SERVICE != HTTP ]] || "$CLIENT_AUTH"; then
+     if [[ $SERVICE != HTTP ]] || [[ "$CLIENT_AUTH" == require ]]; then
           return 0
      fi
      if [[ ! -s $HEADERFILE ]]; then
@@ -6443,7 +6444,7 @@ sub_session_resumption() {
                return 1
           fi
      fi
-     "$CLIENT_AUTH" && return 6
+     [[ "$CLIENT_AUTH" == require ]] && return 6
      if ! "$HAS_TLS13" && "$HAS_NO_SSL2"; then
           addcmd+=" -no_ssl2"
      else
@@ -8365,7 +8366,7 @@ certificate_transparency() {
           fi
      fi
 
-     if [[ $SERVICE != HTTP ]] && ! "$CLIENT_AUTH"; then
+     if [[ $SERVICE != HTTP ]] && [[ "$CLIENT_AUTH" != require ]]; then
           # At the moment Certificate Transparency only applies to HTTPS.
           tm_out "N/A"
      else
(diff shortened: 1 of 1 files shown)


dcooper16:
I believe that this PR is now ready to be reviewed and merged. The `s_client_auth()` function has been modified so that it performs checks similar to `sclient_connect_successful()` to determine whether the connection attempt was successful. An indication of whether client authentication is requested, and if so whether it is required or optional, is output as part of `run_server_defaults()`. If client authentication is requested, the CA list provided by the server is output.

One issue this PR does not address is that it doesn't check whether client authentication is requested if a specific URL is provided. The man page for s_client says:

> some servers only request client authentication after a specific URL is requested. To obtain the list in this case it is necessary to use the -prexit option and send an HTTP request for an appropriate page.

Addressing this scenario would be a future enhancement. Another future enhancement could be to extract and print out information about what signature algorithms the server indicates are acceptable (somewhat related to #1519).

testssl.sh
```diff
@@ -20336,26 +20357,147 @@ check_proxy() {
...
+# representation of the name.
+print_dn() {
+     local dn="$1"
+     local cert name
+     local -i len
+
+     # Use $OPENSSL to print the DN by creating a certificate containing the DN
+     # as the issuer and then having $OPENSSL print the issuer field in the
+     # resulting certificate.
+     cert="A003020102020100300A06082A8648CE3D040302${dn}301E170D3139303830353038333030305A170D3139303830353038333030305A30003019301306072A8648CE3D020106082A8648CE3D030107030200FF"
```
drwetter: Mind to comment this in the code?

testssl.sh
```diff
@@ -20336,26 +20357,147 @@ check_proxy() {
...
+     cert+="300A06082A8648CE3D040302030200FF"
+     len=$((${#cert}/2))
+     if [[ $len -lt 128 ]]; then
+          cert="30$(printf "%02x" $len)$cert"
+     elif [[ $len -lt 256 ]]; then
+          cert="3081$(printf "%02x" $len)$cert"
+     else
+          cert="3082$(printf "%04x" $len)$cert"
+     fi
+     name="$(asciihex_to_binary "$cert" | $OPENSSL x509 -issuer -noout -inform DER -nameopt RFC2253 2>/dev/null)"
```
drwetter, line 20388: Short explanation of what RFC2253 means would be great or maybe a link to the section in the RFC

testssl.sh
```diff
@@ -20336,26 +20357,147 @@ check_proxy() {
...
+          [[ -z "$calist" ]] && break
+          len=2*$(hex2dec "${calist:0:4}")
+          dn="${calist:4:len}"
+          calist_string+="$(print_dn "$dn")\n"
+          calist="${calist:$((len+4))}"
+     done
+     [[ -z "$calist_string" ]] && calist_string="empty"
+     tm_out "$calist_string"
+     return 0
+}
```
drwetter, line 20465: Wow :+1:

drwetter:
Thanks a lot David!

If you could address the minor comment that would be awesome.

Any public server to test this against (except secure.cacert.org?) cc @lesterw1

testssl.sh
dcooper16: I tried adding a comment, but I can't really point to a section of the RFC. RFC 2253 is "Lightweight Directory Access Protocol (v3): UTF-8 String Representation of Distinguished Names." So, the entire RFC is about how to represent a DN as a string.

Of course, we don't have to use the RFC 2253 (RFC 4514) string representation. RFC 2253 prints the attributes in the reverse of the order they actually appear in the DN:
> O = CAcert Inc., OU = http://www.CAcert.org, CN = CAcert Class 3 Root

leaving out "-nameopt RFC2253" would give us:

> CN=CAcert Class 3 Root,OU=http://www.CAcert.org,O=CAcert Inc.

Of course, there are many other choices of options to provide to "-nameopt". It's just a question of personal preference.

dcooper16:
> Any public server to test this against (except secure.cacert.org?) 

None that I am aware of. I didn't even know about secure.cacert.org. I just created a local server using Apache and tested against that.

As a follow-on to this PR, I'd like to make the code work in cases in which client authentication is only required if a specific URL path is provided. It seems easy to do this using $GET_REQ11, but it would be nice to be able to test it on more than just a test server I create myself.

drwetter:
Thanks again, David! 

As a general hint: For referring to sections in RFCs what one could do is using the HTML formatted RFC. The sections there provide HTML anchors.

PR #1722, opened 2020-09-12
upd alpine to 3.12

vitaliytv:
because error with libuv required for other packages. For example: 
Error relocating /usr/bin/node: uv_fs_lutime

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.11
+FROM alpine:3.12
 
 RUN apk update && \
     apk upgrade && \


drwetter:
Hi Vitaly,

thanks. We don't use node.js though. There has to be another reason. Also it needs to be tested thoroughly. 

Alpine 3.11.'s EOL is still a year ahead.

Chees, Dirk

vitaliytv:
> Hi Vitaly,
> 
> thanks. We don't use node.js though. There has to be another reason. Also it needs to be tested thoroughly.
> 
> Alpine 3.11.'s EOL is still a year ahead.
> 
> Chees, Dirk

Thank you i will use my fork and I will report the test results

drwetter:
Looking forward into the future I merged this. Thank you

jauderho:
If this is forward looking, should it be updated to 3.13 instead?

drwetter:
@jauderho It should be time, at some point  I was looking yesterday at the release notes which seemed ok. 

It would be awesome though if we find a way running ``prove -t`` in the container, just for testing.

PR #1746, opened 2020-10-15
Force SNI to be the --xmpphost if passed

horazont:
XMPP can be used with SNI in two contexts:

- Standard RFC 6120 STARTTLS-based connections; in that case, SNI is most likely to be ignored, as XMPP uses another way to signal the target domain name (via the `@to` attribute on the stream header, which is already set correctly by testssl.sh). However, setting SNI to a different value than the `@to` attribute may lead to confusion.

- XEP-0368 (XMPP-over-TLS) connections which omit the STARTTLS phase and go right for TLS (and inside that, XMPP). In that case, SNI is obviously required to be correct. XEP-0368 specifies that the SNI name MUST be the domain name of the service (not necessarily the host name of the endpoint, thanks to SRV records).

Hence, this patch forces the SNI name to be the --xmpphost value, if --xmpphost is given. Note that it blatantly ignores whether XMPP is used otherwise.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -4522,10 +4522,11 @@ modify_clienthello() {
                # the SNI extension or replace it with the correct server name.
                sni_extension_found=true
                if [[ -n "$SNI" ]]; then
+                    servername=${XMPP_HOST:-${NODE}}
                     # Create a server name extension that corresponds to $SNI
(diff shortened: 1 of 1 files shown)


drwetter:
Thank you!

Issue #1706, opened 2020-08-25
Feature Request


CuttlefishJones:
Hi would it be possible to add the ability to define your own User Agent string?

definity:
+1

drwetter:
sure. ``--user-agent`` would be in line with wget or cURL. 

Any volunteers?

definity:
I will try to do this... what could possibly go wrong 🤣

definity:
Challenge accepted - #1748

PR #1748, opened 2020-10-16
Added support for custom user agent

definity:
First stab at adding support... it seemed a bit too complicated to try and add UA_CUSTOM everywhere and try to then add checks to deal with in all the places where checks for SNEAKY were set or not, so I chose to just update UA_STD instead.  I wasn't sure if you were declaring all variables as read-only, but after seeing some where you didn't I figured this would be the easiest way to work in support.

Tested the script running and confirmed I was able to see the user-agent in logging reflect what I put on the command-line.

I'm ready for abuse/flogging/feedback.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -229,7 +229,7 @@ else
      MEASURE_TIME=${MEASURE_TIME:-false}
 fi
 DISPLAY_CIPHERNAMES="openssl"           # display OpenSSL ciphername (but both OpenSSL and RFC ciphernames in wide mode)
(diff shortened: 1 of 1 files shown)


definity:
#1706

drwetter:
Thank you!

drwetter:
I try to just declare those kind of things once and not every where. :-)

If you don't mind you can add the feature to the CHANGES file.

And while am at it (asking): below the doc dir are man pages. If you could just add a line there that would be awesome ;-)

PR #1751, opened 2020-10-19
Modified ssl renegotiation attempts to be variable, default 6.

tosticated:
See #1070 for original check and code by @poupas.

Made the SSL Client initiated renegotiation into a variable amount that can be set with an environment variable. Also, please consider this change to set default to 6 attempts.

Original was 4 attempts since node.js blocks after 3 attempts. I found that a lot of servers of customers I run into are blocking after 5 attempts. For example F5 BIG-IP recommends 5 renegotiations: https://support.f5.com/csp/article/K50023125

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * Added support for certificates with EdDSA signatures and pubilc keys
 * --add-ca can also now be a directory with \*.pem files
 * Warning of 398 day limit for certificates issued after 2020/9/1
+* Added environment variable for amount of attempts for ssl renegotiation check
 
 ### Features implemented / improvements in 3.0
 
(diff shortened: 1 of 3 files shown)


drwetter:
Thanks, @tosticated, especially for the f5 link. As a coincidence I am just doing tests at a customer and I was about to label the F5 finding as a risk.

Would you mind to adjust the indentation? (no tabs, spaces please). Please no blanks at the end of line

image

drwetter:
6 would be fine too. If I see that correctly the F5 here gives up at 6 also when 20 is specified.

tosticated:
My apologies, I should have used a proper editor. Fixed the tabs and whitespaces, and thank you for allowing the 6 attempts. What a coincidence that you were testing an F5.

drwetter:
Thanks!

>  What a coincidence that you were testing an F5.

Speaking up here came at the right moment. Otherwise my report would have contained a false positive. Thx

PR #1764, opened 2020-10-30
fix #1762: Stop labeling X-XSS-Protection as green

### LINKED ISSUE #1762: Stop labeling X-XSS-Protection as green
See https://caniuse.com/?search=x-xss-protection (just stumbled over this through a discussion elsewhere, https://news.ycombinator.com/item?id=20472947)

Needs to be done also for 3.0

keisentraut:
This rates X-XSS-Protection as low severity finding. In order to avoid code duplication, I had to refactor the function ```run_security_headers()" a little. Additionally, I have added the new header ```Permissions-Policy```.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -3204,12 +3204,13 @@ run_security_headers() {
                              "Content-Security-Policy-Report-Only OK" \
                              "Expect-CT OK" \
                              "Permissions-Policy OK" \
-                             "X-XSS-Protection LOW" \
+                             "X-XSS-Protection INFO" \
                              "Access-Control-Allow-Origin INFO" \
                              "Upgrade INFO" \
                              "X-Served-By INFO" \
                              "Referrer-Policy INFO" \
-                             "X-UA-Compatible Cache-Control INFO" \
+                             "X-UA-Compatible INFO" \
+                             "Cache-Control INFO" \
                              "Pragma INFO"; do
           read header svrty <<< "${header_and_svrty}"
           [[ "$DEBUG" -ge 5 ]] &&  echo "testing \"$header\" (severity \"$svrty\")"
--- a/testssl.sh
+++ b/testssl.sh
@@ -3185,9 +3185,7 @@ run_cookie_flags() {     # ARG1: Path
 
 
 run_security_headers() {
-     local good_header="X-Frame-Options X-XSS-Protection X-Content-Type-Options Content-Security-Policy X-Content-Security-Policy X-WebKit-CSP Content-Security-Policy-Report-Only Expect-CT"
-     local other_header="Access-Control-Allow-Origin Upgrade X-Served-By Referrer-Policy X-UA-Compatible Cache-Control Pragma"
-     local header header_output
+     local header header_output svrty header_and_svrty
      local first=true
      local spaces="                              "
      local have_header=false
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -3197,36 +3195,43 @@ run_security_headers() {
...
+                             "X-WebKit-CSP OK" \
+                             "Content-Security-Policy-Report-Only OK" \
+                             "Expect-CT OK" \
+                             "Permissions-Policy OK" \
+                             "X-XSS-Protection LOW" \
+                             "Access-Control-Allow-Origin INFO" \
+                             "Upgrade INFO" \
+                             "X-Served-By INFO" \
+                             "Referrer-Policy INFO" \
+                             "X-UA-Compatible Cache-Control INFO" \
```
drwetter, line 3212: This is probably not what you meant? See, e.g. when testing dev.testssl.sh

testssl.sh
```diff
@@ -3197,36 +3195,43 @@ run_security_headers() {
...
+     # X-XSS-Protection is useless and at worst harmful, see https://news.ycombinator.com/item?id=20472947
+     for header_and_svrty in "X-Frame-Options OK" \
+                             "X-Content-Type-Options OK" \
+                             "Content-Security-Policy OK" \
+                             "X-Content-Security-Policy OK" \
+                             "X-WebKit-CSP OK" \
+                             "Content-Security-Policy-Report-Only OK" \
+                             "Expect-CT OK" \
+                             "Permissions-Policy OK" \
+                             "X-XSS-Protection LOW" \
```
drwetter, line 3207: I would not label it in any way. ``0`` was bad maybe a couple of years back, ``1`` was the default in some browsers ``1; mode=block`` was where some people frowned upon. This was/is a lack of testssl.sh not to look at the values.

Moving forward now I think it should just be neutral or to emphasize in an eye friendly manner. "An eye friendly manner" though hasn't been determined yet. And I think we should do one step at a time

drwetter (changes requested):
Thanks, @keisentraut ! It's even better than I had in mind.

Please see my two comments. Feel free to use dev.testssl.sh for testing.

testssl.sh
keisentraut: Yes, this is a typo and should be split into two lines.

testssl.sh
keisentraut: I assume "neutral" == "INFO". This will (as it was before for the ```other_headers```) output the header with a "neutral" color, which is currently ```pr_litecyan```.
There is a TODO comment for the lack of looking at the values. However, evaluating the values is unique per header and I agree with opening a new pull request.

drwetter:
>  I agree with opening a new pull request.

why don't you just commit to your old branch. It should show up here immediately

keisentraut:
> why don't you just commit to your old branch. It should show up here immediately

I was talking about the new feature of checking values of the security header. This will require a lot of refactoring, probably the whole function will need to be split in multiple ones. Also, I'm unsure how rating would work  Is a CSP  like ```script-src: self unsafe-inline;``` bad? It is still better than having no CSP at all...`

This issue is now ready to be merged, please review.

drwetter:
>I was talking about the new feature of checking values of the security header. This will require a lot of refactoring, probably the whole function will need to be split in multiple ones. Also, I'm unsure how rating would work Is a CSP like script-src: self unsafe-inline; bad? It is still better than having no CSP at all...`

Sorry, didn't get that. 

I appreciate your willingness and don't want to stop you. However IMO there are more important issues open. This is a somewhat difficult one and as you indicated some judging will be for sure disputable. 

> This issue is now ready to be merged, please review.

looks good, I'll wait for travis to finish

drwetter:
Thanks a lot!

PR #1766, opened 2020-11-03
.gitignore: ignore HTML scan results like example.com_p443-20201103-1…

keisentraut:
Ignore HTML scan results like below. We cannot ignore ```*.html``` because there are legit HTML files under version control.
```
$ git status
On branch fix-gitignore-ignore-html-scan-results
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	example.com_p443-20201103-1037.html
	example.com_p443-20201103-1042.html

nothing added to commit but untracked files present (use "git add" to track)
```


drwetter:
make sense. Thanks

PR #1797, opened 2020-11-28
Hex2curves

atroost:
This PR contains only two new files and two small adjustments in the documentation file.

One file hexstream2curves is a direct copy of hexstream2ciphers with small changes to make it a unique file with correct variables for the curves file. (curves-mapping)

The curves-mapping.txt file is the mapping file matching the hex value to the right eliptic curve type. 

This helps people to quickly generate the list of supported curves by looking at the client hello.


drwetter:
Thank you!

PR #1799, opened 2020-12-09
Refactor `apk` usage in Dockerfile

PeterDaveHello:
Here are two changes:

1. Remove `--no-cache` for `apk` in Dockerfile
2. Clean up apk cache in Dockerfile after packages installed

These will make the build faster and the built image smaller.

The details are in the commit message.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM alpine:3.11
 
 RUN apk update && \
     apk upgrade && \
-    apk add --no-cache bash procps drill git coreutils libidn curl && \
+    apk add bash procps drill git coreutils libidn curl && \
     addgroup testssl && \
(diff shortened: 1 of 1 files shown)


drwetter:
Why do you prefer removing the apt-cache instead of **not** building it?

PeterDaveHello:
I'm not sure I understand the question.

The `apk update` will prepare the index, for `apk upgrade` and `apk add`, after those operations, it can and should be removed to minimize the image size, even if anyone uses this image will need more apk operations, which seems unlikely for most cases, the index cache could be expired/invalid and needed to be refreshed, there is no need to keep the cache in it.

drwetter:
sorry, I just looked at the changed lines. 

Thanks!

Issue #1805, opened 2020-12-15
[BUG] -iL / --file doesn't output to outputfile when specified in (input) file 


drwetter:
**Command line / docker command to reproduce**
```
prompt> cat f
--jsonfile example.com_p443_app1_context1.json https://example.com/app1/context2.json
--jsonfile example.com_p443_app1_context2.json https://example.com/app1/context2.json
--jsonfile example.com_p443_appX_contextZ.json https://example.com/appX/contextZ.json
prompt> ./testssl.sh --file f
[ .. echos neatly  each command line and executes it ..]
prompt> ls -l example.com_p443_app?_context?.json
zsh: no matches found: example.com_p443_app?_context?.json
prompt>
```

**Expected behavior**
The files should have been created.

drwetter:
Independent whether it's being done in parallel or serial mode. See also   #1804

drwetter:
Please check the latest and greatest (see PR merged above) and let us know if the problem persists.

PR #1806, opened 2020-12-16
Fix file creation when output file is in input file

drwetter:
see #1805.

When an input file ifor mass testing contains an output directive to a file the files weren't created. This fixes that for CVS, JSON and HTML. It could need more testing for HTML and CSV and the bug needs to be verified for plain log output.


drwetter:
FYI: There's the logic for the log output missing which need to be fixed. And there was another concern which I need to reproduce.

Issue #1770, opened 2020-11-06
FEATURE REQUEST - Provide custom request headers 


chrisdlangton:
Many endpoints are more commonly relying on a customer header to establish a TLS connection, without the header the connection will be dropped or 403 http code response sent.

I understand basicauth via Authorization header can already be added, this requested capability would do the same but not use a hard coded header name.

drwetter:
How does that work? HTTP header comes in the TLS packet.

pnieuwkamp:
I think Chris may not mean the actual establishment of the TLS connection, but the request that gets send after the connection is made. Some web services rely on a different header than the Authorization one to establish authorization. For instance, I have one that requires me to insert a HTTP header like: "ClientID: 0xDEADBEAF" otherwise it will not process the request.

chrisdlangton:
correct, the response headers will not be sent without the correct/expected request headers

drwetter:
ok, that's no big deal

drwetter:
See PR merged above

chrisdlangton:
Legend @drwetter 
Bash master smashes another feature

chrisdlangton:
Legend @drwetter this'll be a huge benefit for so many environments
Bash master @tosticated smashes another feature

PR #1807, opened 2020-12-22
Custom HTTP request headers support added. Addresses #1770

tosticated:
PR to address feature request #1770. 

- The _--customhttpheader_ parameter is supported and can be added multiple times if required. 
- A join_by() function (found at https://stackoverflow.com/questions/1527049/how-can-i-join-elements-of-an-array-in-bash/17841619#17841619) is added as a helper function.
- Documentation updated as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@
 * Added --overwrite argument to support overwriting output files without warning
 * Headerflag X-XSS-Protection is now labeled as INFO
 * Client simulation runs in wide mode which is even better readable
-* Added --customhttpheader to support custom headers in HTTP requests
+* Added --reqheader to support custom headers in HTTP requests
 
 ### Features implemented / improvements in 3.0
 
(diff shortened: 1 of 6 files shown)


drwetter:
Thanks! basically fine but to be honest both the cmd line and the global vars CUSTOMHTTPHEADER(S) got a bit too long for my taste. I'll chew on this during there days.

tosticated:
I agree, the variable is a bit long. Perhaps something like 'reqheader'? Let me know if I need to make any changes.

drwetter:
>  agree, the variable is a bit long. Perhaps something like 'reqheader'?

as I woke up today I was exactly thinking about that.

>  Let me know if I need to make any changes.

Yes, please. That would be great. Thanks for the Changelog too.

Then I can stash your branch I checked out.

Cheers, Dirk

tosticated:
Alright, changed it to _reqheader_ everywhere. Should be ok to merge.

drwetter:
Thank you!

Issue #1248, opened 2019-04-26
HSTS Browser Preloading Check ( Like SSL Labs )


Verequies:
Would be awesome if we could add an API check to see if a site is in preloaded in browsers.

API is available here: https://hstspreload.org
An example call would be: https://hstspreload.org/api/v2/status?domain=example.com

Which should return:
`
{
  "name": "example.com",
  "status": "preloaded",
  "bulk": true
}
`

The 'status' entry can have values "unknown" which most we can take as not preloaded, "pending" meaning its being processed and "preloaded" which obviously means its preloaded.

As for the 'bulk' entry, on my own domain, it was false for a while, and doing an SSL Labs check it returned as only in Chrome. However today I checked and the 'bulk' entry showed true, and SSL Labs shows its preloaded into Firefox, IE and Edge as well as Chrome. So I'm willing to bet thats what 'bulk' means.

This shouldn't be a difficult addition to the great test that is testssl :)

drwetter:
Thanks @Verequies for the suggestion / reminder. I thought it was already an issue here a long while back and I lost track but actually neither is the case. :-)

Likely a user check will need ``--phone-out`` though.

PR #1809, opened 2020-12-27
HSTS preload list check against API

tosticated:
First attempt at addressing #1248: adding support for HSTS preload list lookups.

 - Used the API example as provided in #1248. This seems to be the list that other lists (Firefox, Edge, etc.) are based on. A browser may still have an older copy of the list, but the state on hstspreload.org should propagate eventually.
 - The 'bulk' variable seems to be used for any entry that is added via the form on hstspreload.org, as opposed to any manually added domains such as mail.google.com for example (does not respond with preload header, but is in fact included in the list).
 - Of course only works when --phone-out is being set.

Feedback is welcome. Especially regarding the way the results are displayed. The idea is that the check is done even when the HSTS preload header is missing. This would show unintended and/or undesired situations such as having a list addition rejected because the 'preload' header itself is missing. Based on my understanding of how the list should work, I have currently assigned the following:

hsts_implementation_1
Explanation of some columns:
 - _has preload header_: indicates the response of the server that testssl.sh is testing
 - _preloadedDomain == name_: True if testssl.sh is currently testing the 'responsible domain'. False if it is a subdomain. A subdomain may not have 'preloaded' returned, but still can have a status like 'preloaded' according to the API.
 - _bulk_: if the API 'bulk' variable was true or false. When irrelevant for testssl.sh severity: "N/A".

As I was writing this check, it grew from just a simple 'status' to this entire 'lookup table'. If this is too much, it can also be changed back to just viewing the status without any severity and advice.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -834,6 +834,14 @@ strip_trailing_space() {
      printf "%s" "${1%"${1##*[![:space:]]}"}"
 }
 
+# removes first and last char, if both match with arg2. Leave arg2 empty to remove any first and last char.
+strip_enclosed() {
+    local regexp="^$2.*$2\$"
+    local returnstring=$1
+    [[ "$returnstring" =~ $regexp ]] && returnstring=""${returnstring:1:${#returnstring}-2}""
+    echo $returnstring
+}
+
 is_number() {
      [[ "$1" =~ ^[1-9][0-9]*$ ]] && \
           return 0 || \
@@ -1991,6 +1999,77 @@ check_revocation_ocsp() {
      fi
 }
 
+# API checks against the hstspreload.org HSTS Preload list API.
+# arg1: domain to be checked
+# arg2: JSON key to be checked (status, preloadedDomain, name, bulk)
+# arg3: to compare the JSON value against (case insensitive)
+# Returnvalues:
+#      0 - Nothing checked, just made the API request
+#      1 - Unsuccessful API request because of connection errors
+#     10 - Match
+#     20 - No match because value did not match
+#     21 - No match because key not found in response
+check_hsts_preloadlist_match() {
+     local domain="$1"
+     local key="$2"
+     local value="$3"
+     local tmpfile=$TEMPDIR/$NODE.hsts-preloadlist.txt
+     local uri_api_status="https://hstspreload.org/api/v2/status?domain=$domain"
+
+     "$PHONE_OUT" || return 0
+
+     # Only make a new API request once per domain
+     if [[ ! -f "$tmpfile" ]]; then
+          http_get "$uri_api_status" "$tmpfile"
+          [[ $? -ne 0 ]] && return 1
+     fi
+
+     # If no key was provided, just get the API response, and return
+     [[ -z "$key" ]] && return 0
+
+     # Check if key exists in response. If not, the API may have changed or something.
+     ! grep -Fiaqw "$key" $tmpfile && debugme echo "HSTS preloadlist key unrecognized: $key" && return 21
+
+     # Check if there is a match, return 10 if there is, 20 if not
+     grep -Fiaqw "\"$key\": $value" $tmpfile && return 10 || return 20
+
+     # do not remove tmpfile because it can be used again to limit lookups against API
+}
+
+# Checks for known values of specific key in preloadlist API.
+# Depends on check_hsts_preloadlist_match()
+# arg1: domain to be checked
+# arg2: key to be checked (string): status
+# Returnvalues (string): unknown, pending, rejected, preloaded
+check_hsts_preloadlist_value() {
+     local domain="$1"
+     local key="$2"
+     local values
+     local value
+     local value_ret
+
+     [[ -z "$key" ]] && return 1
+
+     # Currently using preset values and testing for matches instead of echo-ing
+     # directly from the response. No input has to be trusted this way.
+     case $key in
+          "status") values=("\"unknown\"" "\"pending\"" "\"rejected\"" "\"preloaded\"") ;;
+          "bulk") values=("true" "false") ;;
+          *) return 1 ;; # No supported key provided.
+     esac
+
+     for value in ${values[@]}; do
+          check_hsts_preloadlist_match "$domain" "$key" "$value"
+          [[ $? -eq 10 ]] && value_ret="$value" && break
+     done
+
+     value_ret=$(strip_enclosed "$value_ret" "\"")
+
+     [[ ! -z "$value_ret" ]] && echo "$value_ret" && return 0
+
+     return 1
+}
+
 wait_kill(){
      local pid=$1             # pid we wait for or kill
      local maxsleep=$2        # how long we wait before killing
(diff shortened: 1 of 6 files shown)


drwetter:
Sorry, about the late reply. First: much appreciated, thanks!

 I read and tried that in December. I can't get to it and need a few days for a consolidate answer!

testssl.sh
```diff
@@ -1991,6 +1999,77 @@ check_revocation_ocsp() {
...
+     if [[ ! -f "$tmpfile" ]]; then
+          http_get "$uri_api_status" "$tmpfile"
+          [[ $? -ne 0 ]] && return 1
+     fi
+
+     # If no key was provided, just get the API response, and return
+     [[ -z "$key" ]] && return 0
+
+     # Check if key exists in response. If not, the API may have changed or something.
+     ! grep -Fiaqw "$key" $tmpfile && debugme echo "HSTS preloadlist key unrecognized: $key" && return 21
```
drwetter, line 2031: This grep is legacy code. Can't this be changed into something like ``[[ "$tmpfile" =~ \ ${key}\ ]]``   ? (please doublecheck)

testssl.sh
```diff
@@ -1991,6 +1999,77 @@ check_revocation_ocsp() {
...
+
+     # Check if there is a match, return 10 if there is, 20 if not
+     grep -Fiaqw "\"$key\": $value" $tmpfile && return 10 || return 20
```
drwetter, line 2034: See above wrt legacy

testssl.sh
```diff
@@ -2606,19 +2690,72 @@ run_hsts() {
...
 
+     # Check if domain can be found in the HSTS preload list via hstspreload.org API.
+     # Regardless of the HSTS preload header presence, run this part. It may be rejected because it does not contain the right header for example,
+     # or it may be still in the list after it has been removed.
+     if "$PHONE_OUT"; then
+          json_postfix="_preloadlist"
+          out "$indent"; pr_bold " HSTS preload list            "
+
+          # Check if the domain is also the preloadedDomain. If so, it *may* be correct that the server response header does not have 'preloaded' included.
+          check_hsts_preloadlist_match "$NODE" "preloadedDomain" "\"$NODE\""
```
drwetter, line 2714: Is there any reason you use here and other places string why you pass the double quotes here? If they are needed at all I would rather handle this in the function and not in the call.

testssl.sh
```diff
@@ -2606,19 +2690,72 @@ run_hsts() {
...
+          [[ $? -eq 10 ]] && preloadbulk=true || preloadbulk=false
+
+          # Fill combined variable for clear lookup. (preloadmarked = true, preloadsame = true, preloadbulk = true) -> "111"
+          [[ $preloadmarked == true ]] && preloadcombined="${preloadcombined}1" || preloadcombined="${preloadcombined}0"
+          [[ $preloadsame == true ]] && preloadcombined="${preloadcombined}1" || preloadcombined="${preloadcombined}0"
+          [[ $preloadbulk == true ]] && preloadcombined="${preloadcombined}1" || preloadcombined="${preloadcombined}0"
+          debugme echo "Temporary lookupvariable: $preloadcombined"
+
+          # Determine and show the outcome
+          case "$(check_hsts_preloadlist_value "$NODE" "status")" in
```
drwetter, line 2728: for a plain string there should be no need to enclose it in quotes

testssl.sh
```diff
@@ -1991,6 +1999,77 @@ check_revocation_ocsp() {
...
+     local value_ret
+
+     [[ -z "$key" ]] && return 1
+
+     # Currently using preset values and testing for matches instead of echo-ing
+     # directly from the response. No input has to be trusted this way.
+     case $key in
+          "status") values=("\"unknown\"" "\"pending\"" "\"rejected\"" "\"preloaded\"") ;;
+          "bulk") values=("true" "false") ;;
+          *) return 1 ;; # No supported key provided.
```
drwetter, line 2058: Please double check here the usage of quotes. I believe for ``status`` and ``bulk`` they can be removed

For the array values I am not sure whether they need contain quotes.

testssl.sh
```diff
@@ -1991,6 +1999,77 @@ check_revocation_ocsp() {
...
+
+# Checks for known values of specific key in preloadlist API.
+# Depends on check_hsts_preloadlist_match()
+# arg1: domain to be checked
+# arg2: key to be checked (string): status
+# Returnvalues (string): unknown, pending, rejected, preloaded
+check_hsts_preloadlist_value() {
+     local domain="$1"
+     local key="$2"
+     local values
```
drwetter, line 2047: ``local -a value=()`` would declare the type better.

drwetter:
Good work! You spend a lot of time for the lookup table and the decision table you provided is quite impressive. 

The information on the screen for both easy cases (HSTS and no preload anywhere, HSTS and preload eerywhere)

image

I miss use case to test every variant of this. If I sleep over it I probably can test 1-2 more than the basic variants. If you could address the minor points in the code; I'll do a few more tests and then it should be set.

Thanks for your work!

potato-20:
@drwetter @tosticated — #1248 and this PR are still labeled `important`/`3.3dev`, so I'd like to help get it over the line.

@tosticated, thanks for the original implementation and the decision table — full credit to you; if you're not actively working on it, I'm happy to revive it as a fresh PR against `3.3dev` and keep your credit.

My plan: rebase onto current `3.3dev` and address the review comments here —
- replace the legacy `grep -Fiaqw` with `[[ ... =~ ... ]]`,
- move the JSON quoting *into* `check_hsts_preloadlist_match()` so callers pass plain `status`/`bulk`/domain values,
- use `local -a` for the value arrays.

To resolve what stalled this (verifying every variant of the preload decision table), I'd add a `t/` unit test modelled on `t/52_ocsp_revoked.t` that exercises the `--phone-out` lookup against a known-preloaded and a known-unlisted domain.

@drwetter — before I finalize: are you happy with the existing decision-table approach (`status` × header-marked / same-domain / bulk), or would you prefer a simpler `status`-only output? Happy to build to whichever you prefer.

drwetter:
@potato-20 : would be great! Allow me some time for feedback

drwetter:
@potato-20 

> Before I finalize: are you happy with the existing decision-table approach (status × header-marked / same-domain / bulk), or would you prefer a simpler status-only output? Happy to build to whichever you prefer.

It looked fine [like it was to me, see picture](https://github.com/testssl/testssl.sh/pull/1809#issuecomment-773613147) . 

And: unit tests are always appreciated ;-)

potato-20:
I've opened #3060 which revives this onto current `3.3dev` and addresses @drwetter's review comments here (pure-bash matching, quoting handled inside the helper, `local -a`, kept the decision table) plus adds a unit test (`t/53_hsts_preload.t`). Full credit to @tosticated for the original implementation — retained in CREDITS.md. This draft can probably be closed in favour of #3060.

drwetter:
Cool, thanks a lot! I'll get to it next week.

Issue #1073, opened 2018-06-21
Check for MTA STS


drwetter:
https://datatracker.ietf.org/doc/draft-ietf-uta-mta-sts/

drwetter:
has been published now https://www.rfc-editor.org/rfc/rfc8461.txt

PR #1815, opened 2021-01-05
[WIP] Add MTA-STS 

drwetter:
This PR adds an implementation of MTA-STS (RFC 8461), see also issue #1073

Currently it's a WIP.

What doesn't work
- <s>test a hostname which is not equal to domainname</s>
-  <s>test a hostname which has no mx record</s>
- <s>file output</s>
- <s>any</s> compliance parsing of TLS RPT TXT record + .well-known/mta-sts.txt
- <s>when no TXT records or .well-known/mta-sts.txt are there</s>
- <s>colored screen output</s>

Needed:
 - More testing and errors to be addressed.
 - sub.sub.domain.tld probably doesn't work
 - --mx does a check for every single MX instead of keeping the results
 - Policy delegation (CNAME?) --> Not tested yet
 - A simple unit test would be great. Anybody?

There's a new stub function for DANE too.

Besides: to avoid confusion I changed all GET requests over HTTPS from ``tm_out`` to ``safe_echo``. It's actually exactly the same only the name is different but it seems more appropriate here.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7515,7 +7515,7 @@ sub_mta_sts() {
           pr_svrty_low "invalid"
           # Append to every element a string, here: '|'. n would work as well but we'd need $spaces.
           # Catch is also for the last, so maybe we should avoid arrays
-          # Hint is from http://web.archive.org/web/20101114051536/http://codesnippets.joyent.com/posts/show/1826
+          # Hint is from https://web.archive.org/web/20101114051536/http://codesnippets.joyent.com/posts/show/1826
           failreason_mtasts_rec=( "${failreason_mtasts_rec[@]/%/ | }" )
           fileout "${jsonID}_txtrecord" "LOW" "invalid _mta-sts TXT record $mta_sts_record $(printf '%s | ' "{failreason_mtasts_rec[@]}")"
           outln " _mta-sts TXT record '${mta_sts_record}'\n${spaces}   $(printf '%s' "${failreason_mtasts_rec[@]}")"
--- a/testssl.sh
+++ b/testssl.sh
@@ -7421,6 +7421,7 @@ sub_mta_sts() {
 
      if [[ -z "$mta_sts_record" ]]; then
           failreason_mtasts_rec="no record"
+          mta_sts_record_ok=false
      else
           if [[ $(count_char_occurence "$mta_sts_record" ';') -ne 2 ]]; then
                failreason_mtasts_rec+=("number of ; should be 2")
@@ -7442,6 +7443,7 @@ sub_mta_sts() {
                fi
           fi
      fi
+set +x
 
      policy="$(safe_echo "GET /.well-known/mta-sts.txt HTTP/1.1\r\nHost: mta-sts.$domain\r\nUser-Agent: $useragent\r\nAccept-Encoding: identity\r\nAccept: text/*\r\nConnection: Close\r\n\r\n" | $OPENSSL s_client $(s_client_options "-quiet -ign_eof -connect mta-sts.$domain:443 $PROXY -servername mta-sts.$domain") 2>$ERRFILE)"
      # here also the openssl return val needs to be checked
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -7354,6 +7361,249 @@ tls_time() {
...
+     local spaces="$1"
+
+     # Not yet implemeted
+     return 0
+
+     pr_bold " DANE / DNSSEC                "
+
+     tlsa_record="$(get_tlsa_record _$PORT._tcp.$NODE)"
+     # parsing TLSA certificate usage, TLSA selector, TLSA matching type, hash
+     rrsig_record="$(get_rrsig_record $NODE)"
```
vdukhovni, line 7602: To determine whether DANE is securely implemented:

- Determine whether the destination domain's MX RRset is DNSSEC signed and validates.  Best done by inspecting the "AD" bit from the local (loopback) DNS resolver.  If the "AD" bit is on for ". IN SOA ?" and off for "root-servers.net. IN SOA ?", then your local resolver appears to support DNSSEC, and "example.com. IN MX ?" should indicate whether the domain is signed.
- For each MX host determine whether either its A or AAAA RRset is DNSSEC signed (if either is non-existent, NODATA, the denial of existence may be insecure due to NSEC3 opt-out).
- If the A or AAAA RRset of, say, "smtp.example." is obtained via CNAME expansion, and AD=0, the zone may still be signed, check for the AD bit of "smtp.example. IN CNAME ?".
- If any of the MX hosts are in unsigned zones, at best "partial DANE support" is possible, which is subject to active downgrade to the unsigned MX hosts, by blocking connections to the signed ones.
- For each MX host in a signed zone, check for TLSA RRs:
    -  first with the base domain set to the CNAME expansion, if AD=1 for the address records despite CNAME expansion.
    - next with the base domain set to the original MX host name if no CNAME was used, was insecure, or no TLSA RRs found there.
- To check for TLSA RRs prepend "_25._tcp." to the base domain and perform a lookup with type "TLSA".
- OpenSSL `s_client` can be used to perform a DANE-validated STARTTLS session with given TLSA RRDATA, see the manpage.
- Alternatively, the OpenSSL API documentation for DANE can be found at: <https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_dane_enable.html>
- The "posttls-finger" program included with the Postfix source distribution, and in many OS Postfix distribution can perform DANE-based verification, though the output would need to be parsed to determine verification success, there's no exit code that directly reflects DANE authentication status.
- See also https://github.com/baknu/DANE-for-SMTP/wiki/2.-Implementation-resources

drwetter:
Hi @vdukhovni ,

thanks for you input! 

The function above was more a stub thing to signal that some thing is coming. The plan is to check for TLSA records first (as DNSSEC is more a pain) and label it as experimental. Then work on the DNSSEC part. Probably not only for SMTP. 

For DANE/TLSA  I'll have a PoC which is not good enough to publish. I decided though not to use the bultin ``-dane*`` functions of newer openssl, for compatibility reasons.

For DNSSEC I have no clue yet how to that but I think I have good input now. The catch here is probably the resolver which needs to be DNSSEC aware (ad), or can you think of any other way to do a proper test?

vdukhovni:
> For DANE/TLSA I'll have a PoC which is not good enough to publish. I decided though not to use the bultin `-dane*` functions of newer openssl, for compatibility reasons.

I've seen many naive attempts to implement DANE-TLSA certificate chain validation that fail to implement all the required steps correctly.  OpenSSL 1.1.0 is by now broadly available. My advice would be to not build a new validator from the ground up. You can of course do as you see fit.

> For DNSSEC I have no clue yet how to that but I think I have good input now. The catch here is probably the resolver which needs to be DNSSEC aware (ad), or can you think of any other way to do a proper test?

Again, instead of reinventing a ground up DNSSEC implementation, keep track of evolving algorithms, keep track of a potentially changing root zone trust anchor, ... your simplest approach would in fact be to use a local validating resolver and to trust its "ad" bit. Otherwise use a library like libunbound that provides all the necessary support, but still depends on a valid trust-anchor, which requires something on the system that keeps that up to date.

For me the focus would be on the end goal, determining wether a given site has a working DANE deployment.  Implementing a DNSSEC validator and a DANE TLSA certificate chain validator would be left to the libraries that specialise in that sort of thing...

PR #1825, opened 2021-01-13
Travis CI didn't run. Trying to fix it (3.0)

drwetter:


teward:
I thought Travis was being retired Soon(TM)... shouldn't this move onto GH Actions for its tests instead at some point?

drwetter:
It's being moved to travis-ci.com . But to be honest I have no big clue yet and thought at some point I only need to change the hook or so. 

If you or anybody else can port the stuff to GH actions go ahead. 

Or of I am wrong with my assumption that this is easy speak up.

Issue #1828, opened 2021-01-19
possible BUG : heartbleed and CVE


ghost:
Hi,

Just a little bug (I guess)

When "heartbleed" is detected, the field "cve" contains a cwe in place of the "CVE-2014-0160" and the cwe is not mentionned.

there are an extract:

vulnerable:

```
{
   "id":"heartbleed",
   "ip":"x.x.x.x/x.x.x.x",
   "port":"443",
   "severity":"CRITICAL",
   "cve":"CWE-119",
   "finding":"VULNERABLE CVE-2014-0160"
}
```

not vulnerable, but wrong cve:

```
{
   "id":"heartbleed",
   "ip":"x.x.x.x/x.x.x.x",
   "port":"443",
   "severity":"OK",
   "cve":"CWE-119",
   "finding":"not vulnerable CVE-2014-0160"
}
```

not vulnerable (timeout) display ok:

```
{
   "id":"heartbleed",
   "ip":"x.x.x.x/x.x.x.x",
   "port":"443",
   "severity":"OK",
   "cve":"CVE-2014-0160",
   "cwe":"CWE-119",
   "finding":"not vulnerable , timed out"
}
```

Thierry

drwetter:
merci

drwetter:
fixed via PRs above

PR #1830, opened 2021-01-20
Fix file output formatting for heartbleed

drwetter:
Quotes were wrong for different results, which lead to some confusion
for finding, cve and cwe


drwetter:
fixes #1828

Issue #1148, opened 2018-10-25
`--file` mode combined w/ *file arguments produces partial output files


bitsofinfo:
Please make sure that you provide enough information so that we understand what your issue is about.

   
1. testssl version from the banner (testssl.sh -b 2>/dev/null | head -4 | tail -2)

```
testssl.sh       2.9.5-7 from https://testssl.sh/
    (727a288 2018-09-20 22:05:28 -- )
```

2. what exactly was happening, output is needed

Running this produces expected files w/ proper content, fully populated

```
./testssl.sh -S -P -p --fast --htmlfile test.html --csvfile test.csv --jsonfile-pretty test.json https://www.google.com
```

Running this produces files with partial data populated.  They are not complete JSON, invalid HTML etc

```
./testssl.sh --file test.txt
```

Contents of `test.txt`

```
-S -P -p --fast --htmlfile test.html --csvfile test.csv --jsonfile-pretty test.json https://www.google.com
```

3. what did you expect instead?

When running in `--file` mode, the contents of the `test.*` files should look no different then when running in non-file mode.

4. steps to reproduce

See above
 
    1. openssl version used (testssl.sh -b 2>/dev/null | head -16 | tail -3)
 
```
testssl.sh       2.9.5-7 from https://testssl.sh/
    (727a288 2018-09-20 22:05:28 -- )
```

    1. your operating system (uname -a)

`17.6.0 Darwin Kernel Version 17.6.0`

drwetter:
yeah, there's no check of args which appear double. Wasn't aware that it affects the output which shouldn't be the case but need to check myself.

bitsofinfo:
Double args? not sure I specified any unique argument more than once. I just have need to generate all 3-4 outputs per command in the instruction files (csv, json, html and log). Only when the instruction file is used do I see this behavior, but not when straight on the command line w/ just a single command and no instruction file. (thanks for the great tool BTW!)

dcooper16:
What you are seeing is the result of the way that testssl.sh was designed, and it would be very difficult to fix. My suggestion would be to only include options related to output files in the command line, and not include them in test.txt.

To provide a more technical explanation, here is what is happening. testssl.sh implements the `--file` option by calling itself (creating a child process) for each line in the file (test.txt). The options that are sent to the child process are created by combining the options specified on the command line with those listed in the file (test.txt). In your example, the child is called with:
```
[...]

PR #1837, opened 2021-01-26
Mass testing with CSV, HTML, JSON, and/or LOG file names in mass test…

dcooper16:
…ing file

See #1148 and #1805.

As noted in #1148, testssl.sh is not current designed to handle a mass testing file in which CSV, HTML, LOG, and/or JSON file names are provided in the mass testing file. If a child process receives a command line with one of the files, it assumes the same command-line option was provided to the parent so that the output of every test is being written to this one file. If this assumption is wrong, then either the file will not be created at all or it will be malformed since it will be missing header and/or footer information.

This PR partially addresses the problem by introducing new command-line arguments that are for internal use only. These command line arguments allow a child process to distinguish between a CSV, HTML, LOG, or JSON file that it is supposed to create itself versus one that is to be shared by all of the child processes.

There is one major limitation to this PR. The code for handle command-line arguments in the mass testing file is very simple and cannot handle whitespace characters, whether they are enclosed in quotes or are escaped. So, any file names included in the mass testing file cannot have whitespace characters.


drwetter:
thanks, David!

PR #1843, opened 2021-02-12
Simplify Dockerfile, use Alpine 3.13, speed up docker build

goncalor:
Hi Dirk and other contributors.

Just a few changes related with Docker. The extended commit messages explain my rationale behind the changes.

In principle nothing should break from these changes.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.git
+.github
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,8 @@
-FROM alpine:3.11
+FROM alpine:3.13
 
-RUN apk update && \
-    apk upgrade && \
-    apk add bash procps drill git coreutils libidn curl socat openssl && \
-    rm -rf /var/cache/apk/* && \
-    addgroup testssl && \
-    adduser -G testssl -g "testssl user"  -s /bin/bash -D testssl && \
-    ln -s /home/testssl/testssl.sh /usr/local/bin/ && \
-    mkdir -m 755 -p /home/testssl/etc /home/testssl/bin
+RUN apk add --no-cache bash procps drill git coreutils libidn curl socat openssl && \
+    adduser -s /bin/bash -D testssl && \
+    ln -s /home/testssl/testssl.sh /usr/local/bin/
 
 USER testssl
 WORKDIR /home/testssl/


polarathene (approved):
LGTM :+1:

drwetter:
* I am not opposed using 3.13 but what is the goal
* have you tested it thoroughly? "IN principle" is a statement which rather tells me "no".
* _apk update/upgrade is not currently needed prior to apk add_ It also ensures that the content of the image is updated rather than taking just the image.

goncalor:
- No specific goal with using 3.13 other than having things up-to-date
- I did basic tests for a few sites. I did not realise there is a full set of tests. Now I have run them. I've found a problem but unrelated with my changes. Details below
- Fair enough. We can add `apk upgrade --no-cache`

Test `21_baseline_starttls.t` hangs at the XMPP test for `jabber.org:5222`. Running the test manually shows that `testssl.sh` is waiting for user input due to OpenSSL not being able to connect. I've ran the same test with an image produced by the current Dockerfile and the result is the same.

```
$ ./testssl.sh -q --ip=one --color 0 -t xmpp jabber.org:5222

Start 2021-02-20 13:51:06        -->> 208.68.163.218:5222 (jabber.org) <<--

A record via:           supplied IP "208.68.163.218"
rDNS (208.68.163.218):  xmpp.org.
Your OpenSSL cannot connect to jabber.org:5222
The results might look ok but they could be nonsense. Really proceed ? ("yes" to continue) --> 
```
```
$ openssl s_client -connect jabber.org:5222
CONNECTED(00000003)
139932911487816:error:1408F10B:SSL routines:ssl3_get_record:wrong version number:ssl/record/ssl3_record.c:331:
---
no peer certificate available
---
---
SSL handshake has read 5 bytes and written 312 bytes
Verification: OK
---
New, (NONE), Cipher is (NONE)
No ALPN negotiated
Early data was not sent
Verify return code: 0 (ok)
```

Other than this, all other tests pass.

----

Below is the blueprint for the image I used to run the tests from. `goncalor/testssl` was built from the Dockerfile in this PR with `docker build -t goncalor/testssl .`

```
# test.Dockerfile
FROM goncalor/testssl

USER root

RUN rm -rf bin etc testssl.sh
RUN apk add --no-cache wget make perl perl-app-cpanminus perl-test-harness-utils
RUN cpanm --notest Test::More Data::Dumper JSON Text::Diff

USER testssl
RUN git clone --depth 1 https://github.com/drwetter/testssl.sh.git .
```

drwetter:
* If 3.13 works: Ok, then we should use it, see https://github.com/drwetter/testssl.sh/pull/1722. For some reason there's a general, unrelated problem as you pointed out *thanks). See #1891
*  addgroup is missing
* As said I would like to ensure the container built has updates from Alpine repo and not just pulling the image. That is the reason why ``apk update && apk upgrade`` is there. If there's a some kind of proof that they are always up to date we can skip those commands. 
* The current Dockerfile has an mkdir command for etc and bin. IIRC (not sure though any more) it was needed.

Those points and given the fact that we're @ 3.12 already I am closing this.

drwetter:
FYI, necessity of updates:

```
 ---> Running in 6dc74afebd01
fetch http://dl-cdn.alpinelinux.org/alpine/v3.11/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.11/community/x86_64/APKINDEX.tar.gz
v3.11.11-28-g21152dff39 [http://dl-cdn.alpinelinux.org/alpine/v3.11/main]
v3.11.11-24-g127ddb5cee [http://dl-cdn.alpinelinux.org/alpine/v3.11/community]
OK: 11282 distinct packages available
Upgrading critical system libraries and apk-tools:
(1/1) Upgrading apk-tools (2.10.5-r0 -> 2.10.6-r0)
Executing busybox-1.31.1-r9.trigger
Continuing the upgrade transaction with new apk-tools:
(1/7) Upgrading musl (1.1.24-r2 -> 1.1.24-r3)
(2/7) Upgrading busybox (1.31.1-r9 -> 1.31.1-r10)
Executing busybox-1.31.1-r10.post-upgrade
(3/7) Upgrading libcrypto1.1 (1.1.1g-r0 -> 1.1.1k-r0)
(4/7) Upgrading libssl1.1 (1.1.1g-r0 -> 1.1.1k-r0)
(5/7) Upgrading ca-certificates-cacert (20191127-r1 -> 20191127-r2)
(6/7) Upgrading ssl_client (1.31.1-r9 -> 1.31.1-r10)
(7/7) Upgrading musl-utils (1.1.24-r2 -> 1.1.24-r3)
```

PR #1847, opened 2021-02-25
IANA <-> Openssl mapping issue

elfranne:
Seems Camellia got its last digit cut off.


drwetter:
Thanks!

Mind to issue the same PR to the 3.0 branch?

PR #1851, opened 2021-03-11
Added GitHub Action to External/related projects

mbogh:
Hi,

I made a GitHub Action to make testssl.sh easy to use in a workflow and added a link to README.md

Thanks for this awesome project 💯


drwetter:
Thanks! 

For github actions n00bs like me: Is that self-explanatory / good enough explained?

mbogh:
I hope, otherwise  I hope people will raise issues, so it can be improved.

PR #1853, opened 2021-03-16
Cleanup conversion from ASCII-HEX to binary

dcooper16:
This PR makes some minor improvements to code that converts ASCII-HEX to binary.

First, testssl.sh currently has two functions that do the same thing: `asciihex_to_binary()` and `hex2ascii()`. This PR converts all calls to either of these functions with calls to `hex2binary()`, which is based on the current `asciihex_to_binary()`.

This commit also changes direct use of `printf` to calls to `hex2binary()` in `generate-ccm-counter-blocks()`, `ccm-compute-tag()`, and `generate_gcm_counter()`.

Finally, this commit cleans up the implementation of `hex2binary()` a bit and introduces the option to use `xxd`, if that program is available. Using `xxd` rather than multiple calls to `printf` has a couple of advantages. `xxd` is a bit faster. However, the primary advantage is that when debugging (i.e., using `bash -x`), each call to `hex2binary()` only includes a few steps, regardless of how long the string is, whereas using `printf` the call to `hex2binary()` could take hundreds of steps.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -357,6 +357,7 @@ HAS_IDN=false
 HAS_IDN2=false
 HAS_AVAHIRESOLVE=false
 HAS_DIG_NOIDNOUT=false
+HAS_XXD=false
 
 OSSL_CIPHERS_S=""
 PORT=443                                # unless otherwise auto-determined, see below
@@ -782,15 +783,6 @@ debugme1() { [[ "$DEBUG" -ge 2 ]] && "$@"; }
 
 hex2dec() {
      echo $((16#$1))
-}
-
-# convert 414243 into ABC
-hex2ascii() {
-          for (( i=0; i<${#1}; i+=2 )); do
-               # 2>/dev/null added because 'warning: command substitution: ignored null byte in input'
-               # --> didn't help though
-               printf "\x${1:$i:2}" 2>/dev/null
-          done
 }
 
 # convert decimal number < 256 to hex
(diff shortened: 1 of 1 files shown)


drwetter:
Hi David,

good idea,thanks!  One remark though: why ``hex2binary()``? I find the binary part a bit misleading.

(for Travis failure, see #1855 . Don't mind because of this PR)

Cheers, Dirk

dcooper16:
> why `hex2binary()`?

In most cases the function is being called with data such as certificates, CRLs, OCSP responses, public keys, or ciphertext. There are only a few places, such as where `hex2binary()` is replacing `hex2ascii()` in which the data being converted represents text (i.e., ASCII).

Do you have any suggestions for a different name?

drwetter:
Good point. :)  Maybe ``hex2x`` adding a comment that the trailing x can be either ASCII or binary?

dcooper16:
Hi Dirk,

I don't think that `hex2x()` would be any less confusing, as people will probably try to guess what that means before going to the function definition and reading the comment.

How about just having two functions, `hex2binary()` and `hex2ascii()`. I've updated the PR so that the previous calls to `hex2ascii()` remain unchanged, but `hex2ascii()` is now just defined as a call to `hex2binary()`. I also added more text to the comment at the beginning of `hex2binary()`.

drwetter:
Sounds reasonable, thanks!

One last thing maybe: Your xxd shortcut sounded like an excellent idea to me! Do you mind to (re)add that to ``hex2ascii?()`` ? The function is not much in use outside debug statements. But if the usage changes ... and for the sake providing good code.

I can take that on my list as well but I'd need a bit

dcooper16:
> One last thing maybe: Your xxd shortcut sounded like an excellent idea to me! Do you mind to (re)add that to `hex2ascii?()`? 

Hi Dirk,

In the PR `hex2ascii()` is just defined as:
```
# convert 414243 into ABC
hex2ascii() {
     hex2binary $1
}
```
This adds the overhead of an extra function call, but I figured it was okay since `hex2ascii()` is rarely used.

drwetter:
Thanks for pointing me to the correct spot  ;-/

PR #1857, opened 2021-03-25
add codepoints for RFC7905, ChaCha20-Poly1305

jschauma:
https://tools.ietf.org/html/rfc7905 has the code points for ChaCha20-Poly1305, which testssl.sh does verify/report, but the conversion table doesn't currently include.

This PR adds these entries.


drwetter:
Thank you!

Care to do a PR for the 3.0 branch?

jschauma:
Sure thing: https://github.com/drwetter/testssl.sh/pull/1858

PR #1861, opened 2021-03-31
Use locally scoped counter in hex2binary

vin01:
In [the commit 3756cdc](https://github.com/drwetter/testssl.sh/commit/3756cdcb38e95e7f31f90fbd53e95d020faae55a), `xxd` was introduced and works as intended. However, in absence of `xxd` like in the official docker image for `testssl`, the test never finishes and remains tuck in TLS handshake at:

```
 Testing robust forward secrecy (FS) -- omitting Null Authentication/Encryption, 3DES, RC4
  Elliptic curves offered:     prime256v1 secp384r1 secp521r1 X25519 X448
  ...
  ```
  
  It looks like the local scope definition was removed unintentionally, using a locally scoped counter works fine.
  
  Please take a look!

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2066,7 +2066,7 @@ fi
 # key) or a text string (e.g., ASCII-encoded text).
 hex2binary() {
      local s="$1"
-     local -i len remainder
+     local -i i len remainder
 
      len=${#s}
      [[ $len%2 -ne 0 ]] && return 1


dcooper16:
Thanks!  Removing the local declaration if `i` was definitely unintentional.

drwetter:
Thanks!

@vin01 : if you don't mind you can have a look the Dockerfile and add xxd -- supposed it has the same functionality, see e.g @dcooper16's smart check 

`` if type -p xxd &> /dev/null && [[ "$(xxd -r -p <<< "30313233343536373839" 2>/dev/null)" == 0123456789 ]]; then``

PR #1862, opened 2021-04-01
Add xxd to docker image

vin01:
xxd is not a strict requirement and a fallback logic exists to handle its absence. However it is in general more performant and helpful in debugging as well which is desirable in many situations like ci/cd pipelines. See commit 3756cdc for details


drwetter:
Thanks!

PR #1867, opened 2021-04-06
Literal HT character issue in JSON output

multiflexi:
Fixes issue with values which include literal horizontal tabulation (HT). HT cannot be in JSON and scan of e.g. `rr.cesnet.cz` produces an invalid JSON file, because of HTs in header value. New function `sanitize_json_string()` substitutes HT with \t. There might be other characters that should be escaped as "[https://stackoverflow.com/a/11495576](https://stackoverflow.com/a/11495576)" suggests.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1170,10 +1170,18 @@ fileout_section_footer() {
      SECTION_FOOTER_NEEDED=false
 }
 
+sanitize_json_string() {
+     local str="$1"
+     str=${str//	/\\t} # \t (tab)
+
+     echo -n "$str"
+}
+
+
 fileout_json_print_parameter() {
      local parameter="$1"
      local filler="$2"
-     local value="$3"
+     local value=$(sanitize_json_string "$3")
      local not_last="$4"
      local spaces=""
 


drwetter:
Thanks, @multiflexi . 

Quick shot: Can reproduce it, thanks and I believe it's a good idea to sanitize this.
 
However the problem with ``\t`` appears way earlier (input problem) and I am not sure it is the right approach fixing it at one output.

Then I have to admit I do not like the idea of invisible control codes. Also this one will be gone if somebody (including myself) is using ``:retab`` in vim.

I haven't really made my mind up  but maybe a general sanitizing function for HTTP headers would do better.

Comments appreciated.

multiflexi:
I beg to differ. I think input should not be sanitized (other than what would produce BASH issues) and should be kept true to the source. Sanitization should be then separate for each output format considering its specifics (JSON, CSV...).
The issue with invisible control codes can be fixed with something this: `${str//$'\t'/\\t}`

drwetter:
As a general rule of thumb: smuggling invalid characters into the input can have undesired effects in the processing or the output. 

Output can be the command line, or worse: some people have developed web interfaces. The "undesired effect" can be a security issue or just a display issue. 

It is quite tricky however if, where, when to sanitize/validate, as one needs to know the output channel. E.g. a ``<script>alert(1)</script>`` is harmless for JSON, CSV but for HTML it is at least undesired. OTOH a an excel command (google for CSV or excel injection) maybe harmful for CSV. Shell commands maybe harmful under some circumstances for *.log files or when just running testssl.sh .

I stumbled a longer while ago on header payloads I prepared on dev.testssl.sh and fixed those and did some general work on this. I believe there needs to be a another round.

Thus I opened a new issue for that, closing this one.

Thanks for the input!

PR #1868, opened 2021-04-07

dcooper16:
This PR addresses a few issued related to the use of testssl.sh with OpenSSL 3.0.0-alpha14.

First, when `pkey` is used to print a DH key that uses an unknown group, OpenSSL 3.0.0-alpha14 labels the prime and generator using "P:" and "G:" rather than "prime:" and "generator:". (In PR #1586 it was noted that OpenSSL 3.0.0-alpha1 used "prime P:" and "generator G:". The `x509` command in OpenSSL 3.0.0-alpha14 still uses "prime P:" and "generator G:" when printing a DH public key in a certificate, but the `pkey` command just uses "P:" and "G:").

Second, when the `pkey` command is used to print a DH key that uses certain common primes (e.g., groups from RFC 3526 or RFC 7919), OpenSSL 3.0.0-alpha14 simply prints "GROUP: " followed by a short name for the group rather than printing the value of the prime and generator.

Finally, the "-text" option no longer works if the input is a public key. Fortunately, the "-text_pub" option provides the expected results with all versions of OpenSSL and LibreSSL.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -8760,7 +8760,7 @@ certificate_info() {
           "DH")  if [[ -s "$common_primes_file" ]]; then
                       cert_spki_info="${cert_txt##*Subject Public Key Info:}"
                       cert_spki_info="${cert_spki_info##*Public Key Algorithm:}"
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks !

PR #1872, opened 2021-04-11
functionnality proposal (and two bugfix)

ghost:
Hi,

I propose to stream live result in json format on specific file descriptor. This is useful for automating tests.

My proposal add one option `--stream-json-fd4`  which stream one object json per finding on the file descriptor 4 each time a new property is reported.

We can use like this:

```
./testssl.sh --stream-json-fd4 ... 4>pipe
```

or like this (send stdout and stderr on stderr and stream JSON object on stdout)

```
./testssl.sh --stream-json-fd4 ... 50>&1 1>&2 4>&50
```

in other way the pr contains two bugfix:

 - it seems that "finding" are no longer reported (not sure its a bug)
 - problem encoding json with some scan (if the value reported contains specific characters or non-ascci chars)

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1236,8 +1236,18 @@ fileout_json_finding() {
      local cwe="$5"
      local hint="$6"
 
+     local initial_do_pretty_json="$do_pretty_json"
+     local initial_do_json="$do_json"
+     local do_stream_mode="false"
+     if [[ "$do_stream_json" = "true" ]]; then
+          do_json="true"
+          do_stream_mode="true"
+     fi
+
      if "$do_json"; then
-          "$FIRST_FINDING" || echo -n ","
+          if [[ "$do_stream_mode" != "true" ]]; then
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks.

Have you tried e.g. ``testssl.sh --jsonfile=/dev/stderr testssl.sh >/dev/null & tail -f /dev/stderr``?

The char conversion only works with an UTF terminal setting. 

Wrt the ">> $JSONFILE" redirection: have you tried parallel mass testing -- into a (edit: each) single file and into the same file?

ghost:
the `--jsonfile=/dev/stderr` solution works perfectly for me.

I no tried mass testing, but you can forgot this proposal. An output on specific file description, is too tricky. Its a bad idea.

On a massive scan i does on January (about 25 000 scan) I encountered some encoding json cases. unfortunately, I no longer have the specific cases (result were urgent, i does manual fix). 

thanks for answer and thanks for the project.

PR #1877, opened 2021-04-23
Remove grade cap for STARTTLS

mimi89999:


mimi89999:
Why? That's nonsense? Many clients do require STARTTLS.

mimi89999:
Getting a `T` score for a perfectly well configured server makes it useless.

drwetter:
a) There was no description and b) as indicated I strongly believe it is good to remove this. Please read me comment in the issue you filed. This is just wrong

Issue #1879, opened 2021-04-30
Incorrect epoch time at HSTS test


McNobody:
The ssltest.sh 3.0 prints:

 Strict Transport Security    15465600 s = 179 days is too short ( >=15465600 seconds recommended), includeSubDomains

The recommended minimum is 180 days = 15552000 s.

drwetter:
Thanks. It' looked line an "off by one second" mistake in 3.0 branch only and would work if it would say `` > 15465600 seconds recommended`` like in 3.1 dev.

But I believe the recommendation should be ``>= 180 days`` (``>= 15552000 seconds``).

PR #1882, opened 2021-05-08
Fix "off by one" error in HSTS

drwetter:
There was by mistake a 179 days threshold and also the error message
was wrong when HSTS was exactly set to 179 days.

This commit sets it to 180 days and corrects the error messages on
both screen and JSON.


drwetter:
See #1879

Issue #1880, opened 2021-05-06
Finding text not consistent across output formats (console, JSON, CSV)


definity:
**Which version are you referring to**
3.1dev

When using `-p` to test protocols, the text used for the finding is not consistent across output formats.  It seems like "not offered" is the standard text, but I have found that for TLS 1.1 the text "is not offered" is used in the CSV and JSON outputs.  I'm not sure whether "not offered" or "is not offered" is supposed to be used, but I feel it shouldn't be mixed.  One or the other please 😅 

Sample Command: `./testssl.sh -oA test -p any_website_with_tls11_disabled`

**Console Output snippet**
```
 Testing protocols via sockets except NPN+ALPN 
 SSLv2      not offered (OK)
 SSLv3      not offered (OK)
 TLS 1      not offered
 TLS 1.1    not offered                                                        <-------- Here it's OK
 TLS 1.2    offered (OK)
 TLS 1.3    offered (OK): final
 NPN/SPDY   h2, http/1.1 (advertised)
 ALPN/HTTP2 h2, http/1.1 (offered)
 ```

**JSON Output snippet**
```
,         {
              "id"           : "TLS1",
              "ip"           : "some_website/1.2.3.4",
              "port"         : "443",
              "severity"     : "INFO",
              "finding"      : "not offered"
          }
,         {
              "id"           : "TLS1_1",
              "ip"           :  "some_website/1.2.3.4",
              "port"         : "443",
              "severity"     : "INFO",
              "finding"      : "is not offered"                                <-------- Here it's DIFFERENT
          }
```

**CSV Output snippet**
```csv
"TLS1_1","some_website/1.2.3.4","443","INFO","is not offered","",""            <-------- Here it's DIFFERENT
```

drwetter:
Thanks, you're right.

Care for a PR? 

```
prompt> grep -n 'is not offered' testssl.sh | grep -v set_grade_cap
5437:                    fileout "$jsonID" "INFO" "is not offered"    # neither good or bad
5448:                    fileout "$jsonID" "CRITICAL" "TLSv1.1 is not offered, and downgraded to a weaker protocol"
5473:               fileout "$jsonID" "INFO" "is not offered"
5498:     # running TLS 1.3 only when TLS 1.2 is not offered.  0 and 5 are the return codes for
prompt> 
```

Only L5437 + L5473

Cheers, Dirk

definity:
Yep, will try to get to this soonish :)

[...]

PR #1884, opened 2021-05-08
fixed minor bug with json output for TLS1.1

definity:
As requested in https://github.com/drwetter/testssl.sh/issues/1880

Have a great weekend!


drwetter:
Thanks!

Mind to change that for 3.0 too?

https://github.com/drwetter/testssl.sh/issues/1880#issuecomment-833625115

PR #1885, opened 2021-05-09
fixed minor bug with json output for TLS1.1

definity:
As requested in https://github.com/drwetter/testssl.sh/issues/1880


drwetter:
Thanks again!

PR #1890, opened 2021-05-13
Specify an error code in all calls to fatal()

dcooper16:
This PR fix a few places in which calls to `fatal()` do not specify an error code.


drwetter:
Thanks!

PR #1896, opened 2021-05-28
Fix compression detection variable syntax in run_breach()

PeterDaveHello:
The `$` symbol of the variable seems to be missing.

cc #1636


drwetter:
Thanks!

PR #1898, opened 2021-05-29
Fix indentation in testssl.sh

PeterDaveHello:
Just find out that the indentation is 5 spaces here, pretty special 😄


drwetter:
Thanks!

PR #1899, opened 2021-05-29
Correct "GitHub" case as it should be

PeterDaveHello:
It should make GitHub happy 😄


drwetter:
Okayyyy.. whatever ;-)

PR #1900, opened 2021-05-31
Make Shell Scripts' Shebang more consistent and portable

PeterDaveHello:
Some of them were `#!/bin/bash`, for the portability and consistency, they should be `#!/usr/bin/env bash` just like the script `testssl.sh` itself, shouldn't they 😄


drwetter:
Yes, thx

Issue #1901, opened 2021-05-31
Use `EditorConfig` to maintain the consistency of some styling?


PeterDaveHello:
Looks like there are some modelines in the shell/perl scripts, but they are not very consistent:

- `vim:tw=95:ts=5:sw=5:et` (`t/61_diff_testsslsh.t`)
- `vim:ts=5:sw=5:expandtab` (`testssl.sh`)
- `vim:ts=5:sw=5:expandtab` (`utils/00_unittest_baseline.sh`)
- `vim:tw=100:ts=5:sw=5:expandtab` (`utils/ccs-injection.bash`)
- `vim:ts=5:sw=5` (`utils/checkcert.sh`)
- `vim:ts=5:sw=5:expandtab` (`utils/create_ca_hashes.sh`)
- `vim:ts=5:sw=5:expandtab` (`utils/curves.bash`)
- `vim:ts=5:sw=5:expandtab` (`utils/generate_static_cipher_lists.sh`)
- `vim:ts=5:sw=5` (`utils/gmap2testssl.sh`)
- `vim:tw=200:ts=5:sw=5:expandtab` (`utils/heartbleed.bash`)
- `vim:ts=5:sw=5` (`utils/make-openssl.sh`)
- `vim:tw=90:ts=5:sw=5` (`utils/make-openssl111.sh`)
- `vim:tw=110:ts=5:sw=5` (`utils/prototype.ssl2proto-check.bash`)
- `vim:tw=110:ts=5:sw=5` (`utils/prototype.tls-protocol-checker.bash`)
- `vim:ts=5:sw=5` (`utils/ticketbleed.bash`)

Some of the shell scripts and most of the perl scripts do not have it:

- `utils/docker-debian10.tls13only.start.sh`
- `utils/docker-nginx.tls13-earlydata.start.sh`
- `utils/hexstream2cipher.sh`
- `utils/hexstream2curves.sh`
- `utils/resume.sh`
- `t/00_testssl_help.t`
- `t/01_testssl_banner.t`
- `t/02_clientsim_txt_parsable.t`
- `t/05_ca_hashes_up_to_date.t`
- `t/10_baseline_ipv4_http.t`
- `t/21_baseline_starttls.t`
- `t/23_client_simulation.t`
- `t/31_isJSON_valid.t`
- `t/32_isHTML_valid.t`
- `t/33_isJSON_severitylevel_valid.t`
- `t/51_badssl.com.t`

[...]

PR #1903, opened 2021-06-01
Add missing vim modeline config in sh & perl files

PeterDaveHello:
cc #1901


drwetter:
That wouldn;t be needed but it's welcome, thanks!

PR #1909, opened 2021-06-06
Add GitHub Actions testing

mikelolasagasti:
This should do the same test available for Travis in GitHub Actions.

    All tests successful.
    Files=12, Tests=71, 2523 wallclock secs ( 0.05 usr  0.01 sys + 469.64 cusr 273.70 csys = 743.40 CPU)
    Result: PASS

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,26 @@
+name: testssl.sh CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


drwetter:
Ah, ok, it is as simple as that.

Great, thank you!

Mind to do the same for 3.0?

mikelolasagasti:
> Mind to do the same for 3.0?

I should be able later today

drwetter:
> I should be able later today

Let me know if rather I should be doing this.

PR #1912, opened 2021-06-15
Make Shell Scripts' Shebang more consistent

PeterDaveHello:
Consider most of the scripts use bash in the project, should maybe just
use it, instead of /bin/sh in all the scripts.


drwetter:
Not sure why GHA failed here and your following PR was fine (both aren't related to changes in the core). Time will tell.

Thx!

PR #1913, opened 2021-06-15
Correct `exit` using, the exit status should be 0~255

PeterDaveHello:
According to the POSIX Programmer's Manual, the exit status specified by
the unsigned decimal integer. If n is specified, but its value is not
between 0 and 255 inclusively, the exit status is undefined.

By cross reference the usage between different scripts in this project,
it looks like we could simply remove the `-` before the number.


drwetter:
Okay :-)

thx!

Issue #1920, opened 2021-06-18
[Feature request] Fine tuning github actions


drwetter:
**Which version are you referring to**
3.0.x and 3.1dev

**Describe your feature request (if it's a technical feature)**

- The current setup of GHA performs checks with tow perl versions. One should be enough
- ubuntu:latest might not be a good idea. 20.04 or Debian 10 seems better for the time being
- our travis hook had a check for non-code  changes which didn't run CI checks (code below)

```
before_install:
  - |
    echo "Checking if a CI run is needed post commit: ${TRAVIS_COMMIT_RANGE}"
    if ! git diff --name-only ${TRAVIS_COMMIT_RANGE} | grep -qvE '(\.md$)|(\.pem$)|(\.pdf$)|(\.html$)|^(LICENSE)|^(docs)|^(utils)|^(bin)|(Dockerfile)'
    then
      echo "No code was updated, not running the CI."
      exit
    fi
```

PR #1921, opened 2021-06-19
Github actions modifications

drwetter:
- define deny list of files when modified not to run GHA
- specitfy OS to be ubuntu-20.04 (is there a debian at all?)
- only use perl 5.26

See #1920


drwetter:
Shouldn't this be running?

PR #1940, opened 2021-07-29
Print consecutive spaces faster

dcooper16:
This PR adds a new function, `print_n_spaces()`, which prints a sequence of (up to 80) space characters.

This new function is used to replace a few places in testssl.sh in which a sequence of space characters is printed by calling 
`out " "` in a loop. The new function is much faster than the current code, so it will make testssl.sh run slightly faster.


drwetter:
Thanks, David!

Issue #1939, opened 2021-07-29
Future of docker auto builds, continue to use dockerhub?


drwetter:
In the past whenever something was merged into 3.1dev or 3.0 a new container was build automatically containing the latest commit.

That is not anymore the case now, unless I pay for it:

![image](https://user-images.githubusercontent.com/8036727/127483649-12a6ff60-ec82-4110-8336-4e39997b7d7b.png)

Is there an alternative?

jauderho:
You should be able to build using GitHub Actions and push to Docker Hub & GHCR. You can see an example here based on a release tag but will work the same way for based on push.

https://github.com/jauderho/dockerfiles/blob/main/.github/workflows/ansible.yml

drwetter:
Thanks, @jauderho . What bugs me is that the credentials need to be store in clear text

drwetter:
I am not even able to build manually with the webgui.

On the command line I have two problems

``` 
---> Running in 8564f3e503bd
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/main/x86_64/APKINDEX.tar.gz
ERROR: http://dl-cdn.alpinelinux.org/alpine/v3.12/main: DNS lookup error
WARNING: Ignoring http://dl-cdn.alpinelinux.org/alpine/v3.12/main: No such file or directory
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/community/x86_64/APKINDEX.tar.gz
ERROR: http://dl-cdn.alpinelinux.org/alpine/v3.12/community: DNS lookup error
WARNING: Ignoring http://dl-cdn.alpinelinux.org/alpine/v3.12/community: No such file or directory
```

On a server: ``docker login`` doesn't work as it requires X or some binaries to be installed which have X-dependencies

drwetter:
The DNS error was due to a proxy. I believe the build command had a problem with that. Once I removed the proxy statement in config.json it worked.

So I just pushed the recent images manually until we find a proper solution

jauderho:
> 
> 
> Thanks, @jauderho . What bugs me is that the credentials need to be store in clear text

I store the Docker credentials as a GitHub secret so I suppose it's not perfect but it's workable for now. The proper solution would be for Docker to provide proper API token capability. 

If this is a big concern, you can choose to publish only to GHCR which correctly uses just a token. 

---

[...]

PR #1953, opened 2021-07-31
Add Dependenbot checking for GitHub Actions

jauderho:


drwetter:
Is this related to #1939 ? You lost me

jauderho:
Not directly. It looks like you already have a GitHub Action in use so enabling Dependabot should help keep Action versions up to date.

drwetter:
Thanks, @jauderho . You got to start picking me up where I am. :-) The only knowledge I have about dependabot is for other projects to warn about outdated binaries/libraries . 

The action version: isn't that maintained by GH ?

jauderho:
@drwetter 

Each GH action is typically bound to a version for each action. See for example https://github.com/jauderho/dockerfiles/blob/a5f08c8e09241d8082f05fb77591af730f0d4536/.github/workflows/ansible.yml#L30

What dependabot can do is monitor the action for an updated version of the action and then submit a pull request for you to modify your "workflow.yml" to reflect the latest version. 

Here's an example of such a pull request

https://github.com/jauderho/dockerfiles/pull/28

Anyways, it's not a must-have but certainly a useful feature IMO

drwetter:
Ok, thanks (merged).

PR #1958, opened 2021-08-04
Use all+ in calls to tls_sockets()

dcooper16:
`get_server_certificate()` includes a few calls to `tls_sockets()` in which the response will be TLS 1.3 and in which the response will be useless if it cannot be decrypted (since the goal is to obtain the server's certificate). So, these calls to `tls_sockets()` should specify "all+" rather than "all".


drwetter:
Thanks, David!

PR #1643, opened 2020-05-14
Support EdDSA

dcooper16:
This PR adds support for EdDSA (Ed25519 and Ed448). In particular:

* It modifies `prepare_tls_clienthello()` to include Ed25519 and Ed448 in the signature_algorithms extension of the TLS 1.2 and earlier ClientHello (RFC 8422).

* It modifies `run_server_defaults()` and `get_server_certificate()` to check whether the server offers EdDSA certificates with TLS 1.3.

* It modifies `certificate_info()` to handle certificates signed with EdDSA or with EdDSA public keys, even if $OPENSSL does not support pretty printing such keys and signatures.

* It modifies `read_sigalg_from_file()` to recognize EdDSA signatures even if $OPENSSL does not.

Note that I am not aware of any servers that have EdDSA certificates. This PR was created solely using Ed25519 and Ed448 certificates that I created myself and using a local, test Apache server. I has been tested to work with several versions of both OpenSSL and LibreSSL.

PR #1959, opened 2021-08-04
Use all+ in calls to tls_sockets() in 3.0

dcooper16:
`get_server_certificate()` includes two calls to `tls_sockets()` in which the response will be TLS 1.3 and in which the response will be useless if it cannot be decrypted (since the goal is to obtain the server's certificate). So, these calls to `tls_sockets()` should specify "all+" rather than "all".


drwetter:
Thanks David!

What was the reason again we don't have ``elif [[ "$1" =~ tls1_3_EdDSA ]]; then`` here, see #1958 ?

dcooper16:
> What was the reason again we don't have elif [[ "$1" =~ tls1_3_EdDSA ]]; then here, see #1958 ?

Hi Dirk,

Support for EdDSA is new in 3.1dev (#1643).

PR #1960, opened 2021-08-04
Check for RFC 8879 certificate compression

dcooper16:
This PR adds a check for whether the server supports certificate compression ([RFC 8879](https://www.rfc-editor.org/rfc/rfc8879.html)). If it does, then the list of supported compression methods is output in the server's preference order.

At the moment, facebook.com seems to support all three compression methods (zlib, brotli, Zstandard). https://github.com/openssl/openssl/pull/9155 can be compiled to support any or all of the compression methods, depending on the compile-time flags used.


drwetter:
Cool, looks good, thanks @dcooper16 !

PR #1962, opened 2021-08-05
Fix #1961

### LINKED ISSUE #1961: [BUG / possible BUG] XMPP scan of URI with numbers only fails to record IP
When 01337.io, 01337.ru or 616.pub is scanned, the IP in results is not 157.90.24.30 but the domain name. It happens when XMPP, STARTTLS is scanned, not for regular TLS scans.

testssl.sh -t xmpp -oA 0137a 01337.io:5222

**Expected behavior**
IP address 157.90.24.30 (for 01337.io)  as a value for key [ip].

[0137a.zip](https://github.com/drwetter/testssl.sh/files/6938364/0137a.zip)

 - OS: Ubuntu 20.04.2 LTS
 - Platform: Linux 5.4.0-80-generic x86_64
 - Version: testssl.sh       3.0.5 from https://testssl.sh/
 - Version if running from git repo: 91523be19494543f5c1bcf3dcab60498196c485d

dcooper16:
This PR fixes #1961 in the 3.1dev branch by leaving $NODEIP set to the server's IP address rather than changing it to the DNS name in the case of STARTTLS XMPP.

Currently, `determine_service()` sets `$NODEIP` to `$NODE` in the case of STARTTLS XMPP based on the following note:
```
# for XMPP, openssl has a problem using -connect $NODEIP:$PORT. thus we use -connect $NODE:$PORT instead!
```
Based on some limited testing, it seems that the issue is that the XMPP host name needs to be set. According the manual page for s_cilent:
> [The -xmpphost hostname] option, when used with "-starttls xmpp" or "-starttls xmpp-server", specifies the host for the "to" attribute of the stream element.  If this option is not specified, then the host specified with "-connect" will be used.

So, my guess is that if the `-xmpphost` isn't provided to $OPENSSL, then `-connect` needs to be provided a DNS name so that the "to" attribute can be set.

Currently, the `-xmpphost` option is only provided to $OPENSSL if testssl.sh's `--xmpphost` option is used. This PR changes `determine_service()` so that the `-xmpphost` option is always provided to $OPENSSL when STARTTLS XMPP is used, as long as $OPENSSL supports that option.

Note that my ability to test this PR was very limited, so please test it carefully before committing.

If the PR is considered acceptable, then I can also create a PR for the 3.0 branch.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -20983,8 +20983,6 @@ determine_service() {
                ftp|smtp|lmtp|pop3|imap|xmpp|xmpp-server|telnet|ldap|postgres|mysql|nntp)
                     STARTTLS="-starttls $protocol"
                     if [[ "$protocol" == xmpp ]] || [[ "$protocol" == xmpp-server ]]; then
-                         # for XMPP, openssl has a problem using -connect $NODEIP:$PORT. thus we use -connect $NODE:$PORT instead!
-                         NODEIP="$NODE"
                          if [[ -n "$XMPP_HOST" ]]; then
                               if ! "$HAS_XMPP"; then
                                    fatal "Your $OPENSSL does not support the \"-xmpphost\" option" $ERR_OSSLBIN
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks @dcooper16.

I'l lost a little bit the overview here. David's patch makes sense to me but it seems that the complaint in #1961 was different. And I remember several years back XMPP connections were a bit difficult to deal with.

Can somebody please test this thoroughly? Maybe first @multiflexi, @horazont and maybe folks who have a stronger background wrt XMPP like from the XMPP community?

horazont (approved):
This looks and sounds sensible to me, as long as openssl behaves correctly. Didn't have the chance to test that yet, though I guess a single smoke test should suffice.

drwetter:
A test would help as I vaguely remember from the very past that different server might react different.

 A small one has been done by GHA (OT: strange that for PRs fone needs to go into the forked branch of the owner and then into actions)

dcooper16:
> David's patch makes sense to me but it seems that the complaint in #1961 was different.

This should fix #1961 in most cases, since it leaves `$NODEIP` unchanged in most cases. The issue raised in #1961 will still happen if `$OPENSSL` does not support `-xmpphost`. Does #1961 need to be fixed for clients using versions of OpenSSL that do not support `-xmpphost`?

> And I remember several years back XMPP connections were a bit difficult to deal with.

I found the commit where this code was introduced, https://github.com/drwetter/testssl.sh/commit/179d8700d16c967020eda05a751bc56107cbbf66, but I couldn't find any information about the circumstances in which OpenSSL didn't work unless a DNS name was provided to the `-connect` command.

drwetter:
David, would you mind to backport this so that it is properly credited ?

PR #1968, opened 2021-08-09
Add GH Action to build Docker images

jauderho:
This only builds and pushes to GHCR at this time. As mentioned, you will need to create and add a Docker Hub PAT to your repo before I can update the action to push to Docker Hub too. 

Note: This only works for 3.1dev right now. I will create a separate pull request for 3.0


drwetter:
ok, thanks first @jauderho  for this and the other PR!

jauderho:
Let me know when you have added the Docker Hub username and PAT to your secrets. I can then add the lines necessary to push to Docker Hub. 

EDIT: Looks like the build is complete. You can find the images here : https://github.com/drwetter/testssl.sh/pkgs/container/testssl.sh

PR #1975, opened 2021-09-03
WIP: Minor cleanup

a1346054:


CHANGELOG.md
```diff
@@ -442,7 +438,7 @@ Full changelog @  https://github.com/drwetter/testssl.sh/commits/2.2/testssl.sh
 * local hostname in output
 
 1.31 (Halloween Release)
-* bugfix: SSLv2 was kind of borken
+* bugfix: SSLv2 was kind of broken
```
drwetter: this was deliberate

utils/00_unittest_baseline.sh
```diff
@@ -101,4 +101,3 @@ rm $FILE
 # -status
 
 # vim:ts=5:sw=5:expandtab
-
```
drwetter: Not sure I like those and the other empty line removals, especially in the end. That's just a matter of taste I guess

drwetter (changes requested):
Thanks for fixing all the minor details.

Don't know why the license text has (minor) differences. I just copied the text a while ago. The gnu.org server claims the license text is from 2010. No problem to use that one though.

What stops me from merging this: see the two comments. The line feeds might not feel right for you. in the end I rather prefer them. At other places: I can't tell maybe sometime 2x LFs were deliberate.

Care to change the LFs back?

utils/00_unittest_baseline.sh
a1346054: I originally removed them because some files didn't have them and some files did, so I thought the intention was to not have them.

CHANGELOG.md
a1346054: I should have realized :)

utils/00_unittest_baseline.sh
drwetter: ah, ok.

PR #1976, opened 2021-09-06
Fix dig -r check to only run if dig is available

ap-wtioit:
To suppress error messages like this: `/usr/local/bin/testssl.sh: line 20119: dig: command not found`

Info @wt-io-it


drwetter:
Oh, yes, thank you, @ap-wtioit !

Mind to port that back to 3.0?

ap-wtioit:
@drwetter thanks

backported to 3.0 in https://github.com/drwetter/testssl.sh/pull/1977

PR #1977, opened 2021-09-06
Fix dig -r check to only run if dig is available

ap-wtioit:
To suppress error messages like this: `/usr/local/bin/testssl.sh: line 20119: dig: command not found`

Backport of https://github.com/drwetter/testssl.sh/pull/1976

Info @wt-io-it


drwetter:
Thanks!

PR #1990, opened 2021-09-22
Improve check for client authentication

dcooper16:
As noted in #1709, some servers will only request client authentication if a specific URL is requested. This PR modifies the check for client authentication, in the case that a $URL_PATH is provided, by having testssl.sh perform a GET request on the URL provided on the command line.

I have not been able to find or set up a server that only requests client authentication if a specific URL is requested, so I have not been able to test this PR.

@lesterw1: Do you have a server that you can test this on or that you can point me to so that I can test it?

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -21095,7 +21095,18 @@ determine_optimal_proto() {
                   ( "$HAS_TLS13" && ! "$HAS_ENABLE_PHA" && ( [[ -z "$proto" ]] || [[ "$proto" == -tls1_3 ]] ) && [[ $(has_server_protocol "tls1_3") -ne 1 ]] ); then
                     $OPENSSL s_client $(s_client_options "$proto $BUGS -connect "$NODEIP:$PORT" -msg $PROXY $SNI") </dev/null >$TMPFILE 2>>$ERRFILE
                else
-                    safe_echo "$GET_REQ11" | $OPENSSL s_client $(s_client_options "$proto $BUGS -connect "$NODEIP:$PORT" -msg $PROXY $SNI -ign_eof -enable_pha") >$TMPFILE 2>>$ERRFILE
+                    safe_echo "$GET_REQ11" | $OPENSSL s_client $(s_client_options "$proto $BUGS -connect "$NODEIP:$PORT" -msg $PROXY $SNI -ign_eof -enable_pha") >$TMPFILE 2>>$ERRFILE &
+                    wait_kill $! $HEADER_MAXSLEEP
+                    if [[ $? -eq 0 ]]; then
+                         # Issue HTTP GET again as it properly finished within $HEADER_MAXSLEEP and didn't hang.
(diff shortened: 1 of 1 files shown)


drwetter:
Hi David,

is that still a WIP (see title?)

Thx, Dirk

dcooper16:
Hi Dirk,

I haven't tried to work on this in quite a while. All that remains to be done is for it to be tested so that there is some evidence that it actually works. I haven't been able to find a server to test against or to configure Apache to only offer/require client authentication for a particular directory, and it doesn't seem that anyone else has tested it either.

drwetter:
yes, almost forgot that, too.

@ Anybody listening? We look for a web server which has not a general client authentication but on a specific URL only. If you have such a thing or you're willing to set it up, please speak up or just used @dcooper16's branch ``improve_client_auth`` and provide feedback, please

drwetter:
Hi David,

if tested this with the following Apache config:

```
    SSLCACertificateFile /etc/ssl/client+ca.crt
    <Location /somewhere>
        SSLVerifyClient require
    </Location>
```

it returns in introducing section ``Service detected:       Couldn't determine what's running on port 443, assuming no HTTP service => skipping all HTTP checks`` if my cmd line is ``./testssl.sh https://<TAGET>somewhere``, in all other cases as usual ``HTTP``. This is independent on the openssl version (PHA support or not). It also claims (with OpenSSL 1.1.1x in the server defaults section) `` Client Authentication        none``

If I remove the location statements 

```
    SSLCACertificateFile /etc/ssl/client+ca.crt
        SSLVerifyClient require
```

both work as expected and as before.

drwetter:
thanks for your PR btw, and sorry for the wait

dcooper16:
Hi Dirk,

Thanks for the information. When I tried configuring Apache I was using `<Directory /somewhere>` rather than `<Location /somewhere>`. After changing "Directory" to "Location" I was able to do some testing and after trying a few different things found something that seems to work, although it needs to be tested some more.

There is definitely one problem with this PR. If `$OPENSSL` is a version of LibreSSL for which s_client supports TLS 1.3 and the server supports TLS 1.3, then the results will be incorrect. The problem is that LibreSSL's s_client does not seem to support post-handshake authentication, so there is no way to get the server to send a `CertificateRequest` message.

drwetter:
Hi David,

thanks for the feedback. Let me know if I can help you, like e.g. setting up a machine out there (=aws) for you. I did my tests on an in internal machine.

Thx, Dirk

dcooper16:
I've made some changes to this PR so that it works better when using testssl.sh with LibreSSL against a server that supports TLS 1.3. I also modified the code to try to protect against `$OPENSSL` hanging when sending a GET request through it (following the example from `run_http_header()`), just to be safe.

So, at this point I think this PR is ready to be merged,

drwetter:
Works for me now and looks great.

Thanks a lot, David!

PR #1991, opened 2021-09-24
fix html output content type header

klaernie:
The content type given was swapped around, I just copied that into a wrapper and noticed that my browser wanted to download the output file instead of displaying it.


drwetter:
Thanks!

PR #1996, opened 2021-09-30
Update testssl.1.md

dcooper16:
testssl.1.md included `.SS "SINGLE CHECK OPTIONS"`, which belongs in testssl.1, but not in testssl.1.md. This commit removes this extra line.


drwetter:
Thanks!

PR #1997, opened 2021-10-01
Remove the expired DST Root CA X3 cert from all trust stores, and ensure Mozilla's is up to date (fixes ISRG X1 alternate path)

### LINKED ISSUE #1995: [BUG] included openssl binaries report expired chain of trust for letsencrypt certificates
The root cert in the default chain (DST Root CA X3) for certificates issued by Let’s Encrypt expired today, September 30, 2021. 
Their documentation recommends upgrading to OpenSSL 1.1.0 or later: https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/

Indeed, the older openssl binary included with testssl.sh reports expired chain of trust for all letsencrypt certificates.
```
~] ./bin/openssl.Linux.x86_64 version
```

It's perhaps best to demonstrate with https://letsencrypt.org/ itself:
```
~] ./testssl.sh https://letsencrypt.org/
...
...
 Overall Grade                T
 Grade cap reasons            Grade capped to T. Issues with the chain of trust (expired)
```

mig5:
Fixes #1995

Even though it's not strictly necessary (you are using custom trust stores), I took the opportunity to remove the DST Root CA X3 cert from the `/etc/ssl/certs` trust store for good measure.

And though the ISRG X1 CA cert was in the Mozilla.pem cert, and I had removed the DST Root CA X3 cert from that store, that trust store still failed validation for LetsEncrypt certs, even after all the other trust stores started succeeding once I removed the DST Root CA X3 cert. I suspect there was some other issue in that trust store that broke the chain path. I fetched the latest version from https://curl.se/docs/caextract.html and after that - all good!


drwetter:
Hi, 

sorry for the delayed reply. Thanks!

I believe the changes in the Dockerfiles shouldn't be necessary. 

The update script from utils (create ca hashes) needs to update the hashes file in etc. Could you amend the PR with that file?

And would you mind to submit a PR for 3.0 too?

Cheers, Dirk

mig5:
@drwetter thanks - force pushed an updated commit with Dockerfile changes undone and the hashes updated.

3.0 PR at https://github.com/drwetter/testssl.sh/pull/2000

drwetter:
Thanks again

PR #1999, opened 2021-10-01
Remove the expired DST Root CA X3 cert from all trust stores, and ensure Mozilla's is up to date (fixes ISRG X1 alternate path) - 3.0 branch

mig5:
Corresponding PR like https://github.com/drwetter/testssl.sh/pull/1997, but for the 3.0 branch. The Java.pem already seemed correct in 3.0 somehow (the expired cert was not there)


mig5:
Hmm not sure how I've ended up with conflicts here :/ looking into it..

PR #2000, opened 2021-10-01
Remove the expired DST Root CA X3 cert from all trust stores, and ensure Mozilla's is up to date (fixes ISRG X1 alternate path) - 3.0 branch

mig5:
Fixed re-attempt of https://github.com/drwetter/testssl.sh/pull/1999 (it's a real nightmare when tags and branches have identical names..)


drwetter:
Thanks a lot, @mig5

drwetter:
> Fixed re-attempt of #1999 (it's a real nightmare when tags and branches have identical names..)

yes, known problem. My fingers were faster that they should be when preparing  the 3.0 release. I now tried to solve that, see #2004 .

PR #1998, opened 2021-10-01
In determine_trust(), use the OpenSSL '-trusted_first' flag to ignore the now-expired DST Root CA X3 root certificate

smokris:
Fixes #1995.

Tested:

- bundled OpenSSL 1.0.2-chacha on CentOS Stream 8 Linux 4.18.0-338.el8.x86_64
- system OpenSSL 1.1.1k on CentOS Stream 8 Linux 4.18.0-338.el8.x86_64
- bundled OpenSSL 1.0.2-chacha on macOS 11.6
- system LibreSSL 2.8.3 on macOS 11.6

PR #2002, opened 2021-10-02
In determine_trust(), use the OpenSSL '-trusted_first' flag to ignore the now-expired DST Root CA X3 root certificate

### LINKED ISSUE #1995: [BUG] included openssl binaries report expired chain of trust for letsencrypt certificates

smokris:
Fixes #1995.

Tested:

- bundled OpenSSL 1.0.2-chacha on CentOS Stream 8 Linux 4.18.0-338.el8.x86_64
- system OpenSSL 1.1.1k on CentOS Stream 8 Linux 4.18.0-338.el8.x86_64
- bundled OpenSSL 1.0.2-chacha on macOS 11.6
- system LibreSSL 2.8.3 on macOS 11.6

(Resubmitting due to accidental merge-then-revert of #1998.)

### DIFF
--- a/t/51_badssl.com.t
+++ b/t/51_badssl.com.t
@@ -22,6 +22,7 @@ cmp_ok(@$okjson,'>',10,"We have more then 10 findings"); $tests++;
 # Expiration
 pass("Running testssl against expired.badssl.com"); $tests++;
 $out = `./testssl.sh -S --jsonfile tmp.json --color 0 expired.badssl.com`;
+like($out, qr/Chain of trust\s+NOT ok \(expired\)/,"The chain of trust should be expired"); $tests++;
 like($out, qr/Certificate Validity \(UTC\)\s+expired/,"The certificate should be expired"); $tests++;
 $json = json('tmp.json');
 unlink 'tmp.json';
--- a/testssl.sh
+++ b/testssl.sh
@@ -7264,9 +7264,9 @@ determine_trust() {
           # in a subshell because that should be valid here only
           (export SSL_CERT_DIR="/dev/null"; export SSL_CERT_FILE="/dev/null"
           if [[ $certificates_provided -ge 2 ]]; then
-               $OPENSSL verify -purpose sslserver -CAfile <(cat $ADDTL_CA_FILES "$bundle_fname") -untrusted $TEMPDIR/intermediatecerts.pem $HOSTCERT >$TEMPDIR/${certificate_file[i]}.1 2>$TEMPDIR/${certificate_file[i]}.2
+               $OPENSSL verify $TRUSTED1ST -purpose sslserver -CAfile <(cat $ADDTL_CA_FILES "$bundle_fname") -untrusted $TEMPDIR/intermediatecerts.pem $HOSTCERT >$TEMPDIR/${certificate_file[i]}.1 2>$TEMPDIR/${certificate_file[i]}.2
           else
-               $OPENSSL verify -purpose sslserver -CAfile <(cat $ADDTL_CA_FILES "$bundle_fname") $HOSTCERT >$TEMPDIR/${certificate_file[i]}.1 2>$TEMPDIR/${certificate_file[i]}.2
+               $OPENSSL verify $TRUSTED1ST -purpose sslserver -CAfile <(cat $ADDTL_CA_FILES "$bundle_fname") $HOSTCERT >$TEMPDIR/${certificate_file[i]}.1 2>$TEMPDIR/${certificate_file[i]}.2
           fi)
           verify_retcode[i]=$(awk '/error [1-9][0-9]? at [0-9]+ depth lookup:/ { if (!found) {print $2; found=1} }' $TEMPDIR/${certificate_file[i]}.1 $TEMPDIR/${certificate_file[i]}.2)
           [[ -z "${verify_retcode[i]}" ]] && verify_retcode[i]=0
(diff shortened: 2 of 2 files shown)


drwetter:
Thanks, @smokris . 

@dcooper16 : could you please review?

dcooper16:
Hi @smokris, @drwetter,

I tested this PR, and while at first glance it seems to work with LibreSSL, it actually doesn't. I haven't tested this PR thoroughly with OpenSSL 1.0.2-chacha to ensure that adding the `-trusted_first` flag doesn't cause any unexpected problems, but it seems unlikely that it would.

Assuming the `-trusted_first` flag doesn't cause any unexpected problems with OpenSSL 1.0.2-chacha, this PR will need to be modified to first check whether `$OPENSSL` supports the `-trusted_first` flag and then only include it in the `$OPENSSL verify` command line if it is supported.

To see that the current PR is not working as expected, try testing expired.badssl.com. With OpenSSL 1.0.2-chacha or OpenSSL 1.1.1, the chain of trust is correctly reported as "NOT ok (expired)". With LibreSSL, it is reported as "Ok".

The reason that the PR seems to work even though it doesn't is as follows. The relevant code is:
```
if [[ $certificates_provided -ge 2 ]]; then
     $OPENSSL verify -trusted_first -purpose sslserver -CAfile <(cat $ADDTL_CA_FILES "$bundle_fname") -untrusted $TEMPDIR/intermediatecerts.pem $HOSTCERT >$TEMPDIR/${certificate_file[i]}.1 2>$TEMPDIR/${certificate_file[i]}.2
else
     $OPENSSL verify -trusted_first -purpose sslserver -CAfile <(cat $ADDTL_CA_FILES "$bundle_fname") $HOSTCERT >$TEMPDIR/${certificate_file[i]}.1 2>$TEMPDIR/${certificate_file[i]}.2
fi)
verify_retcode[i]=$(awk '/error [1-9][0-9]? at [0-9]+ depth lookup:/ { if (!found) {print $2; found=1} }' $TEMPDIR/${certificate_file[i]}.1 $TEMPDIR/${certificate_file[i]}.2)
[[ -z "${verify_retcode[i]}" ]] && verify_retcode[i]=0
```
The result of running the `$OPENSSL verify` command with LibreSSL 2.8.3 on MacOS is:
```
usage: verify [-verbose] [-CApath path] [-CAfile file] [-purpose purpose] [-crl_check] [-attime timestamp] cert1 cert2 ...
recognized usages:
	sslclient 	SSL client
	sslserver 	SSL server
	nssslserver	Netscape SSL server
	smimesign 	S/MIME signing
	smimeencrypt	S/MIME encryption
	crlsign   	CRL signing
	any       	Any Purpose
	ocsphelper	OCSP helper
	timestampsign	Time Stamp signing
```
Since the output of `$OPENSSL verify` does not contain a string that matches
```
'/error [1-9][0-9]? at [0-9]+ depth lookup:
```
`determine_trusted()` incorrectly thinks that verification of the certificate was successful, even though `$OPENSSL` is reporting that the operation could not be performed.

drwetter:
so I guess a line like 

```
$OPENSSL verify -trusted_first </dev/null 2>&1 | grep -q '^usage' || HAS_TRUSTED1ST=true
```

at a place where all the ``HAS_*`` variables are defined would help (hint for @smokris) ?

dcooper16:
Another option would be to do something like what is done with `OSSL_CIPHERS_S`. Set `TRUSTED1ST` to the empty string and then have a line in `find_openssl_binary()` that is something like:
```
$OPENSSL verify -trusted_first </dev/null 2>&1 | grep -q '^usage' || TRUSTED1ST="-trusted_first"
```
Then `$TRUSTED1ST` could just be used in the `$OPENSSL verify` command line rather than needing an `if` statement.

smokris:
Thanks for the detailed feedback, @drwetter and @dcooper16.  I pushed a new version that checks whether the `-trusted_first` option exists.  I also added a line to `t/51_badssl.com.t` that should (when using LibreSSL) have detected the false-negative that @dcooper16 described above.

drwetter:
Thanks!

Looks to me like it should work. @dcooper16: can you confirm it does on LibreSSL?

dcooper16:
Hi @drwetter , @smokris ,

The PR seems to be working as expected. However, in order to avoid future problems, I would like to suggest one minor change. In the two places in testssl.sh that have the line `OSSL_CIPHERS_S=""`, add `TRUSTED1ST=""`.

Imagine a possible future scenario in which MacOS uses LibreSSL 3.1 or later (i.e., a version that supports TLS 1.3 on the client), but a version of testssl.sh is being used that comes with OpenSSL 1.0.2-chacha. If a TLS 1.3-only server is being tested:

- `find_openssl_binary()` will be called and `TRUSTED1ST` will be set to `-trusted_first`, since OpenSSL 1.0.2-chacha supports that option.
- `determine_optimal_proto()` will determine that the server is TLS 1.3-only, OpenSSL 1.0.2-chacha doesn't support TLS 1.3, but that /usr/bin/openssl does.
- `determine_optimal_proto()` will set `OPENSSL` to `/usr/bin/openssl` and then call `find_openssl_binary()`.
- If `TRUSTED1ST` isn't reinitialized to the empty string, then its value will remain as `-trusted_first` even though `/usr/bin/openssl` doesn't support that option, and so the check in `determine_trust()` won't be performed correctly.

smokris:
Ah, that makes sense.  I just pushed 5c4500e which adds that.

drwetter:
Thanks to both/

ok, @dcooper16 ?

dcooper16:
It looks good to me.

drwetter:
Thanks!

Issue #1750, opened 2020-10-19
Unable to use testssl.sh in a proxy-required / no public DNS environment due to DNS failures


mzpqnxow:
Trying to use testssl.sh (the latest dev branch on commit e51301d9ee2956f2c70b68a0f511925c1f33be28) it seems proxying does work in my configuration because of the DNS checks that take place before any of the actual SSL protocol stuff happens. This seems to be a somewhat known issue- I read https://github.com/drwetter/testssl.sh/issues/748 and what I got out of that was that this was resolved. But it doesn't seem to be the case- or more likely I misunderstood what was "fixed"- maybe having public DNS is an absolute requirement as it's currently written?

My environment is a typical strict/secure "corporate" environment:

  1. No Internet DNS available, only internal DNS
  2. No direct outbound TCP/UDP to WAN
  2. HTTP proxy available, supporting CONNECT

The error I encounter is:

```
17:13:37 › ./testssl.sh -n none https://test.com 
For now I am providing the config file to have GOST support

###########################################################
    testssl.sh       3.1dev from https://testssl.sh/dev/
    (e51301d 2020-10-17 17:04:49 -- )

      This program is free software. Distribution and
             modification under GPLv2 permitted.
      USAGE w/o ANY WARRANTY. USE IT AT YOUR OWN RISK!

       Please file bugs @ https://testssl.sh/bugs/

###########################################################

 Using "OpenSSL 1.0.2-chacha (1.0.2k-dev)" [~179 ciphers]
 on oseu501665:./bin/openssl.Linux.ppc64le
 (built: "Sep 15 18:42:12 2020", platform: "linux-ppc64le")



Fatal error: No IPv4/IPv6 address(es) for "test.com" available
```

I thought that by specifying `-n none` there would not be any DNS requests made, but when I read through `testssl.sh` I see a function that does DNS resolution, so maybe `-n` only disables some of the more detailed DNS checks that `testssl.sh` does?

Is my use-case supported, or do you intend to support it? There are quite a few ways to hack around this but none of them really suit my needs, which is supporting an internal web interface to `testssl.sh`. I could hook the `gethostbyname` libc functions with `LD_PRELOAD` and return a dummy address but I'm not even sure that would be a quick fix as I don't know what will be done with the IP addresses that are returned

Thanks, appreciate your work on this and any info you can provide

[...]

PR #2011, opened 2021-10-15
Add --nodns=strict for #1750 (environments with proxy and no public DNS services)

mzpqnxow:
This is one way to implement a fix for #1750

It's certainly not elegant, though the code changes are minimal

It requires regression testing and would benefit from feedback on a better way to approach it. It's very difficult working on such a gigantic shell script :)

EDIT: I'm open to suggestions for better ways to implement it. Like I said, it's very hard to trace the flow so I took the path of least resistance as I saw it

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6550,7 +6550,7 @@ run_server_preference() {
 
      "$SSL_NATIVE" && using_sockets=false
 
-     if [ "$NODNS" == "strict" ]; then
+     if [[ "$NODNS" == "strict" ]]; then
         NODEIP="$NODE"
      fi
 
@@ -20157,7 +20157,7 @@ get_a_record() {
      local noidnout=""
 
      "$HAS_DIG_NOIDNOUT" && noidnout="+noidnout"
-     if [ "$NODNS" == none ] || [ "$NODNS" == strict ]; then
+     if [[ "$NODNS" == none ]] || [[ "$NODNS" == strict ]]; then
       return 0      # if no DNS lookup was instructed, leave here
      fi
 
@@ -20380,7 +20380,7 @@ determine_ip_addresses() {
      local ip4=""
      local ip6=""
 
-     if [ "$NODNS" == "strict" ]; then
+     if [[ "$NODNS" == "strict" ]]; then
         # Use a sentinel value; avoid 0.0.0.0 as it can be interpreted as loopback
         CMDLINE_IP='-1.-1.-1.-1'
      fi
@@ -20405,7 +20405,7 @@ determine_ip_addresses() {
                ip4="$NODEIP"
           elif is_ipv6addr "$NODEIP"; then
                ip6="$NODEIP"
-          elif [ "$NODNS" == "strict" ]; then
+          elif [[ "$NODNS" == "strict" ]]; then
                ip4="$NODE"
           else
                fatal "couldn't identify supplied \"CMDLINE_IP\"" $ERR_DNSLOOKUP
@@ -20837,7 +20837,7 @@ determine_optimal_proto() {
 
      "$do_tls_sockets" && return 0
 
-     if [ "$NODNS" == "strict" ]; then
+     if [[ "$NODNS" == "strict" ]]; then
         # The checks all use $NODEIP, so in strict mode we need to set it to $NODE
         # There is no such thing as a real $NODEIP value in strict mode
         # This check and assignment can (and probably should) take place outside of this function
@@ -21148,7 +21148,7 @@ display_rdns_etc() {
      local ip further_ip_addrs=""
      local nodeip="$(tr -d '[]' <<< $NODEIP)"     # for displaying IPv6 addresses we don't need []
 
-     if [ "$NODNS" == "strict" ]; then
+     if [[ "$NODNS" == "strict" ]]; then
        return 0
      fi
 
@@ -23074,8 +23074,6 @@ lets_roll() {
 
      [[ -z "$NODE" ]] && parse_hn_port "${URI}"        # NODE, URL_PATH, PORT, IPADDRs and IP46ADDR is set now
      prepare_logging
-
-
 
      if ! determine_ip_addresses; then
           fatal "No IP address could be determined" $ERR_DNSLOOKUP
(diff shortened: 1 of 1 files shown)


drwetter:
ACK & got it. It takes a while until I am able to look into it. 

One thing which was obvious is the code style. We use double square brackets and inside those if no quoting is required unless you need to. Otherwise inside double brackets every single \c\h\a\r\ \i\s \q\u\o\t\e\d during runtime.

mzpqnxow:
> ACK & got it. It takes a while until I am able to look into it.

No rush, we're all busy

> One thing which was obvious is the code style. We use double square brackets and inside those if no quoting is required unless you need to. Otherwise inside double brackets every single \c\h\a\r\ \i\s \q\u\o\t\e\d during runtime.

Gotcha, I'll update it to be consistent with that. I'm usually pretty strict about adhering to project style, sorry 'bout that

mzpqnxow:
I fixed those issues and also removed 2 lines of whitespace that got added somehow. Would you like me to add something to the documentation or is it OK to worry about that later? I'm not sure anyone else is going to be using this feature anyway since I'm the only one to have ever complained about the problem :)

testssl.sh
```diff
@@ -20374,8 +20380,14 @@ determine_ip_addresses() {
      local ip4=""
      local ip6=""
 
+     if [[ "$NODNS" == "strict" ]]; then
```
drwetter, line 20383: here and some instances below the quotes aren't necessary

drwetter:
> Would you like me to add something to the documentation or is it OK to worry about that later

good point. One step at a time

testssl.sh
mzpqnxow: I just pushed a commit to clean all of these up

mzpqnxow (commented):
All set with those

drwetter:
Sorry for the delay.

First of all: basically ``--n  none`` should have sufficed. If not I would suggest to identify the places where this isn't enough. Then decide whether we want to improve those occurrences or introduce another flag.

Then: as indicated you put lines like ``if [[ "$NODNS" == strict ]]`` in low level functions like ``determine_optimal_proto`` or ``run_server_preference()`` rather than using the calling function(s). 

(Minor thing: Indentation is 5 chars, no 2 YAML style here)

So as indicated overall it doesn't make much sense to me, thus I rather close this.

mzpqnxow:
@drwetter I'm confused- you're saying you don't want to accept it? Either I'm misunderstanding or possibly you're not remembering the issue that started this whole thing- unfortunately, it's a long discussion - https://github.com/drwetter/testssl.sh/issues/1750. We discussed at length:

1. The issue
2. The potential fixes

The summary of the issue: When openssl.sh is run in a very restrictive environment (e.g. enterprise environment with no access to public DNS resolvers, only HTTP proxy only for access to the Internet) it bombs out even when you direct it to inhibit DNS. It makes at least one (for what reasons I now don't remember, but it was not necessary)

There were a few hacks to work around the issue- some things like setting the IP address to 0.0.0.0 or something similar, or adding a dummy entry to /etc/hosts every time to you want to assess an endpoint

If you wouldn't mind, can you just confirm that you glanced through that issue so we're on the same page?

Thanks!

drwetter:
See comments above. It's hackish.

Please also see here https://github.com/drwetter/testssl.sh/issues/1750#issuecomment-998817440 .

Don't know we're on the same page but I respect and understood you were saying. :-). Not sure whether you got my points though? The last paragraph, see link above, provides a solution though.

mzpqnxow:
> First of all: basically ``--n  none`` should have sufficed

Yes, I felt the same way :)

> If not I would suggest to identify the places where this isn't enough.
> Then decide whether we want to improve those occurrences or introduce another flag.

From the linked issue (your comments)

> TL;DR:
> testssl.sh --proxy=http://<your_proxyip> -n none --ip=0.0.0.0 https://www.homedepot.com works but requires a confirmation:
 > Your OpenSSL cannot connect to 0.0.0.0:443
 > The results might look ok but they could be nonsense. Really proceed ? ("yes" to continue) --> yes
> wheres ./testssl.sh --proxy=http://<your_proxyip> -n none https://www.homedepot.com doesn't work.

> ./testssl.sh --proxy=auto -n none --ip=0.0.0.0 https://www.homedepot.com works when http_proxy is set to an IP.

> It should work also when in /etc/hosts DNS resolution is supplied.

I had pointed those out as well and asked "do you want a PR, intend to "fix" it yourself or do you want to leave the workarounds as the solution or say it's won't fix/as expected," to which you said "there will be a proper solution, and I'll see what you propose," hence this PR (admittedly 9 months or so late, I forgot to commit the changes back to my repo after they worked in my environment)

I need this either way because of my use-case, which is calling it from a restricted environment from a web application, where HTTP proxies have DNS names. None of the workarounds we discussed in the original issue were great for me

- Hard-coding the proxy IP causes spurious failures because the HTTP proxies in my environment are referenced by a DNS name, to facilitate round-robin and maintenance. I could add application logic in my application to resolve the address immediately before calling testssl.sh, though

- It's not practical for me to add entries to /etc/hosts dynamically. It's just a bad idea. I also don't can't get the IP of the site from my environment

- Having to keep a pipe open to send a newline to the prompt is clunky (though you can say my PR is clunky too, I was aiming for a "proper" solution rather than a workaround)

> So as indicated overall it doesn't make much sense to me, thus I rather close this.

Fair enough, sorry for the confusion and probably the formatting (I'm on mobile)

Thanks for the project and for your patience

mzpqnxow:
> I would suggest to identify the places where this isn't enough. Then decide whether we want to improve those occurrences or introduce another flag.

FWIW- that's exactly what this is/was, though in retrospect making it a third option for --nodns was probably not the right choice and might have been the cause for requiring so many changes throughout the flow

drwetter:
hi @mzpqnxow ,

thanks for your reply.

As said technically your PR works with the constraint that it only tests one hostname using random IPs. That's a constraint of using a proxy resolving hostnames, not your PR.

I understand that it is not easy for you finding the right spot where to add the lines to make it work. 

#1750 is still open and if not you or anybody else can';t do this, I'll find the time probably.

Thanks!

Issue #1812, opened 2021-01-02
[BUG] Script execution massively delayed


patschi:
**Command line / docker command to reproduce**
Running `testssl.sh`, parameters doesn't matter.

**Current issue**
1. Cloned git repo via `git clone --depth 1 https://github.com/drwetter/testssl.sh.git` (This issue is NOT reproducible on latest `3.0` version)
2. Running testssl.sh takes 2.5 minutes, also for simply displaying the version or the current banner. Some process is blocking the execution on the latest git commit state.

Example:
```
~/testssl.sh_git$ time ./testssl.sh -v
+ declare -r ERR_BASH=255
+ declare -r ERR_CMDLINE=254
+ declare -r ERR_FCREATE=253
+ declare -r ERR_FNAMEPARSE=252
+ declare -r ERR_NOSUPPORT=251
+ declare -r ERR_OSSLBIN=250
+ declare -r ERR_DNSBIN=249
+ declare -r ERR_OTHERCLIENT=248
+ declare -r ERR_DNSLOOKUP=247
+ declare -r ERR_CONNECT=246
+ declare -r ERR_CLUELESS=245
+ declare -r ERR_RESOURCE=244
+ declare -r ERR_CHILD=242
+ declare -r ALLOK=0
+ '[' -z 5 ']'
++ kill -l
++ grep -c SIG
+ '[' 13 -eq 0 ']'
+ '[' 5 -lt 3 ']'
+ '[' 5 -le 3 ']'
+ declare -r 'PS4=|${LINENO}> \011${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
|100>   DEBUGTIME=false
|101>   DEBUG_ALLINONE=false
|103>   DEBUG_ALLINONE=false
|104>   [[ braceexpand:hashall:interactive-comments:xtrace =~ xtrace ]]
|105>   false
|110>   false
|111>   exec
|112>   BASH_XTRACEFD=42


###########################################################
    testssl.sh       3.1dev from https://testssl.sh/dev/
    (a9f4bb5 2020-12-29 14:40:58 -- )

      This program is free software. Distribution and
             modification under GPLv2 permitted.
      USAGE w/o ANY WARRANTY. USE IT AT YOUR OWN RISK!

       Please file bugs @ https://testssl.sh/bugs/

###########################################################

 Using "OpenSSL 1.1.1i  8 Dec 2020" [~98 ciphers]
 on XXX:/usr/bin/openssl
[...]

PR #2012, opened 2021-10-18
Addressing lame DNS responses on WSL (3.0)

drwetter:
This PR provides a global variable to the RFC 6761 use of "invalid."
which WSL clients don't seem to handle very well, see #1738, #1812.
"invalid."  is used as a target to find out in a couple of pre-checks what
is supported by the openssl version.

This PR reduces the number of ``openssl s_client -connect`` by a huge factor.
For the remaining invocations the OS used is being determined and if WSL is
assumed (the check is probably not 100% accurate) it uses ``127.0.0.1:0`` instead.
In (unfortunately only a few) pre-tests the response was immediate.

Also it is possible to use another target if needed by

     NXCONNECT=<mytargethere>:<myport> ./testssl.sh <URL>

This is for 3.0. For 3.1dev see #1988 .

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -220,6 +220,7 @@ FNAME=${FNAME:-""}                      # file name to read commands from
 FNAME_PREFIX=${FNAME_PREFIX:-""}        # output filename prefix, see --outprefix
 APPEND=${APPEND:-false}                 # append to csv/json file instead of overwriting it
 [[ -z "$NODNS" ]] && declare NODNS      # If unset it does all DNS lookups per default. "min" only for hosts or "none" at all
+NXCONNECT=${NXCONNECT:-invalid.}        # For WSL this helps avoiding DNS requests to "invalid." which windows seem to handle delayed
 HAS_IPv6=${HAS_IPv6:-false}             # if you have OpenSSL with IPv6 support AND IPv6 networking set it to yes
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -17015,92 +17022,54 @@ find_openssl_binary() {
...
+     $OPENSSL s_client -comp 2>&1 | grep -aiq "unknown option" || HAS_COMP=true
+     $OPENSSL s_client -no_comp 2>&1 | grep -aiq "unknown option" || HAS_NO_COMP=true
 
      OPENSSL_NR_CIPHERS=$(count_ciphers "$(actually_supported_osslciphers 'ALL:COMPLEMENTOFALL' 'ALL')")
 
-     if $OPENSSL s_client -curves "${curves_ossl[0]}" -connect invalid. 2>&1 | grep -aiq "unknown option"; then
+     if $OPENSSL s_client -curves "${curves_ossl[0]}" -connect $NXCONNECT 2>&1 | grep -aiq "unknown option"; then
           for curve in "${curves_ossl[@]}"; do
-               $OPENSSL s_client -groups $curve -connect invalid.:8443 2>&1 | grep -Eiaq "Error with command|unknown option|Failed to set groups"
+               $OPENSSL s_client -groups $curve -connect ${NXCONNECT%:*}:8443 2>&1 | grep -Eiaq "Error with command|unknown option|Failed to set groups"
```
drwetter, line 17047: Note to self: This needs to be in 3.1dev too.

PR #2016, opened 2021-10-20
Fix double ports statement

drwetter:
In order to be in line with 3.0 this adds a statement to avoid double port
statements for WSL users in a precheck condition

   $OPENSSL s_client -groups $curve -connect 127.0.0.1:0:8443

This commit changes that to $OPENSSL s_client -groups $curve -connect 127.0.0.1:8443
which could cause a delay for WSL users and needs further investigation.

Not clear why port 8443 was chosen.

See also #2012

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -19304,7 +19304,7 @@ find_openssl_binary() {
 
      if $OPENSSL s_client -curves "${curves_ossl[0]}" -connect $NXCONNECT 2>&1 | grep -aiq "unknown option"; then
           for curve in "${curves_ossl[@]}"; do
-               $OPENSSL s_client -groups $curve -connect $NXCONNECT:8443 2>&1 | grep -Eiaq "Error with command|unknown option|Failed to set groups"
+               $OPENSSL s_client -groups $curve -connect ${NXCONNECT%:*}:8443 2>&1 | grep -Eiaq "Error with command|unknown option|Failed to set groups"
                [[ $? -ne 0 ]] && OSSL_SUPPORTED_CURVES+=" $curve "
           done
      else


drwetter:
@dcooper16 : not sure whether it was me or you who did this check with the port 8443. If it was you: what was the reason for this specific port?

dcooper16:
Hi Dirk,

It seems that I added the port 8443 in #1598. My guess is that I needed to specify _some_ port number, but the particular number didn't matter. However, I just did some quick testing now, and it seems that the `-connect` option can be omitted entirely, as you did for many other tests in #2012. I haven't tested it with many different versions of OpenSSL/LibreSSL, but it is something to look into.

drwetter:
Hi David,

Thanks. it looks like a port number would be needed also a few lines above. And ``invalid.`` + portnumber doesn't work as I thought.  

And: not sure why there's "${curves_ossl[**0**]}" above. Is that just to test the first curve?

drwetter:
> However, I just did some quick testing now, and it seems that the -connect option can be omitted entirely, as you did for many other tests in #2012.

PS: As soon as I add ``-curves`` to ``openssl s_client`` it seems implicitly to connect to ``127.0.0.1:4433`` with both the supplied openssl and version 1.1.1 from my distributor.

dcooper16:
> Thanks. it looks like a port number would be needed also a few lines above. And invalid. + portnumber doesn't work as I thought.
>
> And: not sure why there's "${curves_ossl[0]}" above. Is that just to test the first curve?

The first line was just intended to determine whether the `-curves` option is supported, so I just picked one curve at random to use on the command line (`${curves_ossl[0]}`). There was no need to specify a port number, since only LibreSSL requires a port number, and LibreSSL will issue an "unknown option" error when the `-curves` option is used, even if `-connect` does not specify a port number.

It seems, however, that all of this could be simplified. (However, again, I would need to test with more different versions of OpenSSL/LibreSSL to be sure). It seems that
```
$OPENSSL s_client -curves 2>&1 | grep -aiq "unknown option"
```
may be sufficient to determine whether the `-curves` option is supported.

> PS: As soon as I add -curves to openssl s_client it seems implicitly to connect to 127.0.0.1:4433 with both the supplied openssl and version 1.1.1 from my distributor.

I am seeing the same thing, using either `-curves` with OpenSSL or `-groups` with LibreSSL, if the specified group is supported. In fact, I'm seeing OpenSSL/LibreSSL connect to 127.0.0.1:4433 with just `openssl s_client`.

drwetter:
> The first line was just intended to determine whether the -curves option is supported,

ok, thanks, David.

I decided to merge this for now. More later.

PR #2022, opened 2021-10-25
Fix s_client_options()

dcooper16:
This PR fixes an error in `s_client_options()` when the function is called with an empty "-cipher" list and $OPENSSL supports "@SECLEVEL". This happens, for example, when `ciphers_by_strength()` is called for TLS 1.3 and OpenSSL 1.1.1 is being used.

The call to `openssl s_client` will fail is the cipher option is provided with an empty list or a list that just contains "@SECLEVEL=0". So, "@SECLEVEL=0" should only be added if the "$ciphers" list is non-empty. If "$ciphers" remains empty, then the "-cipher" option will not be added to the command line.


drwetter:
Thanks, David

Issue #1733, opened 2020-09-25
PrioritizeChaCha detection


army1349:
Would be possible to add [PrioritizeChaCha](https://www.openssl.org/docs/man1.1.1/man3/SSL_CONF_cmd_value_type.html) detection to server cipher order test?
For example SSL Labs positive detection results in: _(P) This server prefers ChaCha20 suites with clients that don't have AES-NI (e.g., Android devices)_
Thanks!

drwetter:
Can you show me a result from SSL labs?

First is security, so I am hesitant for giving any performance related advise. As a hint: maybe

army1349:
Sure, [here](https://www.ssllabs.com/ssltest/analyze.html?d=cloudflare.com&s=104.17.175.85) cloudflare test result.

Well, this is kind of server cipher order preference exception, so definitely about security.

polarathene:
> First is security, so I am hesitant for giving any performance related advise.

Isn't the cipher considered PFS? According to Mozilla's advice, when all ciphers are considered strong due to being PFS, the order doesn't matter as much, so you can allow client preference for devices to choose what they'd be more efficient at handling, such as the case here with devices lacking AES-NI.

drwetter:
Sure. But the point is that AES ciphers are - of course - still supported. It's only slower. Have a look at the ssllabs client handshakes and try yourself at ssllabs.

army1349:
My point is that cipher order information is now not entirely true or at least incomplete when PrioritizeChaCha is enabled.

And it would be nice confirmation that PrioritizeChaCha works correctly.
Internet.nl test misdetects this entirely, but they [plan](https://github.com/NLnetLabs/Internet.nl/issues/461) to add support for correct detection.

drwetter:
ok, For servers using OpenSSL you have to edit a config file if I see that correctly.

Maybe something @dcooper16 at some point ca look into, he did an excellent job for the current preference implementation

army1349:
Direct exposure of OpenSSL options is not very common, so I started to use OpenSSL config files anyway.
Postfix has native [support](http://www.postfix.org/postconf.5.html#tls_ssl_options). Nginx [resisting](https://trac.nginx.org/nginx/ticket/1445).

PR #2024, opened 2021-10-25
Improve cipher_pref_check()

dcooper16:
Some servers are configured to prioritize ChaCha ciphers if those ciphers are preferred by the client, even if the server is generally configured to use the server's cipher preferences rather than the client's. As a result of this, if a ChaCha cipher appears in the ClientHello before a non-ChaCha cipher, the server may select the ChaCha cipher even if the server is configured to prefer the non-ChaCha cipher.

In a few cases, e.g., cloudflare.com for TLS 1.2, this affects the ordering of the ciphers presented by `cipher_pref_check()`. This PR fixes the problem by having `cipher_pref_check()` (and `check_tls12_pref()`) always place any ChaCha ciphers at the end of the cipher list in the ClientHello. This ensures that `cipher_pref_check()` presents the ciphers in the server's preference order.

NOTE: After this PR has been merged, a follow-up PR will be submitted that checks whether the server is configured to prioritize ChaCha ciphers (as was requested in #1733.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6841,14 +6841,24 @@ run_server_preference() {
 # arg1: true if the list that is returned does not need to be ordered by preference.
 check_tls12_pref() {
      local unordered_list_ok="$1"
+     local chacha20_ciphers="" non_chacha20_ciphers=""
      local batchremoved="-CAMELLIA:-IDEA:-KRB5:-PSK:-SRP:-aNULL:-eNULL"
      local batchremoved_success=false
      local tested_cipher="" cipher ciphers_to_test
      local order=""
      local -i nr_ciphers_found_r1=0 nr_ciphers_found_r2=0
 
+     # Place ChaCha20 ciphers at the end of the list to avoid accidentally
+     # triggering the server's PrioritizeChaCha setting.
+     ciphers_to_test="$(actually_supported_osslciphers "ALL:$batchremoved" "" "")"
+     for cipher in $(colon_to_spaces "$ciphers_to_test"); do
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -7074,8 +7096,32 @@ cipher_pref_check() {
...
+                    ciphers_found2[nr_ciphers]=false
+                    hexcode[nr_ciphers]="${hexc:2:2},${hexc:7:2}"
+                    rfc_ciph[nr_ciphers]="${TLS_CIPHER_RFC_NAME[i]}"
+                    if [[ $proto == tls1_3 ]]; then
+                         [[ "${hexc:2:2}" == "13" ]] && nr_ciphers+=1
+                    elif [[ $proto == tls1_2 ]]; then
+                         [[ "${hexc:2:2}" != "13" ]] && nr_ciphers+=1
+                    elif [[ ! "${TLS_CIPHER_RFC_NAME[i]}" =~ SHA256 ]] && \
+                         [[ ! "${TLS_CIPHER_RFC_NAME[i]}" =~ SHA384 ]] && \
+                         [[ "${TLS_CIPHER_RFC_NAME[i]}" != *"_CCM" ]] && \
```
drwetter, line 7117: Could you double check pls whether we need the quotes here in the RHS? (also pls following line)

drwetter:
Hi David,

thanks. Please see the single comment. I always put trust in you that you have tested this, so I did not. So I would after resolving the minor thing merge this.

Doesn't this make sense for 3.0 too?

Cheers, Dirk

dcooper16:
Okay, I removed the unnecessary quotes. I just made a second copy of the code that was already there and didn't notice that it contained quotes that weren't needed.

I'll work on a PR for the 3.0 branch.

drwetter:
Thanks a lot, David

PR #2026, opened 2021-10-26
Speedup cipher_pref_check()

dcooper16:
While OpenSSL 1.0.2-chacha supports more than 128 cipher suites, other versions of OpenSSL tend to support far fewer than that.

Calling `check_tls12_pref()` adds to the run time of `cipher_pref_check()`, but is only needed if OpenSSL would otherwise send a ClientHello with more than 128 cipher suites. So, this PR adds another check and only calls `check_tls12_pref()` if the server has the size limit bug and OpenSSL supports more than 127 ciphers.


drwetter:
thanks!

PR #2028, opened 2021-10-27
Fix #1733

### LINKED ISSUE #1733: PrioritizeChaCha detection

dcooper16:
This PR adds PrioritzeChaCha detection, as requested in #1733. If the server (1) enforces a cipher order for a given protocol, (2) supports at least one non-ChaCha cipher for that protocol, and (3) supports at least one ChaCha cipher that would normally be less preferred than the ChaCha cipher, then a test is performed to see if the server selects the ChaCha cipher if it is listed first in the ClientHello.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6828,7 +6828,6 @@ run_server_preference() {
                fi
                ciphers_by_strength "-$proto_ossl" "$proto_hex" "$proto_txt" "$using_sockets" "true"
           else
-               outln " (server order)"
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, David!

PR #2031, opened 2021-10-28
Fix 2030

dcooper16:
Fix CI issue created by #2028 and by new certificate fingerprint.


drwetter:
Oh, thanks David! Saw that but wasn't able to fix that yesterday, so before I forget this I took a note.

PR #2032, opened 2021-11-01
Consider HTTP 'Age' header field when calculating HTTP clock skew.

Wahnes:
Currently, when "HTTP clock skew" is calculated, this is taken to be the difference between the timestamp of the moment the HTTPS request was sent and the date given in the HTTP `Date` header. This does not yield valid results in case a HTTP cache is used, either on the client side or on the server side. According to the HTTP specs, the `Date` field will contain the timestamp the response was created, which may not be the timestamp the response was delivered.

Consider the following example that queries the Varnish project's web server. Note that Varnish is a popular HTTP caching server, so HTTP caching will of course be used when serving HTTP responses from this project's web server.
`testssl.sh https://varnish-cache.org/`
This will typically output a HTTP clock skew of some thousand seconds.

The patch takes into account the HTTP `Age` header that caching servers add to the HTTP response to signal the response's freshness. As client-side caches normally do not cache HTTPS requests (except maybe for "enterprise" HTTP proxy caches that do MITM HTTPS proxying), this is mostly targeted to HTTPS websites that employ server side HTTP caching.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -393,6 +393,7 @@ TLS_NOW=""                              # Similar
 TLS_DIFFTIME_SET=false                  # Tells TLS functions to measure the TLS difftime or not
 NOW_TIME=""
 HTTP_TIME=""
+HTTP_AGE=0
 REQHEADERS=()
 GET_REQ11=""
 START_TIME=0                            # time in epoch when the action started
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks! 

I believe I learnt something. :-) Your change makes sense. I'll get back later on this

drwetter:
See merges above into a separate PR which will be merged soon.

Thanks!

PR #2034, opened 2021-11-06
[Rating] Fix cipher rating, and add penalty for weak ciphers

### LINKED ISSUE #2033: [BUG] Server grade assessment when NULL ciphers are involved
When testing a server that offers `NULL` ciphers, i.e. just TLS handshake and no encryption, this is clearly marked in red letters as dangerous in the _Testing cipher categories_ section. For instance, the output may look like this:
```
 NULL ciphers (no encryption)                      offered (NOT ok)
 Anonymous NULL Ciphers (no authentication)        offered (NOT ok)
```

Unfortunately, this test result is not reflected in the overall server grade. A server that offers `NULL` encryption (like shown above) may receive an `A` or `A+` overall rating. This not only feels wrong, it also does not conform to the server rating rules.

Consider Table 5 of [SSL Labs Server Rating Guide](https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide#user-content-cipher-strength) which assigns a "0 %" score when there is no encryption. Given the calculation method for cipher strength (add strongest and weakest, divide by two), this should yield a 50 % score at most. With a 50 % score for cipher strength, the overall rating of a server that offers `NULL` encryption should be `C` at best.

The discussion of the "2009n" changes in the _Changes_ section in the [SSL Labs Server Rating Guide](https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide#changes-in-2009n-19-january-2017---3-april-2017) also mentions an "explicit penalty" for weak ciphers, but does not give details. Or at least I could not find any details on that. But I suppose that this would make `F` the appropriate grade for a server that offers `NULL` encryption at all.

magnuslarsen:
Gives a penalty of "F" when using cipher suites less than 112 bits.  
The rating guide is unclear which grade to cap it at, but [this blog-post](https://blog.qualys.com/product-tech/2017/01/18/ssl-labs-grading-changes-january-2017) confirms it (see `Cipher grading adjustments`)

It also fixes the category 3 calculations, and now correctly calculates the final score, [according to this, point 6](https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide#methodology-overview) 

Tested against `null.badssl.com`, which now gives out the correct score (and verified everything works with `google.com` and `dr.dk`)

My editor took out some white-space, hope that is okay, else let me know

fixes https://github.com/drwetter/testssl.sh/issues/2033

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -22019,7 +22019,12 @@ run_rating() {
           fileout "cipher_strength_score_weighted" "INFO" "$c3_wscore"
 
           ## Calculate final score and grade
-          let final_score=$c1_wscore+$c2_wscore+$c3_wscore
+          # If any category resulted in a score of 0, push final grade to 0
+          if [[ $c1_score -eq 0 || $c2_score -eq 0 || $c3_score -eq 0 ]]; then
+               let final_score=0
+          else
+               let final_score=$c1_wscore+$c2_wscore+$c3_wscore
+          fi
 
           pr_bold " Final Score                  "; outln $final_score
           fileout "final_score" "INFO" "$final_score"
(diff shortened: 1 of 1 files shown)


drwetter:
Looked good, thank you!

PR #1963, opened 2021-08-07
Redirect debugme() outputs to stderr (3.1dev) 

drwetter:
... to address #1956 and other places. Further details see commits

PR #2039, opened 2021-11-09
Fix calls to debugme1() with redirects

dcooper16:
As noted in https://github.com/drwetter/testssl.sh/pull/1963#issuecomment-954935175, there are two places in which `debugme1()` is called in which the command to be run prints something to stdout, which is then redirected to a file. Now that `debugme1()` includes a redirect from stdout to stderr, these two lines do not work as expected. The output appears in the terminal and an empty file is created.

This PR fixes the issue by not using the `debugme1()` function in these two places.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7626,7 +7626,7 @@ determine_tls_extensions() {
      fi
 
      # Keep it "on file" for debugging purposes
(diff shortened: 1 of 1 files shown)


drwetter:
Yes, thank you @dcooper16 !

PR #2040, opened 2021-11-09
Support OpenSSL 3.0.1-dev and 3.1.0-dev

dcooper16:
This PR changes a few lines of code that check OpenSSL version numbers to recognize 3.0.1 and 3.1.0 in addtion to 3.0.0.

According to https://www.openssl.org/blog/blog/2018/11/28/version, API/ABI compatibility is guaranteed for the same MAJOR version number (starting with version 3), so it should be sufficient to just check that $OSSL_VER_MAJOR is 3. This blog post and the code in the https://github.com/openssl/openssl/tree/openssl-3.0 branch also suggests that the first bug fix version following 3.0.0 will be 3.0.1 rather than 3.0.0a.

Line 19466 was changed under this assumption that the check was only supposed to catch newer versions of OpenSSL, not LibreSSL.


drwetter:
Thanks again!

Quick check ... I think this one needs to be backported also. Do you mind?

PR #2042, opened 2021-11-15
Check file permissions on ./testssl.sh

dcooper16:
This PR adds a check that ./testssl.sh has both read and execute permission. If ./testssl.sh is lacking execute permission, it will pass the tests in 00_testssl_help.t and 01_testssl_banner.t, which run the program as `bash ./testssl.sh`, but will fail the subsequent tests, which run the program as `./testssl.sh`, but the reason for the failure will not be clear.

I accidentally created a branch in which ./testssl.sh did not have execute permission. That caused the Travis/CI tests to fail, but it took me a long time to discover the reason for the failure.

### DIFF
--- a/t/00_testssl_help.t
+++ b/t/00_testssl_help.t
@@ -4,6 +4,7 @@
 
 use strict;
 use Test::More;
+use File::stat;
 
 my $tests = 0;
 my $fileout="";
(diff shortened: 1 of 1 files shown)


drwetter:
okay & thanks, David!

PR #2043, opened 2021-11-15
Fix some Shellcheck issues

dcooper16:
This PR fixes a few issues found by Shellcheck.

Shellcheck noted that `cipher_pref_check()` was using `ciphers_found` as both a string and an array. It also caught extra ']]' in `run_server_preference()`, and it complained about having expressions of the form `x || y || z` in parenthesis, given that the parenthesis were not necessary.


drwetter:
Thanks, again!

PR #2044, opened 2021-11-22
Fix sub_cipherlists()

dcooper16:
This PR fixes a bug in `sub_cipherlists()` when using sockets to test whether a server supports a set of ciphers with SSLv2. Code that is supposed to extract the list of SSLv2 ciphers to test for is incorrectly reading from the list of non-SSLv2 ciphers.


drwetter:
thanks, David!

PR #2057, opened 2021-12-06
Simplify check for git working tree

mspncp:
Sending the entire log to /dev/null is a rather expensive way of checking whether
we are inside a git working tree. Use `git rev-parse --is-inside-work-tree` instead.


drwetter:
Oh, yes. Good catch! Mind to backport that?

mspncp:
See #2058. I tried to figure out your workflow by looking at the git log. Hope I got it right.

PR #2058, opened 2021-12-06
Simplify check for git working tree in 3.0

mspncp:
Sending the entire log to /dev/null is a rather expensive way of checking whether
we are inside a git working tree. Use `git rev-parse --is-inside-work-tree` instead.

(cherry picked from commit 1e3f9329320028f81e2dfd26b2cb8f01fbba7eb1)


drwetter:
thanks again!

PR #2059, opened 2021-12-07
Simplify extracting Git information

dcooper16:
testssl.sh currently calls `git log --format='%h %ci' -1 2>/dev/null` three times. This PR changes testssl.sh to make this call just once and then use Bash string manipulation to extract the necessary information from the result.


drwetter:
yes, thanks again!

Issue #2053, opened 2021-12-01
[BUG] Invalid json output if there are multiple A records and a test fails


Max7cd:
**Before you open an issue please check this repo whether this is a known problem**
Did not find it.

**Command line / docker command to reproduce**
`testssl.sh --connect-timeout 30s --openssl-timeout 30s --debug 2 --color 0 --jsonfile-pretty /tmp/result.json -p -t smtp mx02.mail.icloud.com.:25`

produced the following or a similar, faulty output multiple times:
```json
{
    "Invocation": "testssl.sh --connect-timeout 30s --openssl-timeout 30s --debug 2 --color 0 --jsonfile-pretty /tmp/result_tls_24E70B5D2838B1D4F36F711012A6C845.json -p -t smtp mx02.mail.icloud.com.:25",
    "at": "f145c430b84d:/app/testssl/bin/openssl.Linux.x86_64",
    "version": "3.1dev ",
    "openssl": "OpenSSL 1.0.2-chacha from Jan 18 17:12:17 2019",
    "startTime": "1638373575",
    "scanResult": [
        {
            "targetHost": "mx02.mail.icloud.com.",
            "ip": "17.42.251.12",
            "port": "25",
            "rDNS": "mx02.mail.icloud.com.",
            "service": "smtp",
            "pretest": [
                {
                    "id": "pre_128cipher",
                    "severity": "INFO",
                    "finding": "No 128 cipher limit bug"
                }
            ],
            "protocols": [
                {
                    "id": "SSLv2",
                    "severity": "OK",
                    "finding": "not offered"
                },
                {
                    "id": "SSLv3",
                    "severity": "OK",
                    "finding": "not offered"
                },
                {
                    "id": "TLS1",
                    "severity": "LOW",
                    "finding": "offered (deprecated)"
                },
                {
                    "id": "TLS1_1",
                    "severity": "LOW",
                    "finding": "offered (deprecated)"
                },
                {
                    "id": "TLS1_2",
                    "severity": "OK",
                    "finding": "offered"
                },
                {
                    "id": "TLS1_3",
                    "severity": "OK",
                    "finding": "offered with final"
                }
            ],
            "grease": [],
            "ciphers": [],
            "fs": [],
            "serverPreferences": [],
            "serverDefaults": [],
            "vulnerabilities": [],
            "cipherTests": [],
            "browserSimulations": [],
            "rating": []
        }
                            {
            "id": "scanProblem",
            "severity": "FATAL",
[...]

PR #2061, opened 2021-12-07
Update reset_hostdepended_vars()

dcooper16:
I noticed in the JSON output shown in https://github.com/drwetter/testssl.sh/issues/2053#issue-1068946357 that two different "scanProblem" error messages appear. While this may not be the reason for it, I discovered that `lets_roll()` was resetting `$NR_SOCKET_FAIL` and `$NR_OSSL_FAIL` after each host was tested, but it was not resetting `$NR_STARTTLS_FAIL`. This led me to compare the variables initialized at the start of testssl.sh to those reset in `reset_hostdepended_vars()` and I discovered a lot a variables were not being reset for each host, but should be.

This PR attempts to fix the problem by adding several variables to `reset_hostdepended_vars()`. For consistency purposes, I moved the resetting of `$NR_SOCKET_FAIL` and `$NR_OSSL_FAIL` from `lets_roll()` to `reset_hostdepended_vars()`. While I tried to find every variable that should be reset for each host, there is a chance that I missed something. In most cases, this PR won't have any effect, since servers with the same DNS name and port number tend to be configured very similarly.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -22965,12 +22965,37 @@ nodeip_to_proper_ip6() {
 
 
 reset_hostdepended_vars() {
+     NR_SOCKET_FAIL=0
+     NR_OSSL_FAIL=0
+     NR_STARTTLS_FAIL=0
+     NR_HEADER_FAIL=0
      TLS_EXTENSIONS=""
      PROTOS_OFFERED=""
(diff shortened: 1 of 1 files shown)


drwetter:
Looks very good, thank you, David!

PR #2063, opened 2021-12-08
Document $keyopts

dcooper16:
Shellcheck complains on line 2234 that keyopts is referenced but not assigned and there is no explanation in the code (or in the documentation) why "$keyopts" is there. This commit adds a comment so that "$keyopts" isn't deleted as part of a code cleanup.


drwetter:
thanks!

PR #2064, opened 2021-12-09
Fix JSON section headers

dcooper16:
This PR fixes two issues with the headers in the structured JSON output. First, if `run_cipher_match()` is performed the header that is used is "pretest" rather than "singleCipher". Second, the headers for "serverPreferences" and "fs" are swapped.


drwetter:
ok, thanks for both PRs!

PR #2072, opened 2021-12-17
Add sieve protocol support when using STARTTLS

k0lter:
This PR add support for the (manage)sieve protocol while using STARTTLS (if the underlying openssl supports it).

### DIFF
--- a/doc/testssl.1.html
+++ b/doc/testssl.1.html
@@ -1,13 +1,13 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
-  <meta name='generator' value='Ronn/v0.7.3 (http://github.com/rtomayko/ronn/tree/0.7.3)'>
+  <meta http-equiv='content-type' content='text/html;charset=utf8'>
+  <meta name='generator' content='Ronn-NG/v0.9.1 (http://github.com/apjanke/ronn-ng/tree/0.9.1)'>
   <title>testssl(1)</title>
   <style type='text/css' media='all'>
   /* style: man */
   body#manpage {margin:0}
-  .mp {padding:0 9ex 1ex 4ex}
+  .mp {max-width:100ex;padding:0 9ex 1ex 4ex}
   .mp p,.mp pre,.mp ul,.mp ol,.mp dl {margin:0 0 20px 0}
   .mp h2 {margin:10px 0 0 0}
   .mp > p,.mp > pre,.mp > ul,.mp > ol,.mp > dl {margin-left:8ex}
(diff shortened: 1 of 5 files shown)


drwetter:
Thank you! 

Basically a good idea, we need to extend that a bit though.

* Is there any public server you know which we can use for unit testing?
* if not: is there any server you know about where I can privately test it myself?
* ``prepare_debug()`` misses a HAS_SIEVE line
* the "reasonable port guess function" in ``ports2starttls()``: can that be amended?
* documentation is missing in ~/doc/

drwetter:
désolé, pushed da wrong button (David Guetta)

k0lter:
I've updated the PR to take into account your proposals:

- Unit test added with a known public imap server which supports (manage)sieve through STARTTLS
- `ports2starttls()` was updated
- doc (markdown, manpage and html) also updated
- `HAS_SIEVE` was already present in `prepare_debug()` so no changes

Note: the OpenSSL binary present in `bin/` (at least the Linux amd64 one) directory seems to be based on OpenSSL 1.0.2 but the sieve protocol support with STARTTLS was added in OpenSSL version 1.1.1

drwetter:
Well done, thanks a lot! 

I would like to merge this. However the roff-formatted man page and the HTML introduce more changes than just adding the word sieve. Probably this is due to the use of ronn-ng which is admittedly better. 

But despite that: let's move forward. There's one thing  though which each version of ronn doesn't do good: It limits the width for HTML to 100 pixels . If you could just change that?

doc/testssl.1.html
```diff
@@ -1,13 +1,13 @@
...
-  <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
-  <meta name='generator' value='Ronn/v0.7.3 (http://github.com/rtomayko/ronn/tree/0.7.3)'>
+  <meta http-equiv='content-type' content='text/html;charset=utf8'>
+  <meta name='generator' content='Ronn-NG/v0.9.1 (http://github.com/apjanke/ronn-ng/tree/0.9.1)'>
   <title>testssl(1)</title>
   <style type='text/css' media='all'>
   /* style: man */
   body#manpage {margin:0}
-  .mp {padding:0 9ex 1ex 4ex}
+  .mp {max-width:100ex;padding:0 9ex 1ex 4ex}
```
drwetter, line 10: This line.

k0lter: I was about to fix this but I found some inconsistencies in doc so I've submitted the PR #2074.

Could you take a look and if you merge it I could update this PR accordingly.

drwetter:
Thanks a lot again!

Your other PR is now colliding with this one though.

doc/testssl.1.html
drwetter: FYI: I fixed this one manually now in another PR

PR #2074, opened 2021-12-23
Various fixes in doc

k0lter:
This PR includes various fixes in Markdown doc (and update manpage and html accordingly)

- Missing carriage return (some markdown parts were not properly converted in roff or html)
- Missing or extra numerous `<code>` backticks


drwetter:
Hi @k0lter ,

thanks. Did you do that manually? 

If so there's rather the general question how to do this in the future. Up to now I used the previous version of ronn, the output was ok using the interpreters (man, firefox or chrome) also when the roff and HTML code weren't good.

k0lter:
Hi @drwetter

I used ronn 0.9.1 ([patched to handle properly level 4 titles](https://github.com/apjanke/ronn-ng/commit/e10a69c18782bcb1e6918ace284e712b3f96be2b)) but that's true that I had to fix manually the multi-lines code block rendering for both nroff and html (but with `git add --patched` that was straightforward).

I also fixed manually the css to avoid the fixed width as you already suggested.

drwetter:
@k0lter , yes, I saw the version you were using in the comments of the generated files.

Markdown is the primary format, so any improvements are welcome and a no-brainer. For the rest, see the following.

My question was more about maintainability: an automated generation with a well defined tool is as far as maintainability concerned always better. A manual patching is ok of roff or HTML, if it's documented what has been done and it's worth the effort. By "worth the effort" I mean a) your work and b) the result when watching the roff page or HTML.

For a) if a future PR generates new man pages in roff and HTML any manual work of you will be lost.  For b) The result is mostly what counts IMO. With ronn version 0.7 the resulted roff/HTML code was everything else than great or compliant. For me(TM) it displayed without error and that sufficed.

I know that ronn is not a perfect tool but aiming for perfecting likely won't pay off.

Bottom line: Keep please a) and b) in mind when continuing on this PR. After merging your previous one this needs be amended to resolve the conflict

k0lter:
I've updated the PR with a proposal to switch from ronn to pandoc to build the doc. I've also added a Makefile so that it should be a) and b) compliant. I've also removed the manpage and html from the versioning as they are now easily reproducible.

I let you take a look.

drwetter:
Thanks. I am little bit confused though that all of a sudden html and roff output was removed. This needs to be reverted. 

And you used pandoc.

Maybe it's the right way to use pandoc instead, can't tell yet. But if you do propose such a change I would have expected a little bit of discussion beforehand or some accompanying arguments in the PR why we should switch to pandoc. Just a statement "oh, I switched to pandoc" isn't enough.

k0lter:
Hi @drwetter

As I said this is just a proposal (switch from ronn to pandoc). The idea was just to use a tool that works better than ronn: the latest release has some bugs and installing an old version is not very handy to those who wants to contribute. Moreover, in another PR you asked me to manually change the css generated in the html output.
Use a Makefile and Pandoc here permits to address these issues and ease the contributions. To be honest before making this PR I didn't know about ronn and I had to discover by myself how the roff and html were generated (that was not hard but not obvious either).

I removed the roff and html output because I thought they were added to versioning to circumvent the ronn issues and to ease the contributions. I could add them again.

More than switching to pandoc, this PR also fixes various issues in the markdown.

drwetter:
@k0lter : Don't get me wrong. I appreciate collaboration but you needed just to explain better why you think why we should use A instead of B.

I looked at pandoc a while back and ended up using ronn. IIRC it was because was requiring too much dependencies. Let me have a closer look during the next days.

Thx, Dirk

drwetter:
Hi @k0lter ,

I suggest this to be split:

1) correct errors In markdown, ROFF and HTML (keep all three formats)
2) provide the Makefile + template for pandoc (again keep all three formats)

Probably with 1) it's good to wait until #2088 has been merged.

Idea behind it: everyone is free to use the converter of choice from the primary format (md). This repo should always contain all formats so that users don't have to worry about the conversion. 

Whether this repo will use ronn or pandoc and how we generate the manpages: time will tell. Pandoc sounds like a good idea as it _seems_ better -- and if it is: maybe a github action can generate HTML and ROFF.

Thx, Dirk

Issue #2079, opened 2022-01-04
[BUG / possible BUG] openssl binary/binaries that default to :4433 timeout after b6b5a67


rvstaveren:
After b6b5a67 which removed DNS timeouts for WSL I see with the FreeBSD binary long scan times.

Upon investigation I discovered that openssl s_client for FreeBSD now uses the defaults, which is not :0 but :4433. This might happen on other platforms too

**Command line / docker command to reproduce**
Under FreeBSD, using 3.1dev at 35ddd91 it takes around 12 minutes to complete a scan
```

./testssl.sh www.verweg.com

###########################################################                     
    testssl.sh       3.1dev from https://testssl.sh/dev/                        
    (35ddd91 2021-12-21 10:54:58 -- )                                           
                                                                                
      This program is free software. Distribution and                           
             modification under GPLv2 permitted.                                
      USAGE w/o ANY WARRANTY. USE IT AT YOUR OWN RISK!                          
                                                                                
       Please file bugs @ https://testssl.sh/bugs/                              
                                                                                
###########################################################                     
                                                                                
 Using "OpenSSL 1.0.2-chacha (1.0.2k-dev)" [~183 ciphers]                       
 on helium:./bin/openssl.FreeBSD.amd64                                          
 (built: "Jan 18 18:46:10 2019", platform: "BSD-x86_64")                        


...
 Done 2022-01-04 17:19:53 [0679s] -->> 94.142.245.8:443 (www.verweg.com) <<--
```
Most time is spend in detecting `openssl s_client` options where no `-connect` is specified and the command tries to connect to :4433

**Expected behavior**

I've kind of fixed in my fork partially undoing b6b5a67, but using `-connect 0:0` where applicable.

https://github.com/rvstaveren/testssl.sh/commit/d82085d35b7c6dc74e2742dd7bae47e985cfa32f

Maybe reverting  b6b5a67 but substituting `NXCONNECT=${NXCONNECT:-invalid.}` with `NXCONNECT=${NXCONNECT:-0:0}` would be good too.

With this, under FreeBSD, completion of a scan takes around a normal ±80 seconds

```

./testssl.sh www.verweg.com
###########################################################
      testssl.sh       3.1dev from https://testssl.sh/dev/
                            (d82085d 2022-01-01 12:20:21 -- )                                           
                                                                                
      This program is free software. Distribution and                           
             modification under GPLv2 permitted. 
[...]

PR #2081, opened 2022-01-06
Send /dev/null to s_client

dcooper16:
As discussed in #2079, this PR adds `</dev/null` to calls `$OPENSSL s_client` that are intended to test whether a certain option is supported. This is done to prevent hanging in the case that the option is supported, a TLS server happens to be listening on the port to which s_client tries to connect, and the connection is successful.

This PR also adds a new helper function, `sclient_supported()`, which is called from `locally_supported()` and `run_prototest_openssl()`. The helper function makes use of the already defined variables `$HAS_SSL2`, `$HAS_SSL3`, and `$HAS_TLS13` in order to avoid calling `$OPENSSL s_client` when it has already been determined whether $OPENSSL supports the specified protocol.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -5055,17 +5055,38 @@ run_client_simulation() {
      return $ret
 }
 
+# generic function whether $1 is supported by s_client.
+# Currently only used for protocols that's why we saved -connect $NXCONNECT. 
+sclient_supported() {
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks for both PRs, David!

Issue #1258, opened 2019-05-03
STARTTLS: provide socket functions for <del>LDAP, </del> IRC


drwetter:
Also maybe reconsider in the meantime whether ``--ssl-native`` should be enforced as for those the procol is barely usable (``FIXME: LDAP+STARTTLS over sockets not supported yet`` in ``fd_sockets()``).

Including unit tests for IRC and LDAP via sockets.

bb5450e3f5f83d9c560e5293c692a8d03d02523f would need partly to be reverted when sockets will be available for LDAP.

PR #2088, opened 2022-01-24
Add prototype for STARTTLS+ LDAP via sockets

drwetter:
See #1258

To do:
* <del>more robustness. At least the success value from the response need to be retrieved and checked via starttls_io().</del>
* <del>double check the pre-handshake before the OID whether it's correct for every case</del>
* <del>documentation</del>
* <del>inline help</del>
* <del>amend CI</del>

It seems to work though against db.debian.org


drwetter:
If anybody has more knowledge than the little I have regarding LDAP + STARTTLS you're welcome to comment or amend the PR.  This PR as of now results from reverse engineering by inspecting the network traffic and trying to match it roughly with the RFC mentioned.

Issue #2103, opened 2022-02-15
Docker container is not working properly. It returns an "/tmp/testssl_output.json" that breaks my Jenkins job when I try to use it to create my Jenkins Test Results


snigel-nelson:
## Description of the Issue

I have a Jenkins job that uses the official _testssl.sh_ docker image (https://hub.docker.com/r/drwetter/testssl.sh) to run a container to verify several domains. This container creates a JSON file, _"/tmp/testssl_output.json"_, which I use to create my Jenkins test results report to be available for consulting.
The problem is that on the last build my Jenkins job downloaded a newer Docker image which produces invalid JSON which in turn breaks the job build causing it to fail. You can see this in the following image:

![image](https://user-images.githubusercontent.com/91899315/154134587-766da413-b6a2-40fd-af8c-8a2a2a6799bd.png)

And it fails here:

![image](https://user-images.githubusercontent.com/91899315/154134671-6c681be6-ec53-488d-bba8-d066536eeacd.png)

## Test results
As I mentioned in the description I'm using the JSON file produced by the container and stored at  _"/tmp/testssl_output.json"_ to create the XML file, with the standard format of JUnit tests, in the _./logs/${tldDomainName}.xml_ location, and allow Jenkins to read and create the Test Results dashboard to report as represented in the image below:
![image](https://user-images.githubusercontent.com/91899315/154136351-b63357b5-3364-42f9-b842-a832ac888a15.png)

The problem is that this new Docker image produces invalid JSON which causes the Jenkins Job to break when it tries to convert the JSON -> XML.  You can verify the structure of the new JSON produced by this new version of testssl.sh Docker image in the attached file:
[testssl_output.txt](https://github.com/drwetter/testssl.sh/files/8074380/testssl_output.txt)

## Command line / docker command to reproduce
These are the commands that I'm using in my Jenkins job to verify the services and create the Test Results dashboard:

`
  String tldDomainName = domain.split('/')[0]; // Get top level domain name
  sh 'docker rm testssl || exit 0' // remove container before running
  sh "docker run -i --name testssl -v ${WORKSPACE}/logs:/home/logs -e DAYS2WARN1='30' -e DAYS2WARN2='20' drwetter/testssl.sh:${params.tag} --jsonfile-pretty=/tmp/testssl_output.json --color=0 ${domain}"
  sh "docker cp testssl:/tmp/testssl_output.json ./logs/${tldDomainName}.json"
[...]

PR #2104, opened 2022-02-16
Fix JSON output bc of missing locale in alpine (3.0)

drwetter:
It is now being tested whether the binary locale exists and
there's a global introduced for that.

Also there's no fileout warning at this early stage anymore
as it leads to non-valid JSON


drwetter:
fixes #2103

PR #2114, opened 2022-03-02
Add support for TN3270/telnet STARTTLS.

enxio:
Modeled after what 1.1.x OpenSSL does in apps/s_client.c for PROTO_TELNET. Adding tests is bit tricky, as I am not aware of any public TN3270 servers with STARTTLS.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -11168,12 +11168,13 @@ starttls_telnet_dialog() {
      "
      local tnres=""
      local ret=""
+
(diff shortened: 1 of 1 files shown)


drwetter:
Cool, thanks @enxio !

Could you do two things to be consistent with the style, pls? 

* add a CR before L11169
* L11176:  better use two square brackets, like  [[ ]] && [[ ]] 

Cheers, Dirk

enxio:
I hope that's better. Thanks for your consideration.

drwetter:
Thanks a lot!

PR #2115, opened 2022-03-02
Update docs for #2114

enxio:
Update the docs with regard to #2114. I assume html and man are generated from the markdown.


drwetter:
Would it be possible to edit the html and roff file accordingly?

enxio:
Here they are. Hope that's what's needed.

drwetter:
Looks good , thanks again!

PR #2117, opened 2022-03-08
Fix sclient_auth

dcooper16:
If `$connect_success` is false, then `sclient_auth()` does not "return" any value, and the calling function treats this as if `sclient_auth()` had returned 0.

This PR fixes `sclient_auth()` so that 1 is returned if `$client_success` is false.


drwetter:
Good catch, thanks @dcooper16 !

PR #2119, opened 2022-03-09
Include RSA-PSS in ClientHello

dcooper16:
This PR changes `prepare_tls_clienthello()` in the 3.0 branch so that the RSA-PSS algorithms are offered in the signature algorithms extension of TLS 1.2 and below ClientHello messages.


drwetter:
Thanks, David!

Issue #2116, opened 2022-03-08
doesn't seem to be a TLS/SSL enabled server....but site comes up fine in browsers


tonyguadagno:
hi, when i run your script against my site...tickets.fujifilmusa.com:443 i get this
```
172.19.5.46:443 doesn't seem to be a TLS/SSL enabled server
 The results might look ok but they could be nonsense. Really proceed ? ("yes" to continue) --> yes
 Service detected:       Couldn't determine what's running on port 443, assuming no HTTP service => skipping all HTTP checks


 Testing protocols via sockets except NPN+ALPN

 SSLv2      not offered (OK)
 SSLv3      not offered (OK)
 TLS 1      not offered
 TLS 1.1    not offered
 TLS 1.2    not offered
 TLS 1.3    not offered

```

however the site comes up fine in all browsers....also, i found this other testing script ([https://github.com/rbsec/sslscan/](https://github.com/rbsec/sslscan/)) and it reports the site as fine.

so, is there a problem with my site or with the script?

thanks for your efforts on this!

drwetter:
Hi @tonyguadagno,

thanks for reporting. It's a bug by our script.

Stay tuned.

Cheers, Dirk

tonyguadagno:
h, no problem...btw, [SSLLabs](https://www.ssllabs.com) is also choking on the site....so i am not ruling out a bad config on my side...but i am just not seeing what i might be doing wrong.
if you see anything wrong on my site, please let me know
thanks

drwetter:
``openssl s_client -tls1_2 -connect tickets.fujifilmusa.com:443`` works. We do some handshakes in sockets (i.e. w/o openssl) before. Those fail, returning a TLS alert.

Maybe @dcooper16 can shed some light in this?

dcooper16:
Hi @tonyguadagno, @drwetter,

I believe that I discovered the issue. It seems that tickets.fujifilmusa.com only works if the Signature Algorithms extension includes the recently-defined RSA-PSS algorithms. At the moment, testssl.sh only includes these algorithms in a TLS 1.3 ClientHello. It seems, however, that OpenSSL includes them in a TLS 1.2 ClientHello, so I guess it's okay for us to modify testssl.sh to also include them in the TLS 1.2 and below ClientHello.

[...]

PR #2120, opened 2022-03-09
Add FFDHE groups to supported_groups

dcooper16:
There seems to be a bug in the code behind tickets.fujifilmusa.com (#2116). With a TLS 1.2 ClientHello, this server will not negotiate a TLS_DHE_* cipher suite if the supported_groups extension is present, but does not include an FFDHE group. The result is that the TLS_DHE_* cipher suites are missed by `cipher_pref_check()` and `ciphers_by_strength()`.

The server's behavior seems to be inconsistent with RFC 7919, but adding FFDHE groups to the supported_groups extension in a TLS 1.2 and below ClientHello doesn't seem to cause problems for other servers. So, this PR adds the FFDHE groups that are currently in supported_groups extension for the TLS 1.3 ClientHello to the TLS 1.2 and earlier ClientHello.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -15025,11 +15025,11 @@ prepare_tls_clienthello() {
                # Supported Groups Extension
                extension_supported_groups="
                00, 0a,                    # Type: Supported Elliptic Curves , see RFC 4492
-               00, 3e, 00, 3c,            # lengths
+               00, 42, 00, 40,            # lengths
                00, 0e, 00, 0d, 00, 19, 00, 1c, 00, 1e, 00, 0b, 00, 0c, 00, 1b,
                00, 18, 00, 09, 00, 0a, 00, 1a, 00, 16, 00, 17, 00, 1d, 00, 08,
                00, 06, 00, 07, 00, 14, 00, 15, 00, 04, 00, 05, 00, 12, 00, 13,
-               00, 01, 00, 02, 00, 03, 00, 0f, 00, 10, 00, 11"
+               00, 01, 00, 02, 00, 03, 00, 0f, 00, 10, 00, 11, 01, 00, 01, 01"
           elif [[ 0x$tls_low_byte -gt 0x03 ]]; then
                # Supported Groups Extension
                if [[ ! "$process_full" =~ all ]] || ( "$HAS_X25519" && "$HAS_X448" ); then


tonyguadagno:
@dcooper16  so it sounds like i have got a problem on my end as well?  is that right?

drwetter:
@tonyguadagno: mind to divulge the Tomcat version and the SSL/TLS library used?

dcooper16:
> @dcooper16 so it sounds like i have got a problem on my end as well? is that right?

Yes, I believe so. However, it seems unlikely to cause problems in practice, since the TLS_DHE_* cipher suites are not very widely used. I would guess that most clients will negotiate one of the TLS_ECDHE_* cipher suites. Clients that don't support any TLS_ECDHE_* cipher suite would be unlikely to include a supported_groups extension in their ClientHello.

There may be some clients that support some TLS_ECDHE_* cipher suites (and so include a supported_groups extension), but don't support any of the TLS_ECDHE_* cipher suites offered by your server. If these clients do support any of the TLS_DHE_* cipher suites that you server supports, then the connection attempt would fail when it should have succeeded. However, this seems to be an unlikely scenario.

tonyguadagno:
@drwetter is it ok if i DM you that info?

drwetter:
Thanks @dcooper16 !

@tonyguadagno : Of course! Just look at the output of ``grep SWCONTACT testssl.sh``

tonyguadagno:
@drwetter i guess the joke is on me, because i could not find your email addr....so i wrote up a description and you can download it here.  i will delete when you are done downloading
https://ffcpftp.fujifilmusa.com/WebClientNew/index.html/?ShareToken=46A73288768FD10079A18656974D92843FC79AC9

drwetter:
Thanks, got that @tonyguadagno . Can remove that!

(and comment later on that)

tonyguadagno:
@drwetter I finally got around to running the new script against my local openssl and it looks a lot better.  btw, it claims that edge would now negotiate any ciphers...is that the OLD edge or new chromium based edge...because i am running new edge (Version 99.0.1150.30 (Official build) (64-bit)) and it seems fine.

i am not sure how to attack the problem on my side....

drwetter:
WRT the client siumulation. It's an older version of Edge we use. We probably need an update. My humble guess is that maybe newer Edge browsers take understand RSA-PSS.

To my knowledge Edge was always Edge with Webkit from Chrome. Not 100% sure though.

\<offtopic\>Feel free to use https://github.com/drwetter/automate_tomcat \</offtopic\>

tonyguadagno:
@drwetter automate_tomcat looks cool, i will check it out...btw, i  just upgraded to the most current of that major version.

as for the original issue, it seems like i am not configuring something properly to cause this to happen...given my environment, do you know where i should look?  is this a bouncycastle issue?

drwetter:
@dcooper16 sent me some input but I won't appear to take credit for that, so maybe he can respond.

dcooper16:
What I sent @drwetter was related to this PR's issue. I found https://github.com/bcgit/bc-java/issues/718 reporting the problem with the way that Bouncy Castle handles FFDHE cipher suites when the ClientHello includes a supported_groups extension that doesn't specify any FFDHE groups.

I am not familiar with Bouncy Castle, so have no idea how to configure it so that its TLS server will use non-PSS signature algorithms.

PR #2125, opened 2022-03-16
Fix CI error

dcooper16:
I found a working NNTP server on http://vivil.free.fr/nntpeng.htm. I'm not sure if it's a reliable server to test against, but it at least temporarily fixes the error in CI testing.


drwetter:
Cool, David! Thanks.

I thought to just comment this out later, but this sems better.

PR #2129, opened 2022-03-24
Reorder output of run_server_preference()

dcooper16:
This PR reorders the output of `run_server_preference()` as discussed in https://github.com/drwetter/testssl.sh/issues/1311#issuecomment-633501128.

This is the first step towards addressing #1311. I have not yet started working on any additional work needed for #1311.


drwetter:
Ok, thanks @dcooper16 !

One has to get use to it but it's better I believe.

I tried to put protocol and cipher into one line but it looked too squeezed IMHO

image

PR #2130, opened 2022-03-24
Use tls_sockets() in run_tls_fallback_scsv()

dcooper16:
This PR adds the use of `tls_sockets()` to `run_tls_fallback_scsv()` to perform testing when the `--ssl-native` flag is not used. With this commit, `run_tls_fallback_scsv()` only uses `tls_sockets()` instead of `$OPENSSL` if the ClientHello needs to include the TLS_FALLBACK_SCSV flag, but it is not supported by `$OPENSSL`, or if the protocol that would be negotiated is SSLv3 and `$OPENSSL` does not support SSLv3.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -13599,6 +13599,7 @@ parse_tls_serverhello() {
      fi
 
      if [[ $tls_alert_ascii_len -gt 0 ]]; then
+          echo "CONNECTED(00000003)" > $TMPFILE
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks a lot, @dcooper16 !

drwetter:
Probably that's due to the editor but sometimes I see inserted spaces which my editor complains about

image

No need to issue a PR for this now.

PR #2133, opened 2022-04-01
Implement fixes in documentation from #2074

drwetter:
kudos @k0lter

* numbering
* some ticks / backticks

For now I left the html and roff files like they were, i.e. only fixed. Pandoc should be reconsidered later.


drwetter:
don't know what the problem is with badssl.com? Doesn't return **any** DNS record

drwetter:
https://github.com/chromium/badssl.com/issues/500

PR #2139, opened 2022-04-05
Fix #2131

### LINKED ISSUE #2131: [BUG / possible BUG] Incorrect PFS output on servers with only brainpool elliptic curves
If a server only accepts brainpool curves (e.g. via "SSLOpenSSLConfCmd Curves brainpoolP256r1" in an Apache2 vhost definition), testssl will say "No ciphers supporting Forward Secrecy offered" when checking for PFS (i.e. testsssl -f) even though PFS ciphers are available and "openssl s_client -connect ip:port -curves brainpoolP256r1 -tls1_2" is able to connect using those ciphers.
Also running with "--ssl-native" provides the correct output.
I only tested this on local installations so unfortunately there is no target IP I could provide.

testssl -f ip:port

**Expected behavior**
Testssl should list all suported PFS ciphers with elliptic curve brainpoolP256r1

 - OS: Kali GNU/Linux Rolling
 - Platform: Linux 5.14.0-kali4-amd64 x86_64
 - Version: testssl 3.0.7 from https://testssl.sh/

**Additional context**
I did a bit of debugging myself and to me it seems like there is a logic issue in run_pfs().
In line 9627 (this is version 3.0.7) there is a check if tls 1.3 is offered. If not, tls_sockets is called with version "04" which I understand is tls1.3 again. I changed this value to "03" in a local copy and got the expected result. I'm not sure though if that is actually the solution or if I just broke several other use cases by doing this.

dcooper16:
This PR fixes #2131 by having `run_fs()` attempt a TLS 1.2 ClientHello if the initial TLS 1.3 ClientHello fails. The TLS 1.2 ClientHello will offer many more curves than the TLS 1.3 ClientHello offers, and so it may succeed if the server supports ECDHE ciphers, but only with curves that were removed by RFC 8446.

While this PR should fix #2131 when `tls_sockets()` is used, it is only a partial fix when testssl.sh is run in `--ssl-native` mode. If the server supports both DHE and ECDHE ciphers and OpenSSL 1.1.1 or newer is used, and only non-RFC 8446 curves are supported with ECDHE, then the ECDHE ciphers will be missed.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -10323,28 +10323,26 @@ run_fs() {
                fileout "$jsonID" "WARN" "tests skipped as you only have $nr_supported_ciphers FS ciphers on the client site. ($CLIENT_MIN_FS are required)"
                return 1
           fi
+          # By default, OpenSSL 1.1.1 and above only include a few curves in the ClientHello, so in order
+          # to test all curves, the -curves option must be added. In addition, OpenSSL limits the number of
+          # curves that can be specified to 28. So, if more than 28 curves are supported, then the curves must
+          # be tested in batches.
+          curves_list1="$(strip_trailing_space "$(strip_leading_space "$OSSL_SUPPORTED_CURVES")")"
+          curves_list1="${curves_list1//  / }"
+          if [[ "$(count_words "$OSSL_SUPPORTED_CURVES")" -gt 28 ]]; then
+               # Place the first 28 supported curves in curves_list1 and the remainder in curves_list2.
+               curves_list2="${curves_list1#* * * * * * * * * * * * * * * * * * * * * * * * * * * * }"
+               curves_list1="${curves_list1%$curves_list2}"
+               curves_list1="$(strip_trailing_space "$curves_list1")"
+               curves_list2="${curves_list2// /:}"
+          fi
+          curves_list1="${curves_list1// /:}"
           $OPENSSL s_client $(s_client_options "-cipher $fs_cipher_list -ciphersuites ALL $STARTTLS $BUGS -connect $NODEIP:$PORT $PROXY $SNI") >$TMPFILE 2>$ERRFILE </dev/null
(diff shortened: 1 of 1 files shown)


dcooper16:
I added to this PR so that it also works in `--ssl-native` mode.

drwetter:
Hi David.

can you explain ``curves_list2="${curves_list1#* * * * * * * * * * * * * * * * * * * * * * * * * * * * }"``  ?

Thx, Dirk

dcooper16:
> can you explain `curves_list2="${curves_list1#* * * * * * * * * * * * * * * * * * * * * * * * * * * * }"` ?

Here is the code in context:
```
          curves_list1="$(strip_trailing_space "$(strip_leading_space "$OSSL_SUPPORTED_CURVES")")"
          curves_list1="${curves_list1//  / }"
          if [[ "$(count_words "$OSSL_SUPPORTED_CURVES")" -gt 28 ]]; then
               # Place the first 28 supported curves in curves_list1 and the remainder in curves_list2.
               curves_list2="${curves_list1#* * * * * * * * * * * * * * * * * * * * * * * * * * * * }"
               curves_list1="${curves_list1%$curves_list2}"
               curves_list1="$(strip_trailing_space "$curves_list1")"
               curves_list2="${curves_list2// /:}"
          fi
          curves_list1="${curves_list1// /:}"
```
curves_list1 is set to the list of OpenSSL supported curves, with no leading or trailing spaces, and with exactly one space between each curve. If there are more than 28 curves in the list, then the line you asked about sets curves_list2 to curves 29 through the end. The next line then removes these curves from curves_list1, so that curves_list1 contains the first 28 curves in the list.

The line works by removing from curves_list1 the shortest prefix that contains 28 spaces, as that prefix will also include 28 curve names.

drwetter:
Thanks, David!

I could be better to put explanations like this into the code.

PR #2144, opened 2022-04-07
 OpenSSL compatibility fix

dcooper16:
OpenSSL 3.0.X uses different names for some elliptic cures in the "Server Temp Key" line than previous previous versions. This PR addresses this issue by checking for both names.


drwetter:
Thanks!

PR #2145, opened 2022-04-11
More OpenSSL compatibility fixes

dcooper16:
This PR fixes two more issues with using OpenSSL 3.X. When $OPENSSL is used to obtain a fingerprint, OpenSSL 3.X prepends the fingerprint with "sha1" or "sha256" rather than "SHA1" or "SHA256". In addition, the way that OpenSSL 3.X writes distinguished names causes a space character to appear at the beginning of "$cn" and "$issuer_CN" in `certificate_info()`.


drwetter:
Thanks, David!

PR #2146, opened 2022-04-11
More OpenSSL compatibility fixes

dcooper16:
This PR fixes yet another issue with using OpenSSL 3.X with the 3.0 branch. When $OPENSSL is used to obtain a fingerprint, OpenSSL 3.X prepends the fingerprint with "sha1" or "sha256" rather than "SHA1" or "SHA256".


drwetter:
thx again

PR #2150, opened 2022-04-13
Fix setting NO_SESSION_ID

dcooper16:
With a TLS 1.3 connection, a session ID will only appears as part of a post-handshake session ticket. However, with OpenSSL 1.1.1 or newer when using `$OPENSSL s_client` as it is called in `determine_optimal_proto()` (i.e., with "< /dev/null"), a post-handshake session ticket will usually not be received, even if the server supports it. With versions of LibreSSL that support TLS 1.3, a post-handshake session ticket is never displayed (even without "< /dev/null"). This can result in NO_SESSION_ID incorrectly being set to true.

This PR fixes the issue by setting NO_SESSION_ID to true by default, and then setting it to false if a session ID is returned by any connection to the server.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -284,7 +284,7 @@ ERRFILE=""
 CLIENT_AUTH="none"
 CLIENT_AUTH_CA_LIST=""
 TLS_TICKETS=false
-NO_SSL_SESSIONID=false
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, David!

Shouldn't this affect 3.0 too?

PR #2154, opened 2022-04-14
Fix run_server_preference() with no default protocol

dcooper16:
`run_server_preference()` calls `default_proto=$(get_protocol $TMPFILE)` even if all attempts to connect to the server failed. This will result in `default_proto` incorrectly being set to TLS 1.2. This PR fixes the issue by only calling `get_protocol()` if an attempt to connect to the server was successful.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -6671,7 +6671,7 @@ run_server_preference() {
                fi
           fi
      fi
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks again, David!

PR #2156, opened 2022-04-14
Fix run_server_defaults()

dcooper16:
PR #1960 created a bug by placing code between the call to `determine_tls_extensions()` and a check of its return code. This PR fixes the problem.


drwetter:
Thanks, David!

PR #2162, opened 2022-04-20
Fix calls to $OPENSSL ciphers

dcooper16:
This commit fixes testssl.sh's calls to the `$OPENSSL ciphers` command.

The main issue it fixes is when `actually_supported_osslciphers()` is called to get a list of non-SSLv2 ciphers supported by `$OPENSSL`. With OpenSSL 1.0.2, the `-tls1` option needs to be used to exclude SSLv2 ciphers. With LibreSSL, the `-tls1` option may be provided, but it has no effect. With OpenSSL 1.1.1 and newer, the `-tls1` option causes TLSv1.2-only ciphers (e.g., AES256-SHA256) to be excluded (when the `-s` option is also used).

This commit fixes the problem by allowing `-no_ssl2` to be provided as an option to `actually_supported_osslciphers()`. For versions of `$OPENSSL` that support SSLv2, `-no_ssl2` is replaced by `-tls1`. For versions of `$OPENSSL` that do not support SSLv2, `-no_ssl2` is simply removed.

This commit also changes `openssl2hexcode()` to include the `-tls1` option when `$OPENSSL` supports SSLv2, since `openssl2hexcode()` should only return a non-SSLv2 cipher.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -924,15 +924,23 @@ join_by() {
 actually_supported_osslciphers() {
      local ciphers="$1"
      local tls13_ciphers="$TLS13_OSSL_CIPHERS"
+     local options="$3 "
 
      [[ "$2" != ALL ]] && tls13_ciphers="$2"
      "$HAS_SECLEVEL" && [[ -n "$ciphers" ]] && ciphers="@SECLEVEL=0:$1"
-     if "$HAS_CIPHERSUITES"; then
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks David for patching this. I can't really tell and my confidence is in you is very high ;-)
 
How about 3.0?

PR #2180, opened 2022-05-09
Check for OpenSSL support for ffdhe groups

dcooper16:
OpenSSL 3.0.0 and later supports specifying the FFDHE groups from RFC 7919 in the "-groups" (or "-curves") option of s_client.

This PR modifies `find_openssl_binary()` to check whether $OPENSSL supports this. This information is then used by `run_client_simulation()`, if client simulation testing is being performed using $OPENSSL. If the "curves" for a client includes FFDHE groups, then they will be included in the simulated ClientHello.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -19554,7 +19554,7 @@ find_openssl_binary() {
      local openssl_location cwd=""
      local ossl_wo_dev_info
      local curve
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, David!

PR #2181, opened 2022-05-09
Continue run_server_preference() on error

dcooper16:
Currently `run_server_preference()` will stop if it cannot determine whether the server enforces a cipher preference order.

This PR changes `run_server_preference()` so it will continue running even if this case, so that the list of ciphers supported with each TLS protocol is provided. Since it is not known whether the server enforces a cipher order, the list of supported ciphers is ordered by strength.


drwetter:
Thank you, David!

Issue #2182, opened 2022-05-11
[Feature request] Improve DNS_CAArecord finding description in the fileoutput


manuel-sommer:
**Which version are you referring to**
3.1dev

**Describe your feature request (if it's a technical feature)**
The output of the finding "DNS_CAArecord" is not descriptive in the fileoutput

**Describe the solution you'd like**
A clear and concise description of what is the problem regarding this vulnerability would help.

Issue #966, opened 2018-01-19
PrettyJSON improvement suggestions


AresS31:
Hi,

I am working a lot with `testssl` prettyJSON outputs, actually I made a script to parse the output and automatically generate Excel spreadsheets from the prettyJSON outputs (see https://github.com/AresS31/testssl2xlsx).

When developing that script I had to handle a few things that I think should be handled by `testssl`. 

The following are my improvement suggestions:

- Unique and more explicit `ids` for `cipherTests` objects. Insead of `secure_renego` use `Secure Renegotiation` for example, and all `ids` should be consistent, either all in lowercase or all in uppercase (even for acronyms). Also `ids` should not be too similar to each others because as you can see on the following screenshots, there are 2 objects with the same `id` of `drown`; `logjam` and `LOGJAM_common primes` both contain the string/sub-string `logjam` from a coding point of view this create the need to add additional checks and line of codes to ensure that the right object is being processed.

![image](https://user-images.githubusercontent.com/11601622/35170998-157729cc-fd5a-11e7-8253-412866b7b74f.png)
![image](https://user-images.githubusercontent.com/11601622/35171022-2d27aea2-fd5a-11e7-97f1-e27f7e25c441.png)

- Instead of using the value `OK` as severity when the host:port is not vulnerable instead use `NOT VULNERABLE` to add more clarity and avoid any confusion.

- For the `finding` field here too adding more consistency, see the following screenshot, for `lucky13` the finding does not start with `lucky13` whereas for `rc4` it starts with `RC4: not`. So more consistency and maybe more verbose in the explanation, something that could directly be used as a remediation.

![image](https://user-images.githubusercontent.com/11601622/35171541-0a80bbbc-fd5c-11e7-8116-a051bbb5d395.png)

- For the `protocols` section the `ids` for the protocols should be more consistent as well like we have got `sslv2`, `sslv3` but then `tls1`, `tls1_1` why not replacing them by `tlsv1.0`, `tlsv1.1`, etc.

- Adding an additional field to the `protocols` section, it could be `enabled: YES` and `enabled: NO` or `offered: YES` and `offered: NO` so we could programmatically just test the value of this `JSON` field to determine if a protocol is being offered rather than working on the finding field and check if it contains the string `not offered`.

This are my suggestions they are all relevant to my case but it is up to you if you judge them relevant to a wider use.

Great job!
Alex

drwetter:
[...]

Issue #194, opened 2015-09-19
providing fixing documentation and/or links to helpful urls


OliverLeitner:
hi guys

very nice toy you got there...

ive been playing with it, the one thing im missing though is good documentation links on howto mitigate those attacks from your website (https://testssl.sh)

i.e. searching for a working fix for the BREACH attack on google could be avoided if you link to a resource you verify as working.

drwetter:
Thx for the flowers. I'd prefer calling it tool.

Links with further infos are generally a good idea. The output needs to take this into the account though and somebody has to implement the "further infos".

For breach mitigation I use for Apache this: https://community.qualys.com/message/20352#20352

drwetter:
As you might have followed the discussion in #213 it's not easy. That was only the attack and the relevance, then people might have lots of servers: nginx, Apache, lighttpd, OLS, F5 BigIP, varnish, IIS 6, IIS 7, IIS *, Cisco, etc. 

It certainly would be a nice to have. So far nobody stepped fwd saying, I am willing to work on this. 
There are some resources in the web though which could be used  for major servers. Links to them here in the wiki would be enough for a start.

Harinus:
A documentation on how to fix CVE on system XY with configuration Z sounds like impossible to me. The Qualys SSL Labs give another example of documentation with the goal to inform the User/Admin about the Vulnerability 
<table><tr class="tableRow">
                            <td class="tableLabel"><font color="red">POODLE (SSLv3)</font></td>
                <td class="tableCell">
                    <font color="red">(<a style="color: red" href="https://community.qualys.com/blogs/securitylabs/2014/10/15/ssl-3-is-dead-killed-by-the-poodle-attack">more info</a>)</font>
                    <span class="greySmall">disable SSL3</span>                </td>
                    </tr></table>
In conclusion, after reading the article you know you should disable SSL3 to be save from the POODLE Attack.
Another examle would be Breach: Conclusion: DisableTLS/SSL-level compression

<table><tr class="tableRow">
                            <td class="tableLabel"><font color="red">BREACH</font></td>
                <td class="tableCell">
                    <font color="red">(<a style="color: red" href="http://breachattack.com/">more info</a>)</font>
                    <span class="greySmall">DisableTLS/SSL-level compression</span>                </td>
                    </tr></table>
How to do it in any particular case should be a google search executed by the admin who is responsible for fixing the vulnerability.
[...]

PR #2183, opened 2022-05-11
improve dns_caa record description for fileoutput #2182

manuel-sommer:


drwetter:
Hi,

basically a good idea.  Thanks!

The idea came up before though -- for EVERY finding (#194, maybe also #966). That didn't take off, so taking this now starting with only that is not a good idea.

Besides: I am hesitant to link to unknown (to me) web sites.

Cheers, Dirk

Issue #2127, opened 2022-03-17
[BUG / possible BUG] Fix censys link in DROWN section


drwetter:
**Command line / docker command to reproduce**
``testssl.sh --drown testssl.sh``

**Expected behavior**
The link should work, currently https://censys.io/ipv4?q=981F34270F9B371B36268C516EC335977F1F09FB93899CB08194E18BA91DBDE9 does not anymore

Probably this is more appropriate: https://search.censys.io/search?resource=hosts&sort=RELEVANCE&per_page=50&virtual_hosts=INCLUDE&q=981F34270F9B371B36268C516EC335977F1F09FB93899CB08194E18BA91DBDE9 
(parameters: TBD, they got quite long to fit in a terminal window)

**Your system (please complete the following information):**
(doesn't matter)

PR #2184, opened 2022-05-13
Try to fix censys URL in DROWN

drwetter:
see #2127. It's very long though: 152 chars. 🤔


drwetter:
Does anybody have an idea to shorten the output properly?

image

Definition of properly: in a terminal window from KDE or Gnome it should be still clickable. And: no privacy compromising link shortener

PR #2188, opened 2022-05-16
Stop using process substitution. On readonly or restricted disks, it …

demiremrece:
Stop using process substitution. On readonly or restricted disks, it causes error ("/dev/fd/62 No such file or directory"). It is basically that process substitution command creates a temp file where it is only readonly directory. For example; if you build your docker images inside AWS ECR and try to run those images on AWS Lambda which allows you to write files only inside "/tmp" directory, you will face an error while using testssl.sh. 

For more info about process substitution, please have a look at [wiki](https://en.wikipedia.org/wiki/Process_substitution#:~:text=In%20computing%2C%20process%20substitution%20is,occur%2C%20by%20the%20command%20shell.)

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -9354,7 +9354,9 @@ certificate_info() {
 #         see #967
 
      out "$indent"; pr_bold " Certificate Validity (UTC)   "
-     IFS=',' read -r startdate enddate diffseconds days2expire yearstart < <(determine_dates_certificate "$cert_txt")
+
+     determine_dates_certificate "$cert_txt" > /tmp/ddc_cert.txt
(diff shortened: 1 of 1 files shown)


drwetter:
Hi @demiremrece ,

thanks but nope ;-)

I'd recommend rather use an appropriate other service which allows you to use testssl.sh. Process substitutions is (as the wiki says) a core of bash since a long time. Working around would make you maybe happy but has disadvantages for all others.

Besides: the patch rolls back recent changes. I am pretty sure you didn't get all instances of process substitution.

Thx, Dirk

PR #2189, opened 2022-05-24
Fix CRIME test on servers only supporting TLS 1.3

EliteTK:


drwetter:
I reconfigured the server that's the reason of the unsuccessful check for 2/3. For the remainder 1/3: The complaint wrt the headerTime is strange though.

Later..

PR #2194, opened 2022-06-07
Reorganize ciphers_by_strength()

dcooper16:
This PR modifies `ciphers_by_strength()` and `run_server_preference()` so that the message indicating that ciphers are listed by strength is not printed until the list of supported ciphers has been determined. This is in support of #1311, as it will allow the message to be modified based on the set of supported ciphers.

This PR also modifies both `ciphers_by_strength()` and `cipher_pref_check()` so that the order in which ciphers are listed (by strength or server preference) is not printed if the server does not support the protocol.


drwetter:
Sorry for the delay, David!

PR #2210, opened 2022-08-04
Fix cipher grading

dcooper16:
This PR fixes the way that ciphers are graded by only calling `set_ciph_str_score()` from `neat_list()` if the cipher suite is supported (i.e., if $how2show is true).


drwetter:
Sorry about the delay. 

You're right, there was a closing round bracket which I overlooked around line 3944 and similar so that I thought $available is an argument to neat_list().

Thanks again!

PR #2220, opened 2022-08-27
Send a proper LHLO command with argument (like EHLO), to fix -t lmtp.

ghen2:
(A sneaky variant seems not necessary, since LMTP is typically not a public service.)


drwetter:
thanks!

PR #1575, opened 2020-04-19
STARTTLS: add support for xmpp-server

horazont:
XMPP client-to-server and server-to-server links historically use
different XML namespaces. Some server implementations are strict
about this and will not proceed with the connection attempt when
the client namespace (`jabber:client`) is used on a
server-to-server link.

openssl s_client also supports `xmpp-server`.

PR #2222, opened 2022-08-31
Fix xmpp-server detection

drwetter:
See #1575


horazont:
Does seem to work with the build from https://github.com/drwetter/testssl.sh/pull/1575#issuecomment-1224128371, as well as the debian-bullseye binary.

PR #2225, opened 2022-09-06
Fix determine_cert_compression() and certificate_transparency()

dcooper16:
`determine_cert_compression()` and `certificate_transparency()` do not work in debug mode, since `tls_sockets()` writes debugging messages to stdout. This PR fixes the problem by having `determine_cert_compression()` and `certificate_transparency()` return their results using a global variable rather than writing the results to stdout and having having `run_server_defaults()` catch the output.


drwetter:
Thanks!

PR #2226, opened 2022-09-06
Fix certificate_transparency()

dcooper16:
`certificate_transparency()` does not work in debug mode, since `tls_sockets()` writes debugging messages to stdout. This PR fixes the problem in the 3.0 branch by having `certificate_transparency()` return its results using a global variable rather than writing the results to stdout and having having `run_server_defaults()` catch the output.


drwetter:
Thanks!

PR #2227, opened 2022-09-06
Fix decrypting TLS 1.3 server response

dcooper16:
There is at least one server that includes a new session ticket in the same packet as the Finished message. This confuses `check_tls_serverhellodone()` since the new session ticket is encrypted under the application traffic keys rather than the handshake keys. `check_tls_serverhellodone()`, being unable to decrypt the new session ticket, reports a failure and does not return any of the decrypted data.

This PR fixes the problem by having `check_tls_serverhellodone()` simply return (or ignore) any data that appears after the Finished message. If such data is returned, then `tls_sockets()` derives the application traffic keys and decrypts it so that it can be parsed by `parse_tls_serverhello()`.


drwetter:
Good work, David! Thank you!

PR #2229, opened 2022-09-07
Consistent overall_grade output

cancom-oss:
Adjust output json to be in line with other 'overall_grade' outputs


drwetter:
Thank you!

PR #2233, opened 2022-09-07
STARTTLS not mandatory

### LINKED ISSUE #1734: Clarify or improve STARTLS check for "Encryption via STARTTLS is not mandatory (opportunistic)."
In `run_rating()`, this grade output is a bit misleading:

```sh
[[ -n "$STARTTLS_PROTOCOL" ]] && set_grade_cap "T" "Encryption via STARTTLS is not mandatory (opportunistic)."
```

Having only seen the result reasoning for the grade cap, I thought it implied that STARTTLS was not set to enforce TLS (reject failures). For port 25 STARTTLS should be opportunistic and allowed to downgrade to insecure plaintext, however for the *submission* port 587, this is not mandatory for public mail servers and can be configured to reject plaintext / insecure connections.

I was testing against a Postfix install that allows for exactly that, so I expected the grade cap for port 25, but not port 587, as I have that configured to only allow successful explicit TLS connections. I am not sure if `testssl.sh` can detect this difference, if it cannot, then rewording the grade cap reason would be helpful, perhaps:

> Encryption via STARTTLS may potentially support plaintext fallback (insecure)

At the same time, when it's known that this is not the case, and is the only reason for a grade cap, the entire rating is void as a result. It would be helpful to still see the score (separate issue?) or some flag to inform the test that STARTTLS is mandatory to interact further(port 587 requires auth credentials to submit mail, unlike port 25), thus avoiding the grade cap?

### Environment

`testssl.sh` official DockerHub image run with docker-compose:

```docker-compose.yaml
# snippet:
  testssl:
    image: drwetter/testssl.sh:3.1dev
    container_name: testssl-test
    volumes:
      - $PWD/data/stepca/home/certs:/tls:ro
      - $PWD/data/testssl/tests:/tests:ro
      - $PWD/data/testssl/results:/results
    working_dir: /results
    dns: 192.168.1.12
    command: --add-ca /tls/root_ca.crt --outprefix profile_intermediate --json-pretty --log --color 0 --severity LOW --wide --show-each --file /tests --mode parallel
    depends_on:
      - coredns
      - mail
```

tests file:

```
example.test:465
example.test:995
example.test:993
--starttls smtp example.test:25
--starttls smtp example.test:587
--starttls pop3 example.test:110
--starttls imap example.test:143
```

If needed, I am working on publishing a repo of the full setup for automated local/offline TLS test with `testssl.sh` against the mailserver, everything runs in Docker via `docker-compose`, might be helpful for this project?

cancom-oss:
Adds a new flag to not cap the grade to `T` when STARTTLS is not mandatory.

Closes #1734

### DIFF
--- a/doc/testssl.1.md
+++ b/doc/testssl.1.md
@@ -120,6 +120,8 @@ The same can be achieved by setting the environment variable `WARNINGS`.
 
 `--xmpphost <jabber_domain>` is an additional option for STARTTLS enabled XMPP: It expects the jabber domain as a parameter. This is only needed if the domain is different from the URI supplied.
 
+`--starttls-not-mandatory` is an additional option to not cap the grade to `T` where STARTTLS is not mandatory.
+
 `--mx <domain|host>` tests all MX records (STARTTLS on port 25) from high to low priority, one after the other.
 
 `--ip <ip>` tests either the supplied IPv4 or IPv6 address instead of resolving host(s) in `<URI>`. IPv6 addresses need to be supplied in square brackets. `--ip=one` means: just test the first A record DNS returns (useful for multiple IPs). If `-6` and  `--ip=one` was supplied an AAAA record will be picked if available. The ``--ip`` option might be also useful if you want to resolve the supplied hostname to a different IP, similar as if you would edit `/etc/hosts` or `/c/Windows/System32/drivers/etc/hosts`. `--ip=proxy` tries a DNS resolution via proxy.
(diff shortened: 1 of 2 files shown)


drwetter:
Sorry but 
a) we need to finish the discussion in #1734 if and how we implement this.
b) this is just a cheat switch w/o any test e.g. test on a port like @polarathene indicated. Or an additional warning, or ...
c) I do not like the name of the switch. It should reflect the cheat and / or a waiver for responsibility

PR #2234, opened 2022-09-07
Fix grade cap reasons not showing weak public keys

### LINKED ISSUE #2232: [BUG] Grade cap reasons not showing weak public keys
**Before you open an issue please check which version you are running and whether it is the latest in stable / dev branch**
I am running version ( commit b3c49b584d843a28915af05e8f6ba4af6ed16be3  / testssl.sh       3.1dev from https://testssl.sh/dev/)

**Before you open an issue please whether this is a known problem by searching the issues**
I couldn't find anything 

``docker run --rm -ti drwetter/testssl.sh https://dh512.badssl.com/``

**Expected behavior**
Grade cap reasons should include `Grade capped to F. Using an insecure public key and/or ephemeral key`. Currently, the cap reason is missing, but overall grade is correctly set to F.

 - OS: ``Alpine Linux v3.13``
 - Platform: ``Linux 5.10.0-14-amd64 x86_64`` 
 - OpenSSL: ``/home/testssl/bin/openssl.Linux.x86_64``

**Additional context**
The reason for this problem is related to the placement of `sorted_reason` and `sorted_warnings`.

```sh
IFS=$'\n' sorted_reasons=($(sort -ru <<<"${GRADE_CAP_REASONS[*]}"))
IFS=$'\n' sorted_warnings=($(sort -u <<<"${GRADE_WARNINGS[*]}"))
IFS=$old_ifs
```

[testssl.sh#L22399-22401](https://github.com/drwetter/testssl.sh/blob/b3c49b584d843a28915af05e8f6ba4af6ed16be3/testssl.sh#L22399-22401)

These are responsible for pretty printing the output. However, set_grade_cap can be invoked **AFTER** the sort already happened, for example in [testssl.sh#L22462-22466](https://github.com/drwetter/testssl.sh/blob/b3c49b584d843a28915af05e8f6ba4af6ed16be3/testssl.sh#L22462-22466) after the KEY_EXCH_SCORE is used.

We could move the sort to the bottom right before pretty printing [testssl.sh#L22572](https://github.com/drwetter/testssl.sh/blob/b3c49b584d843a28915af05e8f6ba4af6ed16be3/testssl.sh#L22572), but `sorted_warnings` is used to measure `A+` or `A-`. We could also just sort it twice, once at the current position, and once right before the `A+` and `A-` measurement.

Or even better, don't rely on `sorted_warnings` to measure the exceptional score, but use `GRADE_WARNINGS` instead. It shouldn't matter if they are sorted or not, since we are relying on the total count of warnings.

We would like to make a Pull Request for this, but not sure, which version you would prefer?

cancom-oss:
Fixes #2232

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -22395,10 +22395,6 @@ run_rating() {
 
      [[ -n "$STARTTLS_PROTOCOL" ]] && set_grade_cap "T" "Encryption via STARTTLS is not mandatory (opportunistic)."
 
-     # Sort the reasons. This is just nicer to read in general
-     IFS=$'\n' sorted_reasons=($(sort -ru <<<"${GRADE_CAP_REASONS[*]}"))
-     IFS=$'\n' sorted_warnings=($(sort -u <<<"${GRADE_WARNINGS[*]}"))
-     IFS=$old_ifs
      pr_bold " Rating specs"; out " (not complete)  "; outln "SSL Labs's 'SSL Server Rating Guide' (version 2009q from 2020-01-30)"
      pr_bold " Specification documentation  "; pr_url "https://github.com/ssllabs/research/wiki/SSL-Server-Rating-Guide"
      outln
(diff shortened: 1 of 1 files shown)


drwetter:
@magnuslarsen : is that what you wanted?

magnuslarsen:
Exactly! This seems like the best approach, and saves sorting the array twice.

drwetter:
Ups, sorry, Thought I merge that.

Thanks @cancom-oss & @magnuslarsen

PR #2236, opened 2022-09-12
Use `grep -E` instead of `egrep`

a1346054:
`egrep` is deprecated and latest GNU grep emits a warning message when invoking `egrep`


drwetter:
Thank you. I thought I catched them a while ago in testssl.sh at least.

Do you mind to have a look at 3.0 too?

ghen2:
3.0 doesn't use `egrep`, but had one instance of `fgrep`, fixed in https://github.com/drwetter/testssl.sh/pull/2243.

PR #2237, opened 2022-09-12
Use bash-builtin `command -v` instead of external `which`

a1346054:
`command -v` is a bash builtin and is a standardized version of `which`


drwetter:
Thanks!

PR #2238, opened 2022-09-14
Fix Shellcheck SC2235

dcooper16:
SC2235 is "Use { ..; } instead of (..) to avoid subshell overhead."

In a large number of places testssl.sh uses parenthesis in complex boolean expressions in order to specify an evaluation order. The parenthesis result in the expression being evaluated in a subshell, which makes evaluation very expensive. This PR addresses the problem by rewriting any expressions that unnecessarily create subshells.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -516,15 +516,15 @@ set_severity_level() {
 show_finding() {
      local severity=$1
 
-     [[ "$severity" == DEBUG ]] ||
-     ( [[ "$severity" == INFO ]] && [[ $SEVERITY_LEVEL -le $INFO ]] ) ||
-     ( [[ "$severity" == OK ]] && [[ $SEVERITY_LEVEL -le $OK ]] ) ||
-     ( [[ "$severity" == LOW ]] && [[ $SEVERITY_LEVEL -le $LOW ]] ) ||
-     ( [[ "$severity" == MEDIUM ]] && [[ $SEVERITY_LEVEL -le $MEDIUM ]] ) ||
-     ( [[ "$severity" == HIGH ]] && [[ $SEVERITY_LEVEL -le $HIGH ]] ) ||
-     ( [[ "$severity" == CRITICAL ]] && [[ $SEVERITY_LEVEL -le $CRITICAL ]] ) ||
-     [[ "$severity" == WARN ]] ||
-     [[ "$severity" == FATAL ]]
+     [[ "$severity" == DEBUG ||
+     ( "$severity" == INFO && $SEVERITY_LEVEL -le $INFO ) ||
+     ( "$severity" == OK && $SEVERITY_LEVEL -le $OK ) ||
+     ( "$severity" == LOW && $SEVERITY_LEVEL -le $LOW ) ||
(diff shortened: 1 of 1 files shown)


drwetter:
HI David,

good point! 

I'd like to merge that but as I see now ``show_finding()`` , I am scratching my head about readability and sense . Any thoughts?

Thx, Dirk

dcooper16:
> I'd like to merge that but as I see now `show_finding()`, I am scratching my head about readability and sense . Any thoughts?

I find switching `( ... )` to `{ ...; }` far more confusing than the change made to `show_finding()`. So, I used that way whenever possible. E.g., using
```
[[ "$tls13_ciphers" == notpresent || -z "$tls13_ciphers" ]]
```
instead of
```
{ [[ "$tls13_ciphers" == notpresent ]] || [[ -z "$tls13_ciphers" ]]; }
```
I would have done that everywhere, expect that is doesn't work for Booleans. E.g., `[[ "$HAS_TLS13" || $proto != -tls1_3 ]]` doesn't work.

We can skip this PR, if you think it hurts readability. In my testing on Ubuntu I found that the revised version of `show_finding()` is about 100 times faster than the current version. However, the overall run time of testssl.sh with this PR wasn't noticeably faster than the current version. Perhaps it would make more of a difference on platforms where creating subshells is more expensive, but it could be that the code changed by this PR isn't used enough to make a meaningful difference in the overall runtime.

drwetter:
Thanks, David!

... and for the time needed to chew on this ;-)

Bottom line / my impression
* IMO ``[[ "$tls13_ciphers" == notpresent || -z "$tls13_ciphers" ]]`` is better to read as ``{ [[ "$tls13_ciphers" == notpresent ]] || [[ -z "$tls13_ciphers" ]]; } ``. It's shorter.
* It's a pity that it doesn't work for Booleans
* Switching from () to {;}:  felt also strange to me. But it's ok. If there's a reason to do that, why not change it.

However now we have still some statements  like ``[[ "${service[i]}" == ANY ]] || [[ "${service[i]}" =~ $client_service ]]`` which feels a bit inconsistent. We could suggest that for new code we use ``[ "${service[i]}" == ANY || "${service[i]}" =~ $client_service ]]`` instead. What's your take on that?

My issue with ``show_finding()`` was a different one. The function didn't appear to me much readable in the first place.  And it's being called once. But looking at it I wasn't really tempted that moment to work on an improvement.

dcooper16:
> However now we have still some statements like `[[ "${service[i]}" == ANY ]] || [[ "${service[i]}" =~ $client_service ]]` which feels a bit inconsistent. We could suggest that for new code we use `[ "${service[i]}" == ANY || "${service[i]}" =~ $client_service ]]` instead. What's your take on that?

It doesn't really matter to me. If there is a performance difference between the two, it is only a few percent. So, it's just a style question, and I don't find `[[ "${service[i]}" == ANY || "${service[i]}" =~ $client_service ]]` any easier to read than `[[ "${service[i]}" == ANY ]] || [[ "${service[i]}" =~ $client_service ]]`. However, if you think having a consistent style would be helpful, then we can go with `[[ "${service[i]}" == ANY || "${service[i]}" =~ $client_service ]]` in the future, and even work on changing the current instances.

PR #2239, opened 2022-09-15
Fix alignment in neat_list()

dcooper16:
When `neat_list()` is printing information about a cipher suite that uses (EC)DH key exchange where the information was obtained using an old version of OpenSSL (one for which $HAS_DH_BITS is false) the rows are not properly aligned, since the key exchange input includes an unexpected trailing space. This PR fixes the problem by removing any trailing spaces from $kx.


drwetter:
Thanks!

PR #2242, opened 2022-09-17
Fix grep 3.8 warnings on unneeded escapes of hyphen, slash, space.

ghen2:
Use -e where needed to avoid confusion of expression as arguments.

Fixes https://github.com/drwetter/testssl.sh/issues/2241.

PR #2243, opened 2022-09-18
 Fix grep 3.8 warnings on unneeded escapes of hyphen, slash, space.

ghen2:
Use -e where needed to avoid confusion of expression as arguments.

Backport of https://github.com/drwetter/testssl.sh/pull/2242, fixes https://github.com/drwetter/testssl.sh/issues/2241.

Fix an invocation of fgrep as well (`fgrep` = `grep -F`).


drwetter:
Thanks again!

PR #2261, opened 2022-11-02
if PROXY variable is set there is no need to do a direct connection attempt

osown:
Through swapping the expressions a direct connection attempt is only done if PROXY is null.
This fixes a script hangup when using testssl with the --proxy option.


drwetter:
Thanks!

PR #2263, opened 2022-11-02
Fix infinite loop in run_pfs()

dcooper16:
This PR is the same as #2262, but for the 3.0 branch.


drwetter:
Not sure why the codespell action ran on the ~/.git/object files...

PR #2264, opened 2022-11-07
Clean up adding padding

dcooper16:
This commit simplifies the adding of padding data in a few places. Rather than adding one or two bytes at a time in a "for" loop, all of the padding is added in one step by extracting it from a long padding string. (The one exception is in `run_robot()`, where a "for" loop is used to add additional padding in cases in which the RSA modulus is longer than the pre-defined padding string.)

Extracting the padding from a long string is faster than using a "for" loop and it makes the debugging file a little cleaner.

The idea is the same as PR #1940.


drwetter:
Good idea, thanks!

Issue #1251, opened 2019-04-28
Is it time to warn for TLSv1.0 and 1.1


Night1:
Is it time to depreciate old TLS versions? 

Security is a critical part of any system, and in part this starts with supported protocols, 
Github has already removed support and PCI compliance also require removal of TLS1.0.
Chrome should also start warning / removing support soon too. 

My suggestions is that we edit 
 Line 4890 to 4895 From

```

         0)   outln "offered"
               fileout "$jsonID" "INFO" "offered"
               latest_supported="0301"
               latest_supported_string="TLSv1.0"
               add_tls_offered tls1 yes
               ;;     
```

To 

```
          0)   prln_svrty_medium "offered (Not ok. Consider removing support)"
               fileout "$jsonID" "MEDIUM" "offered"
               latest_supported="0301"
               latest_supported_string="tls1"
               add_tls_offered tls1 yes

```

and from 4965 to 4970

```
        0)   outln "offered"
               fileout "$jsonID" "INFO" "offered"
               latest_supported="0302"
               latest_supported_string="TLSv1.1"
               add_tls_offered tls1_1 yes
               ;;       
```

To 

```
          0)   prln_svrty_medium "offered (NOT ok. Consider removing support)"
               fileout "$jsonID" "MEDIUM" "offered"
               latest_supported="0302"
               latest_supported_string="tlsv1.1"
               add_tls_offered tls1_1 yes

```

If it is high medium of even critical  can be up for discussion however I think it is time warn for old protocols.  I have a fork ready to create a pull request. 
any comments?

drwetter:
For 3.0 I am hesitating but still thinking. One point for it would be a change in recent (also German) government standard to deprecate TLS 1.0 and 1.1. NIST was ahead, right @dcooper16 ?

For 3.1dev: yes. RSA kx e.g. is another point.

drwetter:
The risk label is important: Counter question: Why "medium"?

Night1:
Lable is current set to Info, which turns on a none warning symbol if one is to display it red or with a notice that it is not ok to still be using old protocols it might entice more to upgrade..

I changed mine to post this our for TLS1.2 and 1.3

```
 SSLv2      not offered (OK)
 SSLv3      not offered (OK)
 TLS 1      not offered (OK)
 TLS 1.1    not offered (OK)
 TLS 1.2    offered (OK)
 TLS 1.3    offered (OK): final
[...]

PR #2265, opened 2022-11-09
Show server supported signature algorithms

### LINKED ISSUE #1519: Show peer signing digest / signed_params hash function

dcooper16:
This PR fixes #1519 by modifying `run_fs()` to show the signature algorithms the server supports in the ServerKeyExchange message for TLS 1.2 and in the CertificateVerify message for TLS 1.3.

Signature algorithms are not shown for TLS 1.1 and earlier, since for those protocol versions the signature algorithm to use is specified by the protocol. While the signature algorithm used in TLS 1.1 and earlier is weak, testssl.sh already warns if these protocol versions are supported.

The code in this PR includes an initial proposal for the rating the set of supported signature algorithms, but it likely needs some changes. It assigns a rating of CRITICAL if the server only supports signature algorithms that use MD5 or SHA1. I assigned the worst rating to this case, since it was my understanding that this could lead to interoperability problems. If the server supports stronger algorithms, but also supports MD5, that is rated HIGH. Support for SHA1 is rated LOW.

Given that TLS 1.1 and earlier use SHA1 or SHA1+MD5 as the hash, it seems that using SHA1 with TLS 1.2 is no worse than using TLS 1.1 or earlier. So, the rating for supporting SHA1 should be no worse than the rating for supporting TLS 1.1 (see #1251).


drwetter:
Hi David,

thanks for your work.

For a reasonable up-to-date web server (testssl.sh) the patch costs at least 5 seconds, in a worse case 10 seconds. Not sure why at the moment. But if you have an idea why, it would be great to know. (a bad server supporting SSLv3 to TLS 1.2 costs 1-3 seconds). 

Three minor points:
* The sequence of TLS 1.2 and TLS 1.3  sig_algs offered should be rather low  to high number.
* _nonsense deleted ;-)_
* In one case I had encountered an empty line for TLS 1.2 sig_algs

image

I am just trying to figure out why

dcooper16:
Hi Dirk,

I reversed the order of the output for TLS 1.2 and TLS 1.3.

I haven't been able to find a server that shows an empty list of signature algorithms for TLS 1.2, but with some additional testing I found 2 server that show an empty list for TLS 1.3. In both cases the server seems to be returning a CertificateVerify message where the SignatureScheme if 0xFFFF. Since testssl.sh doesn't recognize this value, it just stops the testing for supported signature algorithms.

So far, I have not been able to find anything to suggest that testssl.sh is not correctly decrypting or parsing the CertificateVerify message, and I haven't found anything to explain how 0xFFFF could be an acceptable value. So, I do not know what is happening.

It may be the case with the server you tested that the SignatureAndHashAlgorithm specified in the ServerKeyExchange is unrecognized, but I would need to see the server's response to know whether that is the case.

dcooper16:
For the servers that were returning a SignatureScheme of 0xFFFF for TLS 1.3, the problem seems to occur if the signature_algorithms extension includes too many signature schemes. So, I added a commit that breaks the list of signature schemes in two and tries each list separately. This seems to fix the problem.

I don't know whether the same problem happens with some servers with TLS 1.2, but I reduced the size of the signature_algorithms extension for TLS 1.2 as well. For TLS 1.2 this simply involved generally not including values that were inconsistent with the key type being tested. For example, if the cipher suite being offered requires the server to use an RSA key, there is no reason for the signature_algorithms extension to offer signature schemes that require the use of a DSA or ECDSA key.

drwetter:
Hi David, thanks a lot!

yeah, it's a bit faster now. Other than that I can't see whether there's more time we can save now.

For next development stage I was thinking about saving all reusable information we got from far in memory like e.g. we already did regarding protocols (maybe better, I did that a bit too clunky to handle). That may save one or another handshake.

teki69:
Hello,
I have a TLS 1.2 server supporting PFS ciphers, and yet "TLS 1.2 sig_algs offered" is showing empty :

2024-12-20 19_51_25-Clipboard

I'm using testssl 3.2rc3.
Would you like me to send you the report or its public IP in a private message ?

Best regard

drwetter:
yes pls. ``grep SWCONTACT testssl.sh``

drwetter:
Best is to open a separate issue , @teki69 . Otherwise this might get lost. Just a link on your message here is fine

PR #2267, opened 2022-11-10
Restrict permissions, update actions

drwetter:
The actions release numbers were taken from the 3.1dev branch.

    Note: there was one strange codespell error in PR #2263.
    Maybe the updated action will avoid this in the future.


drwetter:
This looks like it won't fix the problem why the codespell action runs on the ˜/.git/object directory. 

Kind of strange as in 3.1dev it does not.

PR #2269, opened 2022-11-10
Hint for mastodon

drwetter:
... ... and to a separate account


drwetter:
The code spell action failed here too. Looks like it's related to https://github.com/codespell-project/codespell/issues/2539 albeit the two branches here shows no significant differences:

image

PR #2050, opened 2021-11-29
Fix #2049

dcooper16:
This PR fixes #2049 by converting newline characters to spaces in JSON and CSV findings.

`fileout()` calls `newline_to_spaces()` on any `$finding` that is to be written to a JSON or CSV file. However, this only affects actual newline characters in the string, not escaped newline characters (i.e., "\n"). Escaped newline characters pass through this function unchanged, but then get converted to newline characters when they are written to the JSON and/or CSV files. This PR fixes the problem by also converting escaped newline characters ("\n") to spaces.

If this PR is accepted, then I can create another one for the 3.0 branch.

Issue #2049, opened 2021-11-29
[BUG] Invalid json generated if value contains newline


Max7cd:
`--jsonfile-pretty` produces invalid json when a value contains an newline character.

```json
{
    "Invocation"  : "testssl.sh --debug 2 --color 0 --jsonfile-pretty /tmp/result_tls_F8BD6DDBF638348FB3075FF8DD29EC58.json -p -t smtp mx.worldserver.net.:25",
    "at"          : "aa38e25567de:/app/testssl/bin/openssl.Linux.x86_64",
    "version"     : "3.1dev ",
    "openssl"     : "OpenSSL 1.0.2-chacha from Jan 18 17:12:17 2019",
    "startTime"   : "1638195820",
    "scanResult"  : [
        {
            "id"           : "scanProblem",
            "severity"     : "FATAL",
            "finding"      : "Can't connect to '217.13.200.135:25'
Make sure a firewall is not between you and your scanning target!"
        }],
    "scanTime"  : "Scan interrupted"
}
```

Reading this json in python produces the following error:
`json.decoder.JSONDecodeError: Invalid control character at: line 11 column 87 (char 670)`

The problem is, that the newline in the string scanResult[0]["finding"] should be escaped.

Expected Output:
```json
{
    "Invocation"  : "testssl.sh --debug 2 --color 0 --jsonfile-pretty /tmp/result_tls_F8BD6DDBF638348FB3075FF8DD29EC58.json -p -t smtp mx.worldserver.net.:25",
    "at"          : "aa38e25567de:/app/testssl/bin/openssl.Linux.x86_64",
    "version"     : "3.1dev ",
    "openssl"     : "OpenSSL 1.0.2-chacha from Jan 18 17:12:17 2019",
    "startTime"   : "1638195820",
    "scanResult"  : [
        {
            "id"           : "scanProblem",
            "severity"     : "FATAL",
            "finding"      : "Can't connect to '217.13.200.135:25'\nMake sure a firewall is not between you and your scanning target!"
        }],
    "scanTime"  : "Scan interrupted"
}
```

**System:**
 - OS: Manjaro Linux
 - Platform: Linux 5.10.79-1-MANJARO x86_64
 - Version: 3.1dev aca7582
 - OpenSSL: OpenSSL 1.0.2-chacha from Jan 18 17:12:17 2019

Max7cd:
@dcooper16 your fix seems to work (at least for my specific case) 

Thank you very much for the fast fix.

PR #2276, opened 2022-11-16
Fix #1747

### LINKED ISSUE #1747: Remove spaces from certificates in JSON
Currently certificates in JSON output look like 

``
-----BEGIN CERTIFICATE----- MIIFXDCCBESgAw[..]foACzLsvA
aJ2+MA0GCSqGSIb3DQEBCwUA MEoxCzAJBgNV[..]eXB0MSMwIQYDVQQD ExpMZXQnc[..] ``

As a convenience feature it would be great of the spaces in the certificate string itself could be removed. That could save later a bit of work when putting it together. Or even omit the BEGIN / END line as MS' ADFS does it (/federationmetadata/2007-06/federationmetadata.xml)

dcooper16:
This PR fixes #1747 by converting PEM encoded certificates that are sent to `fileout()` to a single line. As suggested in #1747, '\n' is added after the '----- BEGIN ... -----' line and before the '------ END ... ------' line.

In order to ensure that '\n' appears in the string in the JSON and CSV files, '\\\n' is sent to `fileout()` so that `printf -- "%b"` converts '\\\n' to '\n' rather than converting '\n' to a newline character.

In order to prevent `fileout()` from converting '\\\n' to '\ ', this PR moves the fix for #2049 (see PR #2050) from `fileout()` to `fatal()`.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -872,6 +872,26 @@ strip_quote() (
      )"
 )
 
+# Converts a string containing PEM encoded data to one line.
+pem_to_one_line() {
+   local pem="$1"
+   local header="" footer=""
+
+   if [[ "$pem" =~ .*-+BEGIN\ [A-Za-z0-9]+-+ ]]; then
+        header="$BASH_REMATCH"
+        pem="${pem/$header/}"
+   fi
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, David.

The CI run ran for hours. Just restarted it

PR #2278, opened 2022-11-23
Fix extract_calist()

dcooper16:
When a server supports client authentication, `extract_calist()` extracts the list of supported certification authorities sent by the server. `extract_calist()` uses different code to extract the list from a TLS 1.3 response than from a TLS 1.2 or earlier response, since the CertificateRequest message was changed for TLS 1.3.

For TLS 1.2 and earlier, `extract_calist()` assumes that the CertificateRequest message is a sequence of certificate types, signature algorithms, and certification authorities. However, the signature algorithms field was added in TLS 1.2 and does not appear in TLS 1.1 and earlier. So, the current code does not work unless the server supports TLS 1.2 or TLS 1.3.

This PR fixes the problem by checking whether the response is a TLS 1.2 response, and skipping over the extraction of the signature algorithms field if the response is neither TLS 1.2 nor TLS 1.3.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -21299,14 +21299,19 @@ print_dn() {
 # distinguished names that are in the CA list.
 extract_calist() {
      local response="$1"
-     local is_tls13=false
+     local is_tls12=false is_tls13=false
      local certreq calist="" certtypes sigalgs dn
      local calist_string=""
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, David!

PR #2279, opened 2022-11-23
Add --abort-no-tls switch

ikstream:
When used in automated pipelines, processing output from other tools with testssl might cause a halt of the whole pipeline if the previous output is detected false positive as TLS/SSL enabled Server.

Enabling this switch will cause testssl to return ALLOK after the first sign, that TLS/SSL is not enabled on the server.


dcooper16:
Rather than defining a new flag for this one particular scenario, why not just use `--warnings batch`?

ikstream:
Thank you for pointing this out. I have been looking for an option like this, but haven't found one.

PR #2281, opened 2022-11-25
Fix whitespace issues

dcooper16:
This PR fixes incorrect indentations introduced in #2276 and #2278, and also removes the extra whitespace that would have been removed by #2279.


drwetter:
Thanks! Didn't see that in the PRs 🤔

PR #2294, opened 2023-01-04
Allow DNS via Proxy

w4ntun:
The current version of the script returns an error when it not possible to determine IP addresses through DNS resolution, even if **--proxy** and **--ip=proxy** flags are set. The main function always tries to determine IP addresses via DNS and exits with a fatal error if it cannot do it. Although the client cannot get the IP, the proxy could, so the SSL/TLS analysis is still possible.

The version proposed allows the analysis for an HTTP service via a proxy server and the DNS traffic can be sent directly or through the proxy using the flag **--ip=proxy**.


drwetter:
Hi,

thanks. Could you please have a look why the patch is >23k lines big?
I need to close it therefore.

Cheers, Dirk

drwetter:
Would be appreciated if you could open another PR with the relevant changes only

w4ntun:
Hi,

I'm sorry, I must have made some mistake while making changes in the repo. Let me review and you will receive a new PR only with relevant changes.

Cheers,

w4ntun:
Hi Dirk,

Fixed in #2295 

Cheers

PR #2295, opened 2023-01-05
fixed DNS via Proxy

w4ntun:
The current version of the script returns an error when it not possible to determine IP addresses through DNS resolution, even if **--proxy** and **--ip=proxy** flags are set. The main function always tries to determine IP addresses via DNS and exits with a fatal error if it cannot do it. Although the client cannot get the IP, the proxy could, so the SSL/TLS analysis is still possible.

The version proposed allows the analysis for an HTTP service via a proxy server and the DNS traffic can be sent directly or through the proxy using the flag **--ip=proxy**.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -196,7 +196,7 @@ TESTSSL_INSTALL_DIR="${TESTSSL_INSTALL_DIR:-""}"  # If you run testssl.sh and it
 CA_BUNDLES_PATH="${CA_BUNDLES_PATH:-""}"          # You can have your CA stores some place else
 EXPERIMENTAL=${EXPERIMENTAL:-false}     # a development hook which allows us to disable code
 PROXY_WAIT=${PROXY_WAIT:-20}            # waiting at max 20 seconds for socket reply through proxy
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks, @w4ntun . Sorry for the wait. 

The logic itself seems almost fine. The whole matter not straightforward though.

What bugs me is that your PR is a breaking one for those who don't have a local resolver. They now need to add  ``--ip=proxy`` to get back the old behavior. One could see that then as "works as intended" as before it did not. ATM I don't see an clean way to address that as ``determine_ip_addresses()`` checks DNS resolving but at the same time it exists if it doesn't resolve. Maybe checking $PROXY in  ``determine_ip_addresses()`` somewhere at `` [[ -z "$IPADDRs" ]];`` would help and then in the calling part remove the ``if`` condition and only use the ``else`` condition? Haven't tried.

Formally: If possible use 5 x spaces as tab and don't add white spaces after the text

drwetter:
Thanks, merged into PR #2328 for further processing

PR #2309, opened 2023-02-01
chore: Use a single `COPY` by better leveraging `.dockerignore` patterns

polarathene:
Extracted from https://github.com/drwetter/testssl.sh/pull/2305 (_specifically [this commit](https://github.com/drwetter/testssl.sh/pull/2305/commits/f15ab97ebef21758737dced54baa05dd0942e58f)_)

- `--chown` can use a single value as both user and group.
- `.dockerignore` can better control what is available to the Docker build context.
- With the `.dockerignore` improvement, your Dockerfile can be a single `COPY`.
- Comments added to both files for additional context / documentation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ USER testssl
 WORKDIR /home/testssl/
 
 # Copy over build context (after filtered by .dockerignore): bin/ etc/ testssl.sh
-COPY --chown=testssl . /home/testssl/
+COPY --chown=testssl:testssl . /home/testssl/
 
 ENTRYPOINT ["testssl.sh"]
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,9 +12,8 @@ RUN apk update && \
 USER testssl
 WORKDIR /home/testssl/
 
-COPY --chown=testssl:testssl etc/. /home/testssl/etc/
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -12,9 +12,8 @@ RUN apk update && \
 USER testssl
 WORKDIR /home/testssl/
 
-COPY --chown=testssl:testssl etc/. /home/testssl/etc/
-COPY --chown=testssl:testssl bin/. /home/testssl/bin/
-COPY --chown=testssl:testssl testssl.sh /home/testssl/
+# Copy over build context (after filtered by .dockerignore): bin/ etc/ testssl.sh
+COPY --chown=testssl . /home/testssl/
```
drwetter: Hi! Could you just add a comment here that ``--chown=testssl``is equal ``--chown=testssl:testssl``? I wasn't aware of that, so maybe others would question this too (with no reason). Or just add ``--chown=testssl:testssl``?

polarathene: No worries, it was more preferred before I realized all 3 `COPY` could be collapsed into one :)

If including the group is more readable for maintainers, lets go with that :+1: 

```suggestion
COPY --chown=testssl:testssl . /home/testssl/
```

drwetter:
Thx, FYI: will merge that once the gh action is done (don't know when you call it a day)

PR #2310, opened 2023-02-02
chore(Dockerfile): Collapse package install commands into a single line

polarathene:
Extracted from https://github.com/drwetter/testssl.sh/pull/2305 (_specifically [this commit](https://github.com/drwetter/testssl.sh/pull/2305/commits/7a286d0c3fa1b3a9959f46b1219cce2f4b299b47)_)

- `--no-cache` implies `--update` (_aka `apk update`_) while not storing the cache to disk, so no need to manually cleanup the cache with `rm -rf` afterwards.
- `--upgrade` updates any dependencies of packages that already installed.

---

I'll update `Dockerfile.git` once you've decided if you'd like this change (_or the alternative suggestion_).

**NOTE:** I have intentionally collapsed some extra information due to concerns of the added verbosity negatively impacting review (_like with the original PR_). That content is only relevant to `apk upgrade` and a related review suggestion for it.

## Using `--upgrade` slightly differs from `apk upgrade`

<details>
<summary>Details of why you might want to keep it, and why it is not presently providing proper security benefits</summary>

From a security standpoint, upgrading packages is argued as important, since there may be some new security updates since the base image was last updated. The linked PR provides additional insights in the "Further consideration" section if that interests you.

However, to use that successfully requires regular image builds that _aren't caching that `RUN` layer_:

- `3.0` branch is not the default branch, it's [scheduled builds](https://github.com/drwetter/testssl.sh/blob/cd2eef30eacc1244cda561bd6979b234a3d9a009/.github/workflows/docker-3.0.yml#L8-L9) are [not triggered](https://github.com/drwetter/testssl.sh/actions/workflows/docker-3.0.yml). Thus no new builds happened for 2 months.
- `3.1dev` branch is the default, so scheduled runs are triggered - however [they leverage the layer cache](https://github.com/drwetter/testssl.sh/actions/runs/3729520712/jobs/6325518571#step:7:201) skipping any security benefit.
- Thus security updates are only applied when bumping the image base version, or modifying the `RUN`. If upgrading packages for security updates newer than the base image is important for this image, then CI should build without cache (_it's not like this project has long build times actually requiring build cache_).

</details>


Dockerfile
```diff
@@ -1,9 +1,6 @@
 FROM alpine:3.17
 
-RUN apk update && \
-    apk upgrade && \
-    apk add bash procps drill git coreutils libidn curl socat openssl xxd && \
-    rm -rf /var/cache/apk/* && \
+RUN apk --no-cache --upgrade add bash procps drill git coreutils libidn curl socat openssl xxd && \
```
polarathene: <details>
<summary>Alternative suggestions that keep `apk upgrade`</summary>

Alternative if you want to upgrade all packages prior to adding new ones:

```suggestion
RUN apk --no-cache upgrade && \
    apk --no-cache add bash procps drill git coreutils libidn curl socat openssl xxd && \
```

---

A recent PR by @jauderho that was closed also tried to include the `-a` (`--available`) option to the `apk upgrade` command.
- This forces upgrades on installed packages that don't have new versions, which [Alpine notes sometimes is required](https://wiki.alpinelinux.org/wiki/Upgrading_Alpine) when upgrading.
- However, [Alpine switched from `uClibc` to `musl` back in 2014](https://en.wikipedia.org/wiki/Alpine_Linux), so those official docs seem a bit outdated.

As such, it's not something I'd bother with, but if you do still want this separate upgrade step, here's the same suggestion as above, but with `--available`:

Alternative if you want to upgrade all packages prior to adding new ones:

```suggestion
RUN apk --no-cache upgrade --available && \
    apk --no-cache add bash procps drill git coreutils libidn curl socat openssl xxd && \
```

</details>

jauderho:
I am not a fan of implicit tasks. IMO, it's better to spell it out explicitly.

polarathene:
> I am not a fan of implicit tasks. IMO, it's better to spell it out explicitly.

I'd usually agree, but I don't see an issue in this case.

<details>
<summary>EDIT: Collapsed response to reduce noise, as no reply for 2 weeks</summary>

- It's fairly common in alpine base images to add packages with `--no-cache`.
  - [`--no-cache` is used in an example on the Official Docker best practice docs](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#use-multi-stage-builds).
  - [Alpine official DockerHub image page](https://hub.docker.com/_/alpine) compares `apk add --no-cache` to `apt-get update && apt-get install`. Another official document [references `--no-cache` to update the package index without redundantly storing](https://github.com/alpinelinux/docker-alpine/blob/master/docs/usage.adoc).
- Likewise, upgrading the packages you don't see as often, and it's not being properly used by the CI if the intent was for improved sense of security.

If it's a concern you could just inline a comment like:

```Dockerfile
# `--no-cache` updates the package index for the operation without caching it to disk.
```

### If implicit knowledge is a concern

There's plenty of other content in the `Dockerfile` that is implicit knowledge. See my closed PR that I extracted this change from.

In that sense, how implicit something is only matters to the maintainer and quickly they can be informed if they lack that implicit knowledge. `--no-cache` is much easier in that regard vs:
- Why are these packages installed? What features are they supporting / fixing, which ones are mandatory or supplemental? Why is `drill` the preferred DNS client choice, etc.
- Why `/home/testssl`? Is it important to use a home user directory for this due to how the shell script or one of the packages used could behave differently? Not a concern if it was just `/testssl` or similar, but a home dir can imply non-obvious config / env influences.
- Why `COPY` with `etc/.` + `bin/.`, or the 3 `COPY` instructions when a single `COPY` can be sufficient? (_semi-beneficial to separate out `testssl.sh` for local cache layers, but not deploying images due to intended frequency of invalidation via `RUN`_).
- The purpose of `mkdir` when there is `COPY --chown`?
- The purpose of the symlink? (_when the `ENTRYPOINT` could just be `./testssl.sh`_).
- Why a separate `Dockerfile.git` and two separate branches to keep in sync (_at least in regards to CI and maintenance_), when `Dockerfile.git` is not necessary to alternatively support building from git? (_especially when it appears to have been added for the purpose of CI usage_).

I'd personally favour a simpler to maintain / grok `Dockerfile` and defer any implicit knowledge concerns to inline comments for context or documentation (_there is an associated `Dockerfile.md` for example_).

</details>

polarathene:
@drwetter I resolved merge conflicts. Good for review :+1: 

After this PR, I would propose:
- https://github.com/drwetter/testssl.sh/issues/2315
- https://github.com/drwetter/testssl.sh/issues/2314

I can then PR all these `3.1dev` Dockerfile updates to `3.0` if you like. Followed by handling the `Dockerfile.git`.

drwetter:
Thanks! 

* Git to be removed (other issue): that was legacy and should not be there
* Also basically openssl-1.1 should work (SMTP injection requirements need to be checked)
* However here the apk upgrade does not do an upgrade of existing alpine packages which in fact happen(ed).  That was a problem before. I wanted to check again. There was also an issue here. That's why the lines are like they are ;-)

polarathene:
> * Also basically openssl-1.1 should work (SMTP injection requirements need to be checked)

The update to Alpine 3.17 bumped OpenSSL to 3.x series, hence the other issue about adding the `openssl1.1-compat` package.

It was for about ~15% faster testing if using `testssl.sh` with `--openssl` path to that version. If you don't think it's worth carrying around for that and rather only ship with OpenSSL 3.x now, that's fine too :+1: 

---

> * That was a problem before.

`apk upgrade` doesn't? [It should](https://man.archlinux.org/man/apk-upgrade.8.en):

> `apk upgrade` upgrades installed packages to the latest version available from configured package repositories. When no packages are specified, all packages are upgraded if possible.
> 
> `--available`
> Reset all packages to versions available from current repositories. This resets all versioned dependencies in world. 
> Additionally, packages are selected from active repositories if possible even if it means replacing or downgrading the package.
> This is useful to reset system against new set of packages after updating repositories.

As mentioned in the suggestion, I reference Alpine docs for the `apk upgrade --available`, where in the past it (2014) it was noted to be necessary to avoid breakage when doing an upgrade.

[`apk add --upgrade`](https://man.archlinux.org/man/apk-add.8.en):

> `apk add` adds the requested packages to world and installs (or upgrades) them if not already present, ensuring all dependencies are met.
> By default apk tries to select a set of packages so that all dependencies are satisfied.
> This means that some packages can be held back in an older version if there is direct or implicit dependencies to the older version.
> 
> `--upgrade`
> Upgrade packages and it's dependencies. Normally `apk add` will avoid changing installed package unless it is required by the newly added packages or their dependencies.
> This changes the default preference to upgrading the package to the latest installable version.

---

Kinda seems like `apk upgrade` and `--upgrade` (_with `apk add`_) isn't necessary usually and could potentially cause problems? :man_shrugging:

If you'd like any changes, let me know. I can add `apk upgrade` prior if you'd like and it would be at parity with what the Dockerfile (prior to this PR) currently does.

Issue #2293, opened 2022-12-29
[Feature request] Trustcor certificate


drwetter:
**Which version are you referring to**
3.0.x and 3.2rcX? 

**Please check this repo whether this is a known feature request**
Nope

**Describe your feature request (if it's a technical feature)**
Seems that Trustcor's certificate was removed from several trust stores: https://groups.google.com/a/mozilla.org/g/dev-security-policy/c/oxX69KFvsm4/m/NxbhuW-4CgAJ

**Describe the solution you'd like**
Inspect our current certificate stores whether they come with the CA and either remove it or flag the CA in the code as not trusted

FairSSL:
Each Root CA does things a little differently and often they try to not really remove a root, but only make new certificates invalid, from a cut-off date. Even when removing fully it takes time, Chrome is removing TrustCor in version 111, likely to go out of beta in March. 

The result is that for each removal, each root store does a different mix of quick removal, slow removal, invalidation within limits or not even doing anything.

The following are examples of Root's that I am aware were removed/disabled and how each root handles it is different. (Some have multiple roots, but I am just giving one example for each)
TrustCor https://crt.sh/?caid=7900 (seems to be removed from Mozilla 10jan23 even thou crt.sh does not show it yet)
CNNIC https://crt.sh/?caid=347
WoSign https://crt.sh/?caid=1425
StartCom https://crt.sh/?caid=84

If this project is not to interested in making more complex solutions that support all versions of what happens to a Root Certificates removal, is it an acceptable solution to simply remove any root that is not fully functional in a root store, so for example removing it from Microsoft before their "dates" may take effect or from Chrome before they are in their non-beta edition?

drwetter:
From time to time I update trust stores from each source. Then, to be honest, I rarely check, for e.g. Startcom or Wosign. But there's a piece of code :

![image](https://user-images.githubusercontent.com/8036727/216152090-5fb69a3a-b59b-48b9-a870-0cbbff6c9c49.png)

Probably this code needs only to be amended. 

If you want move forward: cool. make sure that the pattern matches only that Trustcor certificate.

I just found this problem in the news somewhere and I filed the issue as I couldn't implement it in the quality I wanted bc I also have a live outside testssl.sh :-)

PR #2311, opened 2023-02-02
Feature Trustcor certificates being removed/disabled from root stores #2293

FairSSL:
To add feature TrustCor Certificate #2293 

Change exception for removed root certificates into easy edit multi-value regular expression for Organization name and making it clear that CA's are actively removed from 1+ root stores.

Organization is safer to search on, as it can only contain validated information, if more precise/alternative is needed hash would probably be better but more difficult to maintain.

Other CA's of interest all seem to have no more active certificates i.e. WoSign, DigiNotar, etc. so only TrustCor seems currently relevant.

An active certificate for testing can currently be found here: https://www.trustcor.ca

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -9416,10 +9416,11 @@ certificate_info() {
 
      out "$indent"; pr_bold " Chain of trust"; out "               "
      jsonID="cert_chain_of_trust"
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks a lot!

drwetter:
Hi @SSLbrain ,

do you mind to submit a PR for 3.0x. too?

FairSSL:
Sure not a problem, another PR created https://github.com/drwetter/testssl.sh/pull/2319 for 3.0

PR #2312, opened 2023-02-02
chore(Dockerfile): Simplify `testssl` user creation

polarathene:
Extracted from https://github.com/drwetter/testssl.sh/pull/2305 (_specifically [this commit](https://github.com/drwetter/testssl.sh/pull/2305/commits/77eba8d02626bb5c3c7a03b32ddee729837c50fe)_)

Create a `testssl` user (_and group_) with no password (`-D`) and default their shell to bash (`-s`):
- A group will implicitly be created with the same value as the user. `addgroup testssl` and `-G testssl` are not needed.
- Gecos data (`-g "testssl user"`) doesn't appear relevant to the project to be required? The default gecos value (`Linux User,,,`) should be fine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,7 @@ RUN apk update && \
     apk upgrade && \
     apk add bash procps drill git coreutils libidn curl socat openssl xxd && \
     rm -rf /var/cache/apk/* && \
-    addgroup testssl && \
-    adduser -G testssl -g "testssl user" -s /bin/bash -D testssl && \
+    adduser -D -s /bin/bash testssl && \
     ln -s /home/testssl/testssl.sh /usr/local/bin/ && \
     mkdir -m 755 -p /home/testssl/etc /home/testssl/bin
 


drwetter:
Thanks!

Dockerfile
```diff
@@ -4,9 +4,9 @@ RUN apk update && \
     apk upgrade && \
     apk add bash procps drill git coreutils libidn curl socat openssl xxd && \
     rm -rf /var/cache/apk/* && \
-    addgroup testssl && \
-    adduser -G testssl -g "testssl user" -s /bin/bash -D testssl && \
-    ln -s /home/testssl/testssl.sh /usr/local/bin/
+    adduser -D -s /bin/bash testssl && \
+    ln -s /home/testssl/testssl.sh /usr/local/bin/ && \
+    mkdir -m 755 -p /home/testssl/etc /home/testssl/bin
```
polarathene, line 9: Just a heads up @drwetter when this PR was raised, I think it was separate from the `mkdir` change, which you had merged earlier.

When you updated the branch here before merging, it looks like it didn't diff well, and accidentally brought back the mkdir line? (_the [previous PR change](https://github.com/drwetter/testssl.sh/pull/2313) is actually [missing from the files history](https://github.com/drwetter/testssl.sh/commits/3.1dev/Dockerfile) now_).

You may want to remove the line again.

polarathene (commented):
Looks like the merge caused a minor regression :sweat_smile:

Dockerfile
polarathene: [Both commits are in the `3.1dev` branch history](https://github.com/drwetter/testssl.sh/commits/3.1dev) though :man_shrugging:

drwetter:
Thanks, PR pending/merged

I manually had to resolve the conflict and I just removed w/o thinking what github labeled as conflict.

Issue #2289, opened 2022-12-14
 Why cipherlist_AVERAGE's severity is low? what does cipherlist_AVERAGE contitutes?


bawejaVanshita:
Found the following vulnerability in JSON output and bit unclear:
                          {
                                "id"           : "cipherlist_AVERAGE",
                                "severity"     : "LOW",
                                "cwe"          : "CWE-310",
                                "finding"      : "offered"
                           }

Feature request: You may include a description parameter that will provide more context to each vulnerability.

PR #2316, opened 2023-02-03
Update documentation for cipherlists tests

dcooper16:
After reading #2289, I looked at the [documentation](https://github.com/drwetter/testssl.sh/tree/3.1dev/doc) and noticed that the description of the cipherlists tests had not been updated, even though the set of cipher lists checked by `run_cipherslists()` changed in 3.1dev. This PR changes the documentation to align with the tests in 3.1dev.

One issue that #2289 raises, however, is that the JSON output for these tests may not be clear. The table below shows the text description of each of the cipher list sets that is printed to the terminal along with the corresponding JSON identifier. For the first 5 sets the relationship should be fairly clear. However, the JSON IDs `AVERAGE`, `GOOD`, and `STRONG` are not clearly aligned with the strings that are printed to the terminal.

We could consider changing the JSON IDs, but at the least we should try to add something to the documentation indicating the JSON ID that corresponds to each of these lists of cipher suites.

| terminal string | JSON ID |
| ---------------------- | ----------- |
| NULL ciphers (no encryption) | NULL |
|Anonymous NULL Ciphers (no authentication) | aNULL |
| Export ciphers (w/o ADH+NULL) | EXPORT |
| LOW: 64 Bit + DES, RC[2,4], MD5 (w/o export)  | LOW |
| Triple DES Ciphers / IDEA |3DES_IDEA |
| Obsoleted CBC ciphers (AES, ARIA etc.) |AVERAGE |
| Strong encryption (AEAD ciphers) with no FS | GOOD |
| Forward Secrecy strong encryption (AEAD ciphers) | STRONG |

### DIFF
--- a/doc/testssl.1
+++ b/doc/testssl.1
@@ -128,13 +128,15 @@ Any single check switch supplied as an argument prevents testssl\.sh from doing 
 .IP "\[ci]" 4
 \fBExport ciphers\fR (w/o the preceding ones): 'EXPORT:!ADH:!NULL'
 .IP "\[ci]" 4
-\fBLOW\fR (64 Bit + DES ciphers, without EXPORT ciphers): 'LOW:DES:RC2:RC4:!ADH:!EXP:!NULL:!eNULL'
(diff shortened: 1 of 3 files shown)


drwetter:
Thanks, good catch!

PR #2319, opened 2023-02-05
Feature request Trustcor certificate #2293 for 3.0

FairSSL:
Feature change from 3.1 for 3.0 to recognize TrustCor Systems certificates.

Change exception for removed root certificates into easy edit multi-value regular expression for Organization name and making it clear that CA's are actively removed from 1+ root stores.


drwetter:
Thanks again!

PR #2322, opened 2023-02-06
Remove set_grade_cap() from 3.0

dcooper16:
Grading is a new feature in 3.1dev,, so `set_grade_cap()` is not defined in the 3.0 branch.

This PR removes the call to `set_grade_cap()` in `certificate_info()`.


drwetter:
Thanks, David

PR #2328, opened 2023-03-03
fixed DNS via Proxy 

drwetter:
Merged from  #2295 and whitespaces corrected, for further work.

_The current version of the script returns an error when it not possible to determine IP addresses through DNS resolution, even if --proxy and --ip=proxy flags are set. The main function always tries to determine IP addresses via DNS and exits with a fatal error if it cannot do it. Although the client cannot get the IP, the proxy could, so the SSL/TLS analysis is still possible._

_The version proposed allows the analysis for an HTTP service via a proxy server and the DNS traffic can be sent directly or through the proxy using the flag --ip=proxy._


drwetter:
Comment copied over:

_The logic itself seems almost fine. The whole matter not straightforward though._

_What bugs me is that your PR is a breaking one for those who don't have a local resolver. They now need to add  ``--ip=proxy`` to get back the old behavior. One could see that then as "works as intended" as before it did not. ATM I don't see an clean way to address that as ``determine_ip_addresses()`` checks DNS resolving but at the same time it exists if it doesn't resolve. Maybe checking $PROXY in  ``determine_ip_addresses()`` somewhere at `` [[ -z "$IPADDRs" ]];`` would help and then in the calling part remove the ``if`` condition and only use the ``else`` condition? Haven't tried._

_Formally: If possible use 5 x spaces as tab and don't add white spaces after the text_

w4ntun:
Hi @drwetter!

Sorry for being away for a long time. I have written a cleaner version of the code according to your comments. I totally agree with you when saying that the solution is not straightforward for the user but I wanted to keep the possibility of scanning multiple IPs resolving the same hostname even using a proxy (only if local resolver is available). If the DNS resolving is done by the proxy, only one IP address will be scanned although more than one could be available.

I think the behavior should be as follows:
- If the _--proxy_ option is set, the DNS request must be sent through the proxy as it is de default behavior when using a proxy server. So in the new version, only the _--proxy_ option needs to be set.
- If the _--ip_ option is set, the DNS resolving is not made by the proxy and we have multiple options: the value of an **specific IP** (no need to DNS resolving), the value of **one** (when obtaining multiple IPs from the local resolver, just keep the first one) or the value of **all** (et the IP addresses from a local resolver and scan all of them).

I have included the $PROXY and $DNS_VIA_PROXY checks into `determine_ip_addresses()` as you suggested and I think that the code in the main function looks cleaner too.

I don't know if you would like to have a look at this solution or you prefer to keep the current want. If needed, I can open a new PR. 

All the best.

drwetter:
Hi @w4ntun ,

thanks! I certainly want to have a look. I'd like though to merge this one first.

drwetter:
> If the --proxy option is set, the DNS request must be sent through the proxy as it is de default behavior when using a proxy server. So in the new version, only the --proxy option needs to be set.

There are different types of DNS requests: A, AAAA, TXT, MX and the one for CAA records. Please keep in mind that if ``--proxy`` those won't work through most proxies. OTOH there might be still a local resolver which one can use.  I have't looked at your suggestion yet but ``--ip=proxy`` seems to me a better fit as the user can determine that the resolution takes place exclusively at the proxy.

Issue #2330, opened 2023-03-09
[BUG] Invalid JSON output for Content-Security-Policy finding including TABs (ascii 0x09)


bodik:
Scan produces invalid JSON output.

This is a snippet of the problematic part (note vim-like display of control characters):
```vim
                           },{$
                                "id"           : "Content-Security-Policy",$
                                "severity"     : "OK",$
                                "finding"      : "default-src 'unsafe-inline' 'self'; ^I^I^I^I^I^Iimg-src data: 'self' REDACTED...$
                           },{$
```

hexdump
```
00005a80: 2020 2020 2266 696e 6469 6e67 2220 2020      "finding"   
00005a90: 2020 203a 2022 6465 6661 756c 742d 7372     : "default-sr
00005aa0: 6320 2775 6e73 6166 652d 696e 6c69 6e65  c 'unsafe-inline
00005ab0: 2720 2773 656c 6627 3b20 0909 0909 0909  ' 'self'; ......
00005ac0: 696d 672d 7372 6320 6461 7461 3a20 2773  img-src data: 's
```

The data comes directly from the HTTP request which can produce header values with tabs characters
```
GET / HTTP/1.1
Host: [REDACTED_IPV6_ADDRESS]

HTTP/1.1 200 OK
Date: Thu, 09 Mar 2023 13:42:28 GMT
Server: Apache
Strict-Transport-Security: max-age=15768000
X-Content-Type-Options: nosniff
X-Xss-Protection: 1; mode=block
X-Frame-Options: DENY
Content-Security-Policy: default-src 'unsafe-inline' 'self';                         img-src data: 'self' REDACTED...
Referrer-Policy: no-referrer-when-downgrade
REDACTED...
```

**Before you open an issue please check which version you are running and whether it is the latest in stable / dev branch**
`commit abdd51d8150e54d1fec930cc2925562be792cf48 (grafted, HEAD, tag: v3.0.8)`

**Command line / docker command to reproduce**
"Invocation"  : "testssl.sh --quiet --full -6 --connect-timeout 5 --openssl-timeout 5 --jsonfile-pretty output-0.json [REDACTED_IPV6_ADDRESS]:443"

**Expected behavior**

Control characters in JSON data are expected to be properly encoded (ascii 0x09 TAB as `\t` sequence)

**Your system (please complete the following information):**
 - OS: Debian GNU/Linux 11 (bullseye)
 - Platform: Linux 5.10.0-21-amd64 x86_64
[...]

PR #2332, opened 2023-03-11
Make sure control chars from HTTP header don't end up in html,csv,json

drwetter:
This addresses the bug #2330 by implementing a function which removes control characters from the file output format html,csv,json at the output.

In every instance called there's a check before whether the string contains control chars, hoping it'll save a few milli seconds.

A tr function is used, omitting LF.

It doesn't filter the terminal output and the log file output, yet. It provides a function though which is not being called.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -13,4 +13,4 @@ jobs:
       - uses: codespell-project/actions-codespell@master
         with:
           skip: ca_hashes.txt,tls_data.txt,*.pem,OPENSSL-LICENSE.txt
-          ignore_words_list: borken,gost,ciph,ba,bloc,isnt,chello,fo,alle
+          ignore_words_list: borken,gost,ciph,ba,bloc,isnt,chello,fo,alle,aNULL
--- a/t/32_isHTML_valid.t
+++ b/t/32_isHTML_valid.t
@@ -15,22 +15,23 @@ my $out="";
 my $html="";
 my $debughtml="";
 my $edited_html="";
-my $check2run="--ip=one --ids-friendly --color 0 --htmlfile tmp.html";
+my $htmlfile="tmp.html";
+my $check2run="--ip=one --ids-friendly --color 0 --htmlfile $htmlfile";
 my $diff="";
 die "Unable to open $prg" unless -f $prg;
 
 printf "\n%s\n", "Doing HTML output checks";
-unlink 'tmp.html';
+unlink $htmlfile;
 
 #1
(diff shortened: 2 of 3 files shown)


drwetter:
don't know why codespell complains about aNULL 🧐

drwetter:
hi @DimitriPapadopoulos , you understand by any chance why all of a sudden codespell complains about aNULL and if I list it as an exception codespell doesn't honor this?

DimitriPapadopoulos:
The problem is two-fold:
1. Additions to typo dictionaries are a bit disruptive in CI jobs. Not sure how to best handle that in the future.
2. You need to add `anull` to the typos to ignore, not `aNULL`. From [README.rst](https://github.com/codespell-project/codespell/blob/master/README.rst):
   > **Important note:** The list passed to `-I` is case-sensitive based on how it is listed in the codespell dictionaries.

.github/workflows/codespell.yml
```diff
@@ -13,4 +13,4 @@ jobs:
       - uses: codespell-project/actions-codespell@master
         with:
           skip: ca_hashes.txt,tls_data.txt,*.pem,OPENSSL-LICENSE.txt
-          ignore_words_list: borken,gost,ciph,ba,bloc,isnt,chello,fo,alle
+          ignore_words_list: borken,gost,ciph,ba,bloc,isnt,chello,fo,alle,aNULL
```
DimitriPapadopoulos, line 16: Change `aNULL` → `anull`

drwetter: Will try

drwetter: Ok, for 3.0 it worked, see: #2335.

My question for the future: How am I supposed to know? The complaint was ``aNULL ==> annul``.

drwetter: Thanks, btw! 👍

DimitriPapadopoulos: Words are written with lower case letters in the dictionaries of typos, except for proper nouns (Mircosoft for example). I am not sure there is currently a better way to find out than looking in the dictionaries under [codespell_lib/data](https://github.com/codespell-project/codespell/tree/master/codespell_lib/data).

drwetter: Thanks. I did that, using aNULL. But for some reason git doesn't want to tell me: https://github.com/codespell-project/codespell/search?q=anull .

DimitriPapadopoulos: That's a GitHub issue. Use other means to search files.

drwetter: sigh. The files are plain text files. But sure I can myself.

PR #2344, opened 2023-03-22
refactor(dockerfile): Change base Alpine (3.17) => openSUSE Leap (15.4)

polarathene:
Based on feedback at: https://github.com/drwetter/testssl.sh/issues/2299#issuecomment-1471704041

Image size on disk:
- Alpine: `36.6 MiB`
- Leap: `55.5 MiB`

Main benefit is to move off `musl`, this comes at the expense of a slightly larger image.

One issue related to this was the static built `openssl` relies on `glibc` for some DNS logic which would fail on Alpine (_non-issue when using the `openssl` package from Alpine_).

Commits have messages for context on changes. Some inline docs have been dropped but could be brought back if desired. Less practical without the BuildKit features that were dropped though.

---

`docker run --rm -it drwetter/testssl.sh:3.1dev --openssl /usr/bin/openssl google.com`:
- Alpine (OpenSSL 3.0.8): 196 sec
- Leap (OpenSSL 1.1.1l): 108 sec

Run on a Fedora 37 VPS host with 2 vCPU.

---

For reference, since `git blame` becomes less helpful, these links should assist with any traceability going forward:

- Dependencies were documented in [this earlier commit](https://github.com/drwetter/testssl.sh/blob/0b86094ab99cdbb1e07df7699108af67d0260559/Dockerfile), while [this commit](https://github.com/drwetter/testssl.sh/pull/2344/commits/718eb3461c3ebf12e3c458444fbd6e02aea63232) was the version prior to dropping BuildKit features if a maintainer decides to adopt them in future.
- Original image history (Debian to Alpine) and [dependencies were also documented in an earlier PR of mine](https://github.com/drwetter/testssl.sh/pull/2305).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,21 +6,18 @@ ARG INSTALL_ROOT=/rootfs
 FROM opensuse/leap:${LEAP_VERSION} as builder
 ARG CACHE_ZYPPER=/tmp/cache/zypper
 ARG INSTALL_ROOT
-# Usually you could source this from /etc/os-release
-# Might not always match the image tag:
-ARG VERSION_ID=15.4
-RUN  zypper --releasever "${VERSION_ID}" --installroot "${INSTALL_ROOT}" --cache-dir "${CACHE_ZYPPER}" \
-       --gpg-auto-import-keys refresh \
-  && zypper --releasever "${VERSION_ID}" --installroot "${INSTALL_ROOT}" --cache-dir "${CACHE_ZYPPER}" \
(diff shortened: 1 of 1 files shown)


drwetter:
Very cool, thanks! 

Performance benefit: 👍

I reconfigured on dockerhub to include this branch, so that others including myself can try. That is stuff not for this day but should get this finalized hopefully by Friday

drwetter:
First tests: works like a charm 👍 Good work!

PR #2346, opened 2023-03-22
Add information to RSA-PSS report

odinmylord:
This is a small change that shouldn't give problems.
Sometimes a user may want to know exactly which rsa-pss signature algorithm the server uses.
At the moment regardless of the iana code both rsa-pss-rsae and rsa-pss-pss are described as RSA-PSS in the report.
With this update they are now described as RSA-PSS-RSAE (from 0x0804 to 0x0806) and RSA-PSS-PSS (0x0809 to 0x080B


drwetter:
Thanks!

@dcooper16 : as it was your code, could you you review this?

dcooper16:
This seems like a good idea. While the signature algorithm is the same, there are two different code points. The difference is whether the subjectPublicKeyInfo in the certificate uses the rsaEncryption OID or the RSASSA-PSS OID.

It would seem that testssl.sh already implicitly indicates which code points (RSA-PSS-RSAE or RSA-PSS-PSS) are accepted. If `run_fs()` indicates that RSA-PSS+SHA256 is supported and `run_server_defaults()` shows a certificate with RSA key, then it is RSA-PSS-RSAE. If `run_server_defaults()` shows a certificate with an RSASSA-PSS key, then it is RSA-PSS-PSS.

While I can see the benefit in making it more clear by indicating which one it is in the `run_fs()` output, I really don't like 
"RSA-PSS-RSAE+SHA256" and "RSA-PSS-PSS+SHA256." I had to go to the IANA registry and then to RFC 8446 to find out what they meant. However, I there's probably not a short string that clearly indicates that there are two "different" signature algorithms that only differ in the OID used to identify the public key type in the certificate.

dcooper16:
In testing this issue I discovered in problem with `get_server_certificate()`. It seems that when testing with TLS 1.3 it doesn't include the RSA-PSS-PSS code points in the signature algorithms extension when using either $OPENSSL or `tls_sockets()`. So, that is something that will need to be fixed. Otherwise, testssl.sh will miss a certificate with an RSASSA-PSS subject public key if the server only supports TLS 1.3.

I also tried testing on a related issue. TLS now has two sets of code points for the Brainpool curves, one for use with TLS 1.2 and earlier (brainpoolP256r1, brainpoolP384r1, brainpoolP512r1) and one for use with TLS 1.3 (brainpoolP256r1tls13, brainpoolP384r1tls13, brainpoolP512r1tls13). I was wondering what testssl.sh displayed for these. I discovered, however, the testssl.sh does not yet support the TLS 1.3 Brainpool curves.

So, two more issues that will require PRs.

drwetter:
Sounds like it's good to merge now at least, @dcooper16 ?

dcooper16:
Merging this PR is fine with me.

drwetter:
Thanks for the PR @Odinmylord !

PR #2352, opened 2023-03-29
Find RSASSA-PSS certificates with TLS 1.3

dcooper16:
This PR fixes `get_server_certificate()` so that it can obtain certificates with RSASSA-PSS subject public keys over TLS 1.3. It also adds X+SHA512 to the list of supported signature algorithms sent when using OpenSSL.


drwetter:
Thanks again!

PR #2353, opened 2023-03-29
Add support for brainpool curves with TLS 1.3

dcooper16:
This PR adds support for the curves brainpoolP256r1tls13, brainpoolP384r1tls13, and brainpoolP512r1tls13.

Based on #2346, I included the "tls13" in the names that used, even though brainpoolP256r1 and brainpoolP256r1tls13 are both the same curve, just for use with different TLS versions. This is the reason for lines 10743 - 10745. OpenSSL reports the curve as just brainpoolP...r1, even if it is the brainpoolP...r1tls13 code point, so I need to look at which version of TLS was negotiated to determine which code point was used.

I was not able to test utils/update_client_sim_data.pl, but the change that I made to it seems straightforward.

### DIFF
--- a/etc/curves.txt
+++ b/etc/curves.txt
@@ -28,4 +28,7 @@
 27, brainpoolP384r1, 
 28, brainpoolP512r1, 
 29, curve25519,
-30, curve448
+30, curve448,
+31, brainpoolP256r1tls13,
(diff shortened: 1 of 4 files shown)


drwetter:
Thanks a lot, David!

PR #2357, opened 2023-04-26
Extended the dns caa rr functionality

kasper-hald:
Extended the dns caa rr functionality to handle: 
1) That the caa rr points to the caa rr of another url. 
2) That we climb the dns tree in case no caa rr is found (as instructed in rfc 8659). 

If a scan towards translate.google.com is performed, both 1) and 2) is activated. 
The first caa rr returned will point towards www3.l.google.com (activating 1), and no caa rr is found for www3.l.google.com or l.google.com (thus activating 2) ), but one is found for google.com.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -21126,18 +21126,20 @@ get_caa_rr_record() {
           #caa_property_value="$(awk '{ print $3 }' <<< "$raw_caa")"
           safe_echo "$(sort <<< "$(awk '{ print $2"="$3 }' <<< "$raw_caa")")"
           return 0
-     elif [[ "$(echo "$raw_caa" | sed 's/[[:space:]]*$//g')" == *. ]]; then
+     elif [[ "$(get_last_char "$(strip_spaces "$raw_caa")")" == "." ]]; then
           # trimmed caa ends with a . which indicates, that the caa should be found by querying the caa for the returned url (remove . at the end to align with the normal call convention)
-          NEW_CAA_HOST="$(echo "$raw_caa" | sed 's/[[:space:]]*$//g' | sed 's/.$//')"
+          TRIMMED_CAA_HOST="$(strip_spaces "$raw_caa")"
+          NEW_CAA_HOST="${TRIMMED_CAA_HOST:0:-1}"
           outln "CAA for $1 points to the CAA for $NEW_CAA_HOST"
-          get_caa_rr_record $NEW_CAA_HOST
+          get_caa_rr_record "$NEW_CAA_HOST"
           return $?
      else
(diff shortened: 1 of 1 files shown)


drwetter:
Hi @kasper-hald ,

thanks. Not tried yet.

Code leaves spaces for improvements though:

* echo "$str" | somecmd ---> somecmd <<< "$str"
* prefer bash pattern substitution over  sed
* if sed needs to be used avoid multiple 'sed' commands  by using '-e pattern -e pattern ' 
* make use of unoiveral helper functions (like ``get_last_char()`` , ``count_lines()`` and maybe ``strip_spaces()`` or ``count_chars()``
* quote vars unless you know for sure they are safe -->  ``"$NEW_CAA_HOST"``

kasper-hald:
Hi @drwetter 
Added a commit which should address the review comments above.
I do not have access to a MAC OS machine, so I am not sure if the proposed code is able to run there.

drwetter:
Thanks. There are still some issues with the code. But before we work on that: Do you have better examples?

* l.google.com doesn't resolve for me.
* www3.l.google.com is probably not meant to serve as an example (see below). Also it seems to have a DNS CAA record (first: unpatched, second: with your PR):


image


image

kasper-hald:
Hi,
Another example could be danskebank.mitid.dk or nordea.mitid.dk

Running the current version of testssl.sh will give the dns caa for the mitid.dk domain, and not the dns caa for what is pointed at:
danskebank.mitid.dk --> mitid-broker.danskebank.com --> danskebank.com (this contains different issuers than mitid.dk)
nordea.mitid.dk --> identify.nordea.com --> nordea.com (this contains different issuers than mitid.dk)

I can see, that ssllabs return the same dns caa as testssl.sh, thus not following the cname returned in the first query.
I might be reading rfc 8659 and rfc 1034 wrong, since I want to follow the cname returned in the initial query, and not just throw it away and climb the dns tree of the original dns name.

On another note, then this digging has also led me to the conclusion, that the current PR is not the way to solve "the error", if a such should actually be there.

In case you believe that the current dns caa functionality implements the rfc's correctly, this PR should just be deleted. 
Otherwise, I can try to "start over" and implement a solution that will use the cname as the next dns to query (instead of climbing the dns tree)

drwetter:
Sorry for letting this sit. 

The idea is good and I would support that. Two points though:

* I myself need to read the RFCs again
* the code quality still needs one or another touch (e.g. single square brackets, slicing  with var:0:-1 doesn't work with MacOS)

@dcooper16 : have you a strong opinion on the RFC(s)?

drwetter:
Also

image

drwetter:
Thanks, but there are too many issues in this PR, so I am closing this. I'll filed a feature request though, see above

Issue #2359, opened 2023-05-12
Test for secure renegotiation fails for server that uses SNI


mum-viadee:
I am running version 3.2rc2 from https://testssl.sh/dev/  (commit aa5235e65826ae1104a3ef3d0cba208ef4e4b548)

Couldn't find anything related in open issues

**Command line / docker command to reproduce**
./testssl.sh --openssl /usr/bin/openssl  -R  www.megamartin.de 

> Secure Renegotiation (RFC 5746)          **OpenSSL handshake didn't succeed**

**Expected behavior**

> Secure Renegotiation (RFC 5746)           **supported (OK)**

**Your system (please complete the following information):**
 - OS: Kali GNU/Linux Rolling
 - Platform: Linux 6.1.0-kali7-amd64 x86_64 
 - OpenSSL: OpenSSL 3.0.8 7 Feb 2023 (Library: OpenSSL 3.0.8 7 Feb 2023)

**Additional context**
The internal openssl call for testing secure renegotiation needs the SNI attribute, otherwise openssl fails with return code 1. That's the reason the test fails with the error message "OpenSSL handshake didn't succeed".

PR #2360, opened 2023-05-12
Secure renegotiations tests need servername for servers that use SNI

mum-viadee:
Secure renegotiations tests need servername for servers that use SNI. Otherwise openssl fails with return code 1. 
See https://github.com/drwetter/testssl.sh/issues/2359

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -16982,7 +16982,7 @@ run_renego() {
      else
           # We need up to two tries here, as some LiteSpeed servers don't answer on "R" and block. Thus first try in the background
           # msg enables us to look deeper into it while debugging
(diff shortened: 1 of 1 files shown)


drwetter:
Hi,

okay & thanks. A server with noservername.... I'll wait until the CI run passes.

Cheers, Dirk

drwetter:
I don't know what actually happened. Maybe there was a prob in front of the keyboard.

I wanted to change the comment into "In cases where there's no default host configured we need SNI here as otherwise openssl then would return an error and the test will fail"

Could you clean that up ? Sorry!

mum-viadee:
Hi @drwetter,
I had the same issue before when using the github online editor. After changing just one line the file looked like everything else was changed, too. 

Your commit was reverted and the comment was added again. 

Greeting
  Martin

drwetter:
Thanks, @mum-viadee !

Also in my case it was the online editor. :-/

PR #2375, opened 2023-07-03
Fix line endings in etc/curves-mapping.txt

dcooper16:
The file etc/curves-mapping.txt currently has CRLF line endings and this seems to cause problems with git since .gitattributes now specifies LF line endings for .txt files. This commit changes the line endings for the file in order to (hopefully) prevent problems with git.


drwetter:
Thanks, David! 

I kinda of thought that'll happen automatically when dealing next time with such a file. But while looking @ your PR I encountered some handling problem myself ;-/

PR #2383, opened 2023-08-28
Fix another grep-3.8 warning on needlessly escaped exclamation mark.

ghen2:
`grep: warning: stray \ before !`
See also #2241 and #2242.


drwetter:
thx

PR #2390, opened 2023-09-07
Support OpenSSL with no TLSv1

dcooper16:
RFC 8996, _Deprecating TLS 1.0 and TLS 1.1_, states that TLS clients MUST NOT send a TLS 1.0 or TLS 1.1 ClientHello and MUST respond to a TLS 1.0 or TLS 1.1 ServerHello with a "protocol_version" alert.

At the moment, all versions of OpenSSL support TLS 1.0, TLS 1.1, and TLS 1.2. However, TLS 1.0 and TLS 1.1 are disabled in LibreSSL 3.8.1 and it is possible to compile OpenSSL without support for these protocols (using the configure options no-tls1, no-tls1_1, and no-tls1_2). This commit adds support for versions of $OPENSSL that do not support TLS 1.0 or TLS 1.1.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -328,6 +328,9 @@ HAS_CURVES=false
 OSSL_SUPPORTED_CURVES=""
 HAS_SSL2=false
 HAS_SSL3=false
+HAS_TLS1=false
+HAS_TLS11=false
+HAS_TLS12=false
 HAS_TLS13=false
 HAS_X448=false
 HAS_X25519=false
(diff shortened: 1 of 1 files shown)


dcooper16:
I created this and posted it to https://github.com/dcooper16/testssl.sh/commits/no_tls1 some time ago, but didn't create the PR until now since I didn't think it would be needed for a while. However, I just noticed that LibreSSL 3.8.1 came out on August 31 and it disables TLS 1.0 and TLS 1.1.

I believe this PR is ready to merge. However, there is still a bit of work to do for `run_tls_fallback_scsv()`. In the case that sockets are not being used and a "high" protocol can not be found, it become more difficult to determine what message to provide as fewer and fewer TLS versions can be tested. However, this is an edge case, and this PR still greatly improves the way that testssl.sh runs (particularly in `--ssl-native` mode) if $OPENSSL does not support TLS 1.0 and TLS 1.1.

drwetter:
Hi @dcooper16 ,

looks good, thanks a lot! 

It might be also that some distributors disabled TLS 1.0/1.1 in /etc/ssl/openssl (DEFAULT@SECLEVEL). Shouldn't concern us though as we don't provide that config, IIRC.

In the future it looks like it makes coverage for openssl based tests more complicated: we either need to do everything in sockets < TLS 1.2 if we only want one binary (off the shelf) or we need two binaries: the old one and a one which also supports 1.3. Or we need a news set of binaries where we reenable those protocols . 

Thx, Dirk

PR #2393, opened 2023-09-11
Fix LibreSSL returns 0 on failure

dcooper16:
In some cases LibreSSL's s_client returns a status of 0 even when the connection fails. I have not been able to figure out under what circumstances it happens. However, when it does happen there always seems to be a line in the output of the form "SSL handshake has read 0 bytes and written ... bytes". This PR fixes the problem by not treating a connection as successful if the string "SSL handshake has read 0 bytes and written" is present. This fix seem safe as the connection could not have been successful is no data was received from the server.

So far, I have only encountered this error when testing against the local CUPS server on my computer (Ubuntu 22.04):
```
testssl.sh --openssl=<path to LibreSSL binary> --ssl-native --std 127.0.0.1:631
```

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -7756,7 +7756,7 @@ sclient_connect_successful() {
      local re='Master-Key: ([^\
 ]*)'
 
-     [[ $1 -eq 0 ]] && connect_success=true
+     [[ $1 -eq 0 ]] && [[ ! "$server_hello" =~ SSL\ handshake\ has\ read\ 0\ bytes\ and\ written ]] && connect_success=true
      if ! "$connect_success" && [[ "$server_hello" =~ $re ]]; then
           [[ -n "${BASH_REMATCH[1]}" ]] && connect_success=true
      fi
@@ -21570,7 +21570,7 @@ sclient_auth() {
 ]*)'
      local connect_success=false
 
-     [[ $1 -eq 0 ]] && connect_success=true
+     [[ $1 -eq 0 ]] && [[ ! "$server_hello" =~ SSL\ handshake\ has\ read\ 0\ bytes\ and\ written ]] && connect_success=true
 
      ! "$connect_success" && [[ "$server_hello" =~ $re ]] && \
           [[ -n "${BASH_REMATCH[1]}" ]] && connect_success=true


drwetter:
Hi David,

thanks! 

I have still to find a host in the LAN here to reproduce your finding. Debian 11's cups e.g. didn't show that with the native LibreSSL (3.3.6?) on a Mac. 

So at the moment it appears too much of an edge case to me and I am tending not to merge it, unless one spots more cases. Adding this to a central function for an edge case may be too much.

However give me a few days for more research.

Cheers, Dirk

drwetter:
Hi David,

did you spot where exactly the difference is, e.g. in the output? or doesn't it then work at all?

All I spotted so far was like this (using binaries supplied vs LibreSSL 3.7.3 compiled from source)

image

dcooper16:
Hi Dirk,

The issue only appears in `--ssl-native` mode, and the only differences I see (other than places where one indicates "Local Problem") are in `--std` and `--client-simulation`. In the case of `--std`, using LibreSSL results in categories being marked as "offered" even though they are not. In the case of `--client-simulation` the issue is with IE 8 XP, where it is shown as connecting with TLSv1.0, but no cipher suite is listed.

Screenshot_20230920_102718
Screenshot_20230920_103451

Here is an example of a command from `sub_cipherlists()` that is failing:
```
> ./libressl-3.7.3/apps/openssl/openssl s_client -connect 127.0.0.1:631 -no_ssl2 -noservername -no_comp -no_tls1_3 -cipher @SECLEVEL=0:NULL:eNULL
WARNING: can't open config file: /usr/local/etc/ssl/openssl.cnf
CONNECTED(00000003)
read:errno=0
---
no peer certificate available
---
---
SSL handshake has read 0 bytes and written 120 bytes
---
New, (NONE), Cipher is (NONE)
Compression: NONE
Expansion: NONE
No ALPN negotiated
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : 0000
    Session-ID: 
    Session-ID-ctx: 
    Master-Key: 
    Start Time: 1695232066
    Timeout   : 7200 (sec)
    Verify return code: 0 (ok)
---
> echo $?
0
```

drwetter:
seems I overlooked "--ssl-native", sorry. Let me look again into it.

drwetter:
Hi @dcooper16,

I am not so fond any more of ``--ssl-native`` as it is now. That was a hack from the early days. 

Both openssl and libressl throw a lot of warnings when using ``--ssl-native`` and only considering that makes testssl.sh quite unusable. Also both hang when using forward secrecy for a longer moment. I see your point  though of the fix but in the end it tries to fix one of many problems, albeit a false detection is something we do not want. And it keeps the option alive we shouldn't use anymore.

The best solution IMO would be removing ``--ssl-native`` completely form the code. The next one would be remove the cmdline option, next would be a warning which has to be confirmed with YES.

If you or someone else can help with the best solution, that would be much appreciated. I'll lack time. Otherwise I'll go for option 2 or 3 and do 1 later in 3.3dev.

Thus: thanks but I'd rather close this.

dcooper16:
Hi Dirk,

Closing this PR is fine, as it does seem like a very niche case. My concern was just that `$OPENSSL s_client` was returning 0 even though the connection failed, and I did not know under what circumstances that could happen. So far, I have only noticed it happening when testing against one server in `--ssl-native` mode. However, not knowing why it happened, I had no assurance that it wouldn't affect normal testing (not in `--ssl-native` mode) against other servers.

drwetter:
Hi David,

appreciate your concerns! Let's keep that in mind.

PR #2400, opened 2023-09-20
Add --server-preference alias to documentation

dekkers:
Both `--server-preference` and `--preference` are accepted as options, but only `--server-preference` is listed in `--help` and only `--preference` in the documetation. This also adds the `--server-preference` option to the documentation.


drwetter:
Good catch, thanks!

PR #2411, opened 2023-10-07
Improve detection for POP3, IMAP and SMTP services.

ghen2:
"POP3" is a common banner (so we drop some more specific ones containing this word) as well as "IMAPrev1" (case insensitive).
Extended regular expression for SMTP was missing -E.


drwetter:
Thanks a lot @ghen2 !

Issue #2378, opened 2023-08-01
[BUG / possible BUG] DNS resolution fails in a container


mythi:
I'm checking my Kubernetes webhook TLS [services](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/) using `testssl.sh` deployed. Things used to work with `drwetter/testssl.sh:3.0` but now fails with `drwetter/testssl.sh:3.2` (and `:latest`):

```
Fatal error: No IPv4/IPv6 address(es) for "<my service>" available
```

**Before you open an issue please check which version you are running and whether it is the latest in stable / dev branch**
I am running version ( ``git log | head -1``  / ``testssl.sh -v | grep from``): `drwetter/testssl.sh:3.2`

**Before you open an issue please whether this is a known problem by searching the issues**
couldn't find anything 

**Command line / docker command to reproduce**
`testssl.sh --openssl=/usr/bin/openssl --mapping iana -s -f -p -P -U  <my service>`

**Expected behavior**
`testssl.sh` to succeed the same way it does with `drwetter/testssl.sh:3.0`.

**Your system (please complete the following information):**
 - OS:  `drwetter/testssl.sh:3.2`
 
**Additional context**

drwetter:
Hi @mythi ,

not sure whether you scan from within the k8s cluster or outside? From the inside: it's hard to tell what's going wrong as it depends on your DNS service and how you set it up. testssl.sh depends on a proper working resolver.

As a litmus test: Does (ping|host|dig|netcat) return the correct IP address?

There was one bigger DNS related change I recall btw 3.0 and 3.2 (d001bba86bbeee07ff2ced2dc146603f71f76a53) but as you don't use a proxy that shouldn't affect you.

mythi:
> From the inside: it's hard to tell what's going wrong as it depends on your DNS service and how you set it up. testssl.sh depends on a proper working resolver.

From inside. I'm assuming the resolver works properly since it's the exact same cluster that works with `3.0` image but not when I move to `3.2`. I'll take a closer look on https://github.com/drwetter/testssl.sh/commit/d001bba86bbeee07ff2ced2dc146603f71f76a53 and try out the utils you suggested.

drwetter:
if you have DNS query logs that would be also a useful source to look at

mythi:
[...]

PR #2417, opened 2023-10-10
Added drill command to get_a[aaa]_record() for local hostname

vsbs-david:
When the hostname ends with `.local`, the functions `get_a_record` and `get_aaaa_record` expect that command `avahi-resolve` or `dig` are available to resolve, but they are not in the docker image. However we have `drill` available, so we can use it as an alternative as mentioned in https://github.com/drwetter/testssl.sh/issues/2378#issuecomment-1753250043

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -21095,7 +21095,7 @@ get_a_record() {
           else
                fatal "Local hostname given but neither 'avahi-resolve', 'dig' nor 'drill' is available." $ERR_DNSBIN
           fi
-          [[ -z "$ip4" && "$DEBUG" -ge 1 ]] && echo ".local IP address requested but mDNS resolution failed" 1>&2
+          [[ -z "$ip4" ]] && debugme echo ".local IP address requested but mDNS resolution failed"
      fi
      if [[ -z "$ip4" ]] && "$HAS_DIG"; then
           ip4=$(filter_ip4_address $(dig +search $DIG_R +short +timeout=2 +tries=2 $noidnout -t a "$1" 2>/dev/null | awk '/^[0-9]/ { print $1 }'))
@@ -21142,7 +21142,7 @@ get_aaaa_record() {
                else
                     fatal "Local hostname given but neither 'avahi-resolve', 'dig' nor 'drill' is available." $ERR_DNSBIN
                fi
-               [[ -z "$ip6" && "$DEBUG" -ge 1 ]] && echo ".local IP address requested but mDNS resolution failed" 1>&2
+               [[ -z "$ip6" ]] && debugme echo ".local IP address requested but mDNS resolution failed"
           fi
      fi
      if [[ -z "$ip6" ]]; then
--- a/testssl.sh
+++ b/testssl.sh
@@ -5612,7 +5612,7 @@ run_protocols() {
                add_proto_offered tls1 no
                if [[ -z $latest_supported ]]; then
                     outln
-                    fileout "$jsonID" "INFO" "not offered"       # neither good or bad
+                    fileout "$jsonID" "INFO" "not offered"       # neither good nor bad
                else
                     prln_svrty_critical " -- connection failed rather than downgrading to $latest_supported_string (NOT ok)"
                     fileout "$jsonID" "CRITICAL" "connection failed rather than downgrading to $latest_supported_string"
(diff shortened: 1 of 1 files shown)


vsbs-david:
For testing in my case I don't have hosts with name `.local` for multicast DNS, but for standard DNS. For example, `dig hostname.local` shows a warning like this:
```
;; WARNING: .local is reserved for Multicast DNS
;; You are currently testing what happens when an mDNS query is leaked to DNS
```
But the answer section has the IP correctly. With `drill` there is no warning though, and the IP appears correctly too.
With the commands using the address `224.0.0.251` for mDNS as mentioned in the issue, there are errors like these:
```
# drill @224.0.0.251 -p 5353 <redacted>.local
Error: error sending query: Could not send or receive, because of network error
# dig @224.0.0.251 -p 5353 +short -t a +notcp <redacted>.local
;; communications error to 224.0.0.251#5353: timed out
;; communications error to 224.0.0.251#5353: timed out
;; communications error to 224.0.0.251#5353: timed out
;; no servers could be reached
```
Then, when running testssl.sh via docker, with `docker run --rm -t vsbs-david/testssl.sh hostname.local`, the message `Error: error sending query: Could not send or receive, because of network error` appears before using standard DNS (as the functions use it when the mDNS query has not returned IP addresses) and then the test run successfully.

So, this would work for our case, but maybe we should handle these errors from `dig` or `drill` to not appear directly in testssl.sh. Also it would be needed to test an actual mDNS use case. Any thoughts?
Thanks

drwetter:
> So, this would work for our case, but maybe we should handle these errors from dig or drill to not appear directly in testssl.sh. 

Probably it would help to redirect stderr to /dev/null for drill as for dig with the inner round brackets. 

As a result of an error $ip6 and $ip4 should be empty, so this is where we could add a message like ``[[ -z "$ip4" ]] && fatal ".local IP address requested but mDNS resolution failed"`` -- e.g. after ``else .. fi``.

drwetter:
PS: _neither /n/or_ probably would sound better

vsbs-david:
Ok thanks, that works but I'd prefer that the script continues with a standard DNS query, instead of forcefully existing with fatal error. To be honest I was not sure if having .local hosts in our network but without using mDNS is correct from the point of view of standards, I'm having a look and apparently it can be fine to use both multicast and unicast DNS according to this RFC: https://datatracker.ietf.org/doc/html/rfc6762#:~:text=Implementers%20MAY%20choose%20to%20look%20up%20such%20names%20concurrently%20via%20other%0A%20%20%20mechanisms%20(e.g.%2C%20Unicast%20DNS)%20and%20coalesce%20the%20results%20in%20some%0A%20%20%20fashion.

So if you agree I was going to add a line like you mentioned but with warning instead of fatal, how could it be better to do it with the current functions?

drwetter:
Good point. Seems I remembered that differently,  (as your link didn't work for me at least as intended: _Implementers MAY choose to look up such names concurrently via other  mechanisms (e.g., Unicast DNS) and coalesce the results in some  fashion._  

Sigh, sounds unusually a bit wishy-washy to me but so be it.

But then at least lets add a debug statement like ``"$DEBUG" && echo ".local IP address requested but mDNS resolution failed"``

vsbs-david:
@drwetter I pushed one more commit with the changes we talked about, but the CI check seems to fail because of network error, particularly when connecting to badssl.com (104.154.89.105), could it have been a problem with that service itself? Maybe can we retry the CI action to double check? Thanks

testssl.sh
```diff
@@ -21090,9 +21090,12 @@ get_a_record() {
                ip4=$(filter_ip4_address $(avahi-resolve -4 -n "$1" 2>/dev/null | awk '{ print $2 }'))
           elif "$HAS_DIG"; then
                ip4=$(filter_ip4_address $(dig $DIG_R @224.0.0.251 -p 5353 +short -t a +notcp "$1" 2>/dev/null | sed '/^;;/d'))
+          elif "$HAS_DRILL"; then
+               ip4=$(filter_ip4_address $(drill @224.0.0.251 -p 5353 "$1" 2>/dev/null | awk '/ANSWER SECTION/,/AUTHORITY SECTION/ { print $NF }' | awk '/^[0-9]/'))
           else
-               fatal "Local hostname given but no 'avahi-resolve' or 'dig' available." $ERR_DNSBIN
+               fatal "Local hostname given but neither 'avahi-resolve', 'dig' nor 'drill' is available." $ERR_DNSBIN
           fi
+          [[ -z "$ip4" && "$DEBUG" -ge 1 ]] && echo ".local IP address requested but mDNS resolution failed" 1>&2
```
drwetter, line 21098: better would be ``"$DEBUG" -ge 2`` here an below. As debug level one is too much for this.

Even better: ``[[ -z "ipv4" ]] && debugme echo ....``

drwetter:
> but the CI check seems to fail because of network error

yes, thanks. Experienced the same yesterday too and re-run the job. If you push the suggested change it'll run again and we'll see whether I have to push the button again

testssl.sh
vsbs-david: That makes sense, updated now to use `debugme` function, thanks

drwetter:
Thanks a lot, @vsbs-david !

PR #2428, opened 2023-10-19
add shellcheck to CI

MFTabriz:
- add shellcheck test [initially] with `--severity=error` to CI.
- fix errors!

shellcheck report for f3c7eb43336889711b15bfff6b7be833f745cb04 can be accessed from [here](https://github.com/MFTabriz/testssl.sh/actions/runs/6572471937/job/17853649730).

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,20 @@
+---
+name: Shellcheck
+on: [push, pull_request]
+permissions:
+  contents: read
+
+jobs:
+  shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-latest
+    env:
+      LANG: C.UTF-8
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install shellcheck
+        run: sudo apt update && sudo apt install -y shellcheck && shellcheck --version
+
+      - name: Check shell scripts
+        run: shellcheck -x -P "SCRIPTDIR"/utils --severity=error testssl.sh
--- a/testssl.sh
+++ b/testssl.sh
@@ -17325,19 +17325,19 @@ run_breach() {
 
                # Final verdict (if not happened preemptively before). We reuse $detected_compression here
                detected_compression=""
-               if [[ ${has_compression[@]} =~ warn ]]; then
+               if [[ ${has_compression[*]} =~ warn ]]; then
                     # warn_empty / warn_stalled
-                    if [[ ${has_compression[@]} =~ warn_empty ]]; then
-                         pr_warning "At least 1/4 checks failed (HTTP header request was empty, debug: ${has_compression[@]}"
-                         out ", debug: ${has_compression[@]})"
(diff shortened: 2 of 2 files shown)


MFTabriz:
@drwetter in [L22185](https://github.com/drwetter/testssl.sh/blob/f3c7eb43336889711b15bfff6b7be833f745cb04/testssl.sh#L22185) `PROXY` is defined as a string and not a variable. I was going to fix it but before that I need to know how that part is being tested?

drwetter:
Thanks!

To add this is basically a very good idea however we need either to fix the errors or to mark them so that the CI passes. 

> In [L22185](https://github.com/drwetter/testssl.sh/blob/f3c7eb43336889711b15bfff6b7be833f745cb04/testssl.sh#L22185) PROXY is defined as a string and not a variable. I was going to fix it but before that I need to know how that part is being tested?

Not sure I can follow you. ``$PROXY`` is a global. It is meant to be used as a string variable as it will contain, when defined,  ``$PROXYIP:$PROXPORT`` 

What do you mean by tested?

MFTabriz:
> To add this is basically a very good idea however we need either to fix the errors or to mark them so that the CI passes.

Yes! This is why this PR is still marked as a draft. :D

> Not sure I can follow you. `$PROXY` is a global. It is meant to be used as a string variable as it will contain, when defined, `$PROXYIP:$PROXPORT`

`$PROXY` is indeed a variable but in [this line]( https://github.com/drwetter/testssl.sh/blob/f3c7eb43336889711b15bfff6b7be833f745cb04/testssl.sh#L22185) we are using a literal string `"PROXY"` not the variable `"$PROXY"`. `[[ -n "PROXY" ]]` will be always evaluated as true!

> What do you mean by tested?

As this issue was not caught earlier, I’d expect the functionality testing is not working correctly [or maybe we are missing the coverage]. I’d rather fix the test first and then the code itself to check & document test functionality.

drwetter:
> ` `[[ -n "PROXY" ]]`` will be always evaluated as true!

Oops, thanks for the clarification. :-)  PR pending. 

$DNS_VIA_PROXY needs also be true and it affects only the output. So, no, the output wasn´t tested in CI. Guess I tested that manually but obviously my eyes weren´t open enough.

MFTabriz:
@drwetter I rebased this PR on #2436 and [fixed the shellcheck errors](https://github.com/drwetter/testssl.sh/commit/4886b85e10d28f9d6d010530d54dc91217ff05a1).

drwetter:
Ok, thanks!

CI went fine. However in some instances I am not sure yet whether the fixes are correct when changing from [@] to [*] Have you looked at the contexts whether it's correct to remove the expansion?

drwetter:
Hi @MFTabriz ,

pls see my last comment.

drwetter:
Hi @MFTabriz,

the errors should be gone with #2922 now.

Mind to work on this thing again? Best would be just to toss this and just do the GHA. 

Two remarks:
* sudo should not be necessary.
* ubuntu-latest should be rather ubuntu-24.04 , in-line with the other GHAs

Avoiding shellcheck errors would be a great thing and move this project another step forward.

Issue #2356, opened 2023-04-20
[FEATURE/BUG] openssl compiled without sieve support


nextgens:
**Before you open an issue please check which version you are running and whether it is the latest in stable / dev branch**

I am running version 3.2rc2

**Before you open an issue please whether this is a known problem by searching the issues**

Is couldn't find anything 

**Command line / docker command to reproduce**

docker run --rm -it drwetter/testssl.sh -t sieve $SERVER:4190
...
Fatal error: Your /home/testssl/bin/openssl.Linux.x86_64 does not support the "-starttls sieve" option

**Expected behavior**

I would expect openssl to be compiled with support for sieve

**Your system (please complete the following information):**
current 'latest' on docker:
Digest: sha256:363d162b04a483826bb91c2e04c3498d16d60b3a953fd599b3cb0e8dc9076eb3

drwetter:
It's more a missing feature in the currently supplied version as I haven't backported that.

``--openssl=/usr/bin/openssl`` should do it though.

The supplied [openssl-bad](https://github.com/drwetter/openssl-1.0.2.bad) will be deprecated after 3.2 and may be removed or not (fallback). If not: there's a good chance sieve will be backported. Thus I'll leave this open.

nextgens:
Thank you. I can confirm that the workaround works.

I've initially tagged it "bug" because the help message does suggest that it should work out of the box:
```
  and [options] is/are:

     -t, --starttls <protocol>     Does a run against a STARTTLS enabled service which is one of ftp, smtp, lmtp, pop3, imap,
                                   sieve, xmpp, xmpp-server, telnet, ldap, nntp, postgres, mysql
```

drwetter:
Got your point 😬 

I should have backported this last time I (and others) compiled the binaries (September 2022) after the implementation was contributed by @k0lter. There's a unit test too which doesn't hiccup but should.

That somehow need to be fixed. Need to sleep over it how though.

drwetter:
Note: See also #2607

drwetter:
see https://github.com/testssl/openssl-1.0.2.bad/issues/2

drwetter:
FYI: Actually there was a problem in the code of testssl.sh, too. Means the STARTTLS handshake message from the server was not correctly interpreted . That is fixed now.

Also using this  https://testssl.sh/openssl-1.0.2k-bad/openssl.Linux.x86_64 binary now testing sieve works. (Binary will be pushed to github when testing is done).

[...]

PR #2432, opened 2023-10-29
add Linux aarch64 support

rampageX:
Tested on Oracle Cloud ARM64 VPS.


drwetter:
Hi,

thanks! I had this architecture also in mind but Oracle somehow didn´t want me to give me a cloud vm (for free?) a while back.

Two points though:
* Don´t know whether the php file was intentionally included (seems so) but this generally cannot be merged for several reasons. Unfortunately it includes a perfect example how not to sanitize user params : domain. 
* Then: in the supplied openssl-bad binaries I missed to complete the sieve support, see  #2356.  I (or somebody else) would need to fix this before any new binaries. This is planned for 3.2 and maybe there'll be a last 3.0.9. to include this.

Thus I am closing this and hope you can help when sieve support is fully accomplished.

Issue #2445, opened 2023-12-01
[BUG] openssl core dumps under Ubuntu 22.04 when specifying the network service name instead of the port


sanderjo:
Wrong input, so PEBKAC, but core dumped, so worth reporting?

**Before you open an issue please check which version you are running and whether it is the latest in stable / dev branch**
I am running version 
```

commit 1a9a486474cacd377c444b6fc08a9e754ef95480
testssl.sh       3.2rc3 from https://testssl.sh/dev/

```

**Before you open an issue please whether this is a known problem by searching the issues**

Probably **not** related to https://github.com/drwetter/testssl.sh/issues/1934

**Command line / docker command to reproduce**

```
./testssl.sh news.newshosting.com:nntps
```

No problem when I use:
```
./testssl.sh news.newshosting.com:563
```

**Expected behavior**
Accept the `:nntps`, or reject it without core dump.

**Your system (please complete the following information):**
 - OS: Ubuntu 22.04.3 LTS
 - Platform:  Linux 6.2.0-37-generic x86_64
 - OpenSSL: ./bin/openssl.Linux.x86_64

**Additional context**

```
sander@zwart2204:~/git/poging2_testssl$ ./testssl.sh news.newshosting.com:nntps

###########################################################
    testssl.sh       3.2rc3 from https://testssl.sh/dev/
    (1a9a486 2023-11-10 19:41:08)

      This program is free software. Distribution and
             modification under GPLv2 permitted.
      USAGE w/o ANY WARRANTY. USE IT AT YOUR OWN RISK!

       Please file bugs @ https://testssl.sh/bugs/

###########################################################

 Using "OpenSSL 1.0.2-bad (1.0.2k-dev)" [~179 ciphers]
 on zwart2204:./bin/openssl.Linux.x86_64
 (built: "Sep  1 14:03:44 2022", platform: "linux-x86_64")


Testing all IPv4 addresses (port nntps): 85.12.62.225 185.90.196.65 85.12.62.251 185.90.196.97 185.90.196.129
--------------------------------------------------------------------------------------------------------------
 Start 2023-12-01 21:32:04                -->> 85.12.62.225:nntps (news.newshosting.com) <<--

 Further IP addresses:   185.90.196.97 185.90.196.65 185.90.196.129 85.12.62.251 
 rDNS (85.12.62.225):    --
[...]

PR #2446, opened 2023-12-04
convert service name to port number

sanderjo:
Works with

`$ ./testssl.sh news.newshosting.com:nntps
`

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -20941,6 +20941,11 @@ parse_hn_port() {
           grep -q ':' <<< "$NODE" && \
                PORT=$(sed 's/^.*\://' <<< "$NODE") && NODE=$(sed 's/\:.*$//' <<< "$NODE")
      fi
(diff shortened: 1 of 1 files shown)


sanderjo:
Closes https://github.com/drwetter/testssl.sh/issues/2445

drwetter:
Hi,
thanks first. Doesn't work on MacOS though.

```
 /tmp  getent services | wc -l
       0
 /tmp  
```

drwetter:
MacOS has /etc/services though, don't know about WSL? We need to put some thought in this, so that the code doesn't become ugly and it works on every OS supported and on docker.

And we work around a weird bug in the old openssl we're using. I anyway wanted to complete the sieve implementation. Don't know how difficult it is yet to add/backport the service resolution into it.
 
Fallback would be a private LUT (lookup table) , i.e. ``ports2starttls()`` kind of reversed.

drwetter:
see comment in the issue

PR #2447, opened 2023-12-13
Add Brainpool signature algorithms to output

odinmylord:
With the release of openssl 3.2.0 it is possible to use ecdsa_brainpool signature algorithms. At the moment testssl does not recognize them.


drwetter:
Thank you!

PR #2454, opened 2024-01-02
Make cert_keysize output consistent

odinmylord:
These changes make the `cert_keysize` field output more consistent by including the `short_keyAlgo` even if the size of the key is not recognized


drwetter:
Thanks!

PR #2455, opened 2024-01-03
add utf8 support to intermediate cert names

odinmylord:
Since UTF-8 is not used while decoding Intermediate certificates sites like no-subject.badssl.com have an invalid json file as output (in this case the `™` is the issue)


drwetter:
Thanks!

PR #2459, opened 2024-01-12
Secure Client-Initiated Renegotiation : fixes/enhancements

### LINKED ISSUE #2444: [BUG] Secure Client-Initiated Renegotiation test does not detect exponential renegotiation backoff mitigation.
testssl.sh 3.2 branch

ASSUME_HTTP=true testssl.sh -R https://target.fr  (targets to reproduce available on request)

**Expected behavior**
"not vulnerable (OK) -- mitigated" result.

The target is using exponential backoff/throttling between each renegotiation tries.
testssl.sh has no provision to detect this mitigation (compared to the python tool sslyze) and give false positive.

~~Here is a proposition of a simple approach to fix this: #2443~~

Replaced by this PR with more fixes/enhancements https://github.com/drwetter/testssl.sh/pull/2459

Tazmaniac:
Replace https://github.com/drwetter/testssl.sh/pull/2443
and fix https://github.com/drwetter/testssl.sh/issues/2444

This original PR was about false positives on target doing sort of exponential backoff between renegotiation tries.

Here is a proposed simple but efficient way of detecting such mitigation.

Instead of trying to do precise timing measurement, we count the number of successful renegotiation.
As we separate each try by one second without waiting for any result, some try will be lost in case of aggressive exponential slow down between each try even with the command buffering done by openssl.
On the tested targets, the result are pretty good. With the default number of six of tries , only four success. With ten tries, only five are successful.
As a conservative disposition, we only consider the target mitigated if we lost 1/3 or more of tries.

On top of this, tree more fixes/enhancements for the Secure Client-Initiated Renegotiation test:
- there is case of connection hang/block after some renego retry. Implement a loop test timeout without loosing any test feature. This is the most tricky part, but finally work perfectly.
- push the number of retry to 10 for Stormshield based equipment's and others like it is done on sslyse. Some exhibits connections hang/block (for 6<Retry=<10) and need the previous loop timeout.
- decrease the wait from 1s to 0.25s between renego to speedup the loop. The loop timeout stay based on retries*1s.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -17036,6 +17036,13 @@ run_renego() {
           fileout "$jsonID" "WARN" "client x509-based authentication prevents this from being tested"
           sec_client_renego=1
      else
+	  # We will need $ERRFILE for mitigation detection
+	  if [[ $ERRFILE =~ dev.null ]]; then
+	       ERRFILE=$TEMPDIR/errorfile.txt || exit $ERR_FCREATE
+	       restore_errfile=1
+	  else
+	       restore_errfile=0
+	  fi
           # We need up to two tries here, as some LiteSpeed servers don't answer on "R" and block. Thus first try in the background
           # msg enables us to look deeper into it while debugging
           echo R | $OPENSSL s_client $(s_client_options "$proto $BUGS $legacycmd $STARTTLS -connect $NODEIP:$PORT $PROXY $SNI") >$TMPFILE 2>>$ERRFILE &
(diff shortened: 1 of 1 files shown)


drwetter:
> I will rebase when https://github.com/drwetter/testssl.sh/pull/2443 will be merged.

Isn't it better to close #2443  and we work on this? Don't know how complete this one is (except the typo codespell complaint about) .

Tazmaniac:
Yes. Will do that.

drwetter:
Ok, thanks! 👍

Still missing a couple of test cases for the extended test here but the ones I tried works. Thanks!

PR #2460, opened 2024-01-19
Fixed regression in printing results

Tazmaniac:
My last PR introduce a regression in printing the result. The $tmp_result value could be wrongly overwritten when not equal to 0.
And improve it with the effective number of renego before disconnection as it is now tracked.


drwetter:
Thanks, @Tazmaniac

PR #2461, opened 2024-01-20
Add mTLS new feature to support scans with client authentication

akabe1:
This PR adds a new feature to support mutual TLS, a new cmd line option `--mTLS <client cert>` is added to perform scans when the remote server requires client authentication.
It requires that user specifies the filepath of a PEM file containing both the x509 client certificate aand private key (not encrypted), then this file will be used to perform client TLS authentication during the scan.
The PEM file is then used when the binary openssl is called by adding the `-cert <path_to_client_cert>` option to the code of the script testssl.sh.

What is your chnage about?

- [ ] Typo fix
- [x] Documentation update
- [ ] Update of other files

## If it's a code change please check the boxes which are applicable: 
- [x] For the __new feature__ I have made corresponding changes to the documentation and / or to ``help()``

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -389,6 +389,7 @@ XMPP_HOST=""
 PROXYIP=""                              # $PROXYIP:$PROXPORT is your proxy if --proxy is defined ...
 PROXYPORT=""                            # ... and openssl has proxy support
 PROXY=""                                # Once check_proxy() executed it contains $PROXYIP:$PROXPORT
+MTLS=""                                 # mTLS authentication with client certificate and private key
 VULN_COUNT=0
 SERVICE=""                              # Is the server running an HTTP server, SMTP, POP or IMAP?
 URI=""
@@ -2316,6 +2317,12 @@ s_client_options() {
      fi
      # $keyopts may be set as an environment variable to enable client authentication (see PR #1383)
      tm_out "$options $keyopts"
+
+     # In case of mutual TLS authentication is required by the server
+     # Note: the PEM certificate file must contain: client certificate and certificate key (not encrypted)
+     if [[ -n "$MTLS" ]]; then
+          options+=" -cert $MTLS"
+     fi
 }
 
 ###### check code starts here ######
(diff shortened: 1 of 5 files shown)


drwetter:
Thanks a lot for this!

I would feel more comfortable merging this if another person can test it before. Seems for me not possible at the moment. Anybody can give a hand here?

testssl.sh
```diff
@@ -20500,6 +20529,7 @@ tuning / connect options (most also can be preset via environment variables):
      --ids-friendly                skips a few vulnerability checks which may cause IDSs to block the scanning IP
      --phone-out                   allow to contact external servers for CRL download and querying OCSP responder
      --add-ca <CA files|CA dir>    path to <CAdir> with *.pem or a comma separated list of CA files to include in trust check
+     --mtls <CLIENT CERT file>     path to <CLIENT CERT> file, it must be in PEM format and contain client certificate with certificate key (not encrypted)
```
drwetter, line 20532: The line is a bit too long here. It shouldn´t be longer than the one for --color .

Maybe like ``path to <CLIENT CERT> file in PEM format with certificate key, not encrypted``

drwetter:
Related to  #1383 . If merged that would need to be reverted.

drwetter:
Ok, thanks again for the PR!

Seems nobody couldn´t give a hand. Code seems quite good but as another test failed I'd rather label this as beta feature in the next PR. Happy to remove the label if someone can test it thoroughly

dcooper16:
Sorry I didn't have a chance to look at this PR earlier (before it was merged), but I'm overloaded with other projects at the moment.

I haven't been able to try to test this PR, but there is something that seems incorrect. There is code in `s_client_options()` that adds `-cert $MTLS` to all calls to `$OPENSSL s_client` whenever $MTLS is not empty. Then, there is also code in `run_http_header()`, `sub_session_resumption()`, and `sub_breach_helper()` that adds `-cert $MTLS` to calls to `$OPENSSL s_client`. This would seem to add `-cert $MTLS` to every `$OPENSSL s_client` command line, and add it twice when called from `run_http_header()`, `sub_session_resumption()`, and `sub_breach_helper()`.

It may not be a problem to include the `-cert` option twice in the `$OPENSSL s_client` command line. But, the testssl.sh script could be simplified by either using `s_client_options()` to include `-cert $MTLS` in every call to `$OPENSSL s_client` or adding it only where needed (e.g., `run_http_header()`, `sub_session_resumption()`, and `sub_breach_helper()`), but not both.

akabe1:
Hi @dcooper16  and @drwetter ,
I have added the `-cert $MTLS` to these functions on the `$OPENSSL s_client` calls, because the expansion of `$(s_client_options)` does not return the `-cert $MTLS`. Then without this workaround it will be not added at all.

Below some images to highlight it:
 
image

image

dcooper16:
> Hi @dcooper16 and @drwetter ,
> I have added the `-cert $MTLS` to these functions on the `$OPENSSL s_client` calls, because the expansion of `$(s_client_options)` does not return the `-cert $MTLS`. Then without this workaround it will be not added at all.

Hi @akabe1,

Yes, I see that now. The problem is that in `s_client_options()` the new code to add `-cert $MTLS` was added at the end of the function, after the `tm_out "$options"` line. The reason that doesn't work is that `tm_out "$options"` is the line that causes the value of `$options` to be returned. So, any changes that you make to `$options` after the `tm_out "$options"` line are essentially just thrown away.

akabe1:
Hi @dcooper16 ,
now it is clear. 
Will you fix it? Or would you prefer me to make a new PR with the correction instead?

dcooper16:
> Will you fix it? Or would you prefer me to make a new PR with the correction instead?

I'll leave it to you (and @drwetter) to decide whether to remove the code from `s_client_options()`, or to fix `s_client_options()` and remove the code from the other functions, and then make the new PR.

drwetter:
@akabe1 : the best is if you could go ahead.

PR #2467, opened 2024-02-13
MongoDB identification fix

### LINKED ISSUE #2468: [BUG] MongoDB identification
The identification string for MongoDB is too loose.

Some HTTP sites are identified as "MongoDB" protocol:
https://www.mazarsrecrute.fr

The page is about ~2Mo but on one line and match the "MongoDB" keyword.

The proposed PR: https://github.com/drwetter/testssl.sh/pull/2467 fix this with a more precise string match.

Tazmaniac:
The actual code grep for "MongoDB" keyword in the head of the HTTP session.
In case of "compressed" HTML, a big page is on one line. On a IT page, we could encounter the "MongoDB" keyword and miss-identify the application protocol.

Fixed by matching on a longuer string taken from a live MogoDB server.

Fix https://github.com/drwetter/testssl.sh/issues/2468


Tazmaniac:
Fix bug https://github.com/drwetter/testssl.sh/issues/2468

drwetter:
Thanks a lot!

PR #2472, opened 2024-02-19
Fix subshell killing to avoid zombies

Tazmaniac:
Learned from the rest of the code ...

Bug introduced in https://github.com/drwetter/testssl.sh/pull/2459

Now survive many our of automated target testing without zombies.


drwetter:
ok, thanks

drwetter:
please mind the PR template next time

Tazmaniac:
Ok. Will be more careful. Thank you !

PR #2622, opened 2025-01-17
Support draft-kwiatkowski-tls-ecdhe-mlkem

dcooper16:
## Describe your changes

This commit adds support for the three code points in draft-kwiatkowski-tls-ecdhe-mlkem.

- [ ] Typo fix
- [ ] Documentation update
- [ ] Update of other files


drwetter:
FYI: I reran the Ci jobs after stopping the previous one which ran after 4 hours

testssl.sh
```diff
@@ -14608,6 +14651,10 @@ parse_tls_serverhello() {
                                     258) dh_bits=4096 ; named_curve_str="ffdhe4096" ;;
                                     259) dh_bits=6144 ; named_curve_str="ffdhe6144" ;;
                                     260) dh_bits=8192 ; named_curve_str="ffdhe8192" ;;
+                                    4587) dh_bits=192 ; named_curve_str="SecP256r1MLKEM768" ;;
+                                    4588) dh_bits=192 ; named_curve_str="X25519MLKEM768" ;;
+                                    4589) dh_bits=256 ; named_curve_str="SecP384r1MLKEM1024" ;;
+                                    25497) dh_bits=128 ; named_curve_str="X25519Kyber768Draft00" ;;
```
ghen2, line 14657: Why do you mark this one as 128 bits, but the MLKEM768 equivalent as 192 bits?  It's virtually the same algorithm.

drwetter: I haven't understand where in general the bit info were taken from?

drwetter:
whohoo ! 👍

cloudflare.com and google.com miss to show any TLS 1.3 sig algorithms with this patch . For Geert's site it's fine.

image

drwetter:
<del>Correction: that seems to be a MacOS thing. On Debian it works like expected. 🧐</del> .Not MacOS, for some reason using the supplied openssl is fine . So ``./testssl.sh --openssl=/usr/bin/openssl -f google.com`` misses the TLS 1.3 sig algs too on Debian (12) with this patch.

testssl.sh
ghen2: According to the NIST PQC security levels, ML-KEM-512, 768 and 1024 are equal in strength to AES-128, 192 and 256, respectively.
So this mapping assigns to each hybrid KEM the strength of the strongest component algorithm - except for Kyber768.

dcooper16:
Thanks for catching the mistake. I was adding the new PQC groups to the ClientHello whenever OpenSSL supported X25519 and X448, which was a mistake. The result was that when searching for supported signature algorithms, a PQC group was being selected, the response could not be decrypted, and the signature algorithm could not be determined.

I may not have caught this because I was getting the same results when using `--openssl=/usr/bin/openssl` with and without this change. The reason was that the current testssl.sh was not finding any TLS 1.3 signature algorithms when using `--openssl=/usr/bin/openssl`. I found the reason for that and submitted #2625 to fix that.

testssl.sh
drwetter: https://pq-crystals.org/kyber/

drwetter:
Thanks a lot, David! 🚀

PR #2656, opened 2025-02-13
Enhance ticketbleed testing

dcooper16:
## Describe your changes

Some versions of OpenSSL/LibreSSL do not support TLS 1.1 and earlier, either because they do not support the protocol (e.g, `$OEPNSSL s_client -tls1` results in a "unknown option" error) or because the cryptography needed to support these protocol versions (e.g., MD5/SHA1) is not available.

Given the limitations of some versions of $OPENSSL, this PR enhances ticketbleed testing in two ways. First, it performs the testing using the newest (non-TLS 1.3) version supported by the server, so that TLS 1 and TLS 1.1 aren't used unless TLS 1.2 is not supported. Second, it adds tests for whether the protocol version to be used is supported by $OPENSSL and for whether connection attempts were successful, rather than assuming connection attempts succeed.

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Typo fix
- [ ] Documentation update
- [ ] Update of other files

- [X] For the main program: My edits contain no tabs and the indentation is five spaces
- [X] I've read CONTRIBUTING.md and Coding_Convention.md 
- [X] I have tested this __fix__ against >=2 hosts and I couldn't spot a problem

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -16907,25 +16907,11 @@ run_ccs_injection(){
      return $ret
 }
 
-sub_session_ticket_tls() {
-     local tls_proto="$1"
-     local sessticket_tls=""
-     #FIXME: we likely have done this already before (either @ run_server_defaults() or at least the output
-     #       from a previous handshake) --> would save 1x connect. We have TLS_TICKET but not yet the ticket itself #FIXME
-     #ATTENTION: we DO NOT use SNI here as we assume ticketbleed is a vulnerability of the TLS stack. If we'd do SNI here, we'd also need
-     #           it in the ClientHello of run_ticketbleed() otherwise the ticket will be different and the whole thing won't work!
-     #
-     sessticket_tls="$($OPENSSL s_client $(s_client_options "$BUGS $tls_proto $PROXY $SNI -connect $NODEIP:$PORT") </dev/null 2>$ERRFILE | awk '/TLS session ticket:/,/^$/' | awk '!/TLS session ticket/')"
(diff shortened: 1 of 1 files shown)


testssl.sh
```diff
@@ -16961,27 +16947,52 @@ run_ticketbleed() {
...
+          fileout "$jsonID" "WARN" "Cannot test for ticketbleed. $OPENSSL doesn't support \"s_client $tls_proto\"."
+          return 1
+     fi
      debugme echo "using protocol $tls_hexcode"
 
-     session_tckt_tls="$(sub_session_ticket_tls "$tls_proto")"
+     #FIXME: we likely have done this already before (either @ run_server_defaults() or at least the output
+     #       from a previous handshake) --> would save 1x connect. We have TLS_TICKET but not yet the ticket itself #FIXME
+     #ATTENTION: we DO NOT use SNI here as we assume ticketbleed is a vulnerability of the TLS stack. If we'd do SNI here, we'd also need
+     #           it in the ClientHello of run_ticketbleed() otherwise the ticket will be different and the whole thing won't work!
```
dcooper16, line 16984: I didn't change this part of the code or the comment. However, the comment is out of date since #2465. Was the comment just a  guess that the "whole thing won't work" or were there some servers for which it did not work if SNI was included in one ClientHello but not the other? At the moment, the SNI is included here but not in `$client_hello`.

drwetter: Short notice upfront: Nope, I am not aware that "were there some servers for which it did not work if SNI was included in one ClientHello". I was always w/o SNI until #2465

drwetter:
Thanks a lot @dcooper16 !

PR #2682, opened 2025-03-06
  Speed up startup checks for supported curves and more

drwetter:
## Describe your changes

In order to avoid delays due to lookups of  `-connect invalid.` we  just avoid using it whenever possible. :-)

Except when a WSL system is discovered, we just test before whether `$OPENSSL s_client 2>&1 </dev/null` does a connect. If that succeeds we omit the part `-connect invalid.` to check whether the curve is supported. In some quick testing this in fact improved the startup time up to 10s.

This seemed to work under Linux with several openssl and one LibreSSL binary. More testing would be required, especially e.g. under WSL / WSL2.

Also in `sclient_supported()` the `$OPENSSL s_client` statement was changed in a similar fashion. That worked so far but would need to be observed more closely.

- [x] Improvement (non-breaking change)
- [ ] Typo fix
- [ ] Documentation update
- [ ] Update of other files

- [x] I have tested this __fix__ or _improvement__ against >=2 hosts and I couldn't spot a problem

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -184,7 +184,8 @@ FNAME_PREFIX=${FNAME_PREFIX:-""}        # output filename prefix, see --outprefi
 APPEND=${APPEND:-false}                 # append to csv/json/html/log file
 OVERWRITE=${OVERWRITE:-false}           # overwriting csv/json/html/log file
 [[ -z "$NODNS" ]] && declare NODNS      # If unset it does all DNS lookups per default. "min" only for hosts or "none" at all
-NXCONNECT=${NXCONNECT:-invalid.}        # For WSL this helps avoiding DNS requests to "invalid." which windows seem to handle delayed
+NXDNS=${NXDNS:-invalid.}                # For WSL this helps avoiding DNS requests to "invalid." which windows seem to handle delayed
+NXCONNECT=""                            # needed when when need to test capabilities of the openssl binary
 HAS_IPv6=${HAS_IPv6:-false}             # if you have OpenSSL with IPv6 support AND IPv6 networking set it to yes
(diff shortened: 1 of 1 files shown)


dcooper16:
Hi Dirk,

In https://mailarchive.ietf.org/arch/msg/tls/5Cy9xDDMgQAlVRwuTJeFbhe_5ZM/, I was noted that the command `openssl list -tls1_3 -tls-groups` will output a list of supported groups. As far as I can tell, this is only supported in OpenSSL 3.5.0-dev at the moment. However, it seems that using this command to determine the list of supported groups would be faster than checking the groups one at a time. I created a commit that makes use of this command (https://github.com/testssl/testssl.sh/commit/64075034cb5f045a8a7ec456493a47932d7590e1).

If you think this would be useful, you can add it to your PR.

In the commit that I created, I check both the OpenSSL version number and that the `$OPENSSL list -tls-groups` command works. It should be fine to just check one or the other. I included the version check, since I think use of OpenSSL 3.5 will be limited at the moment, and testing `$OPENSSL list -tls-groups` is relatively expensive. The check for whether `$OPENSSL list -tls-groups` works is just to be safe, in case it is possible to compile OpenSSL 3.5.0 without support for this command.

drwetter:
Hi David,

good point. Looking forward that sounds like a good idea!

The list command is something I feel we didn't pay enough attention to yet.

drwetter:
works under a recent version of WSL. @dcooper16 : if  you want to add your changes, go ahead but we can also work on that later. As you prefer

dcooper16:
> @dcooper16 : if you want to add your changes, go ahead but we can also work on that later. As you prefer

Go ahead and merge your changes, if they are ready. I just made some changes to my commit to make it more efficient. However, my commit only helps for OpenSSL 3.5.0, and that won't be released until next month anyway.

Issue #1105, opened 2018-08-16
IPv6 proxy support


drwetter:
Where IPv6 works for testing servers with testssl.sh and when a special OpenSSL [1] is used a missing feature is still IPv6 proxy support. It seems LibreSSL is supporting this (see #11). For testssl.sh this is transparent and should already work, but for the few cases we still use OpenSSL as a vehicle it does not.

So it would be great to port this to OpenSSL.

[1] either from Fedora,RHEL, Gentoo or from this repo

drwetter:
see #1093

drwetter:
Some problems here:
* <del>For the binaries which can deal with an IPv6 proxy: when sending an openssl s_client connect the square bracket's are missing. </del>
* <del>Seems like when specifying an IPv6 proxy via DNS and an IPv4 target the CONNECT method uses IPv4 only (on a dual stack client) -- looks like the bash sockets are to blame</del>
* <del>There's also a parser problem regarding the proxy port</del>
* <del>docu is missing, at least how to specify the ip address</del>
* we probably want to test whether the binary has IPv6 proxy support.
* <del>Looks like LibreSSL doesn't like the square brackets but openssl (v3 at least) needs it.</del>

drwetter:
> we probably want to test whether the binary has IPv6 proxy support

 we'll use the version banner. Testing the proxy for IPv6 support seems either difficult or errorprone

PR #2721, opened 2025-03-25
Fix some ipv6proxy issues

drwetter:
## Describe your changes

As a quick hack this PR enables *basically* the IPv6 proxy which results that testssl.sh  will use an IPv6 proxy when

 * the binary supports that
 * <del>the binary is used and not tls_sockets()<del>
 * <del>there's no A record but an AAAA record of the proxy or an IPv6 address as proxy address was specified.</del>

The latter should guarantee that it doesn't break anything.

Command line should be ./testssl.sh  --proxy='IPV6PROXY:PORT' TARGET whereas IPV6PROXY should be an IPv6 address . A IPv6 only hostname should work too but wasn't tested. 

See #1105

Done:

* Distinguished between LibreSSL and OpenSSL IPv6 proxy (for the binaries which can deal with an IPv6 proxy: when sending an openssl s_client connect the square bracket's were missing when using LibreSSL)
* Parser problem for IPv6 and  proxy port fixed
* Seemed like when specifying an IPv6 proxy via DNS and an IPv4 target the CONNECT method used IPv4 only (on a dual stack client) -- looks like the bash sockets are to blame
* docu added

- [ ] Bug fix
- [ ] Improvement
- [ ] Typo fix
- [x] Documentation update
- [ ] Update of other files


drwetter:
@dcooper16 : Do you know by any chance since when (in OpenSSL and LibreSSL) IPv6 is supported for the proxy? I'd like to add a check here. Unfortunately that probably needs to be version-based  rather as testing-based as we normally do.

dcooper16:
> @dcooper16 : Do you know by any chance since when (in OpenSSL and LibreSSL) IPv6 is supported for the proxy? I'd like to add a check here. Unfortunately that probably needs to be version-based rather as testing-based as we normally do.

Sorry, no. I haven't really worked with IPv6 or proxies.

drwetter:
Thanks anyway. PS: The git log (of openssl) wasn't really helpful. As I didn't find even less in the git log of LibreSSL I guess it was implemented before the fork. **Edit**: 1.1.1 works, and 1.1.0m as well

PR #2744, opened 2025-04-22
Remove redundant commands in Dockerfile

### LINKED ISSUE #2439: Review the recent change of the Dockerfile as suggested
**Which version are you referring to** 3.2? (please check also how old your version is compare to the ones here)

by @polarathene, see https://github.com/drwetter/testssl.sh/issues/2420#issuecomment-1762749767

drwetter:
.. see https://github.com/testssl/testssl.sh/issues/2420#issuecomment-1762749767

As suggested by @polarathene the not needed repos are more elegantly removed, commands for removing util-linux removal and zypper up were redudant and thus squashed.

First stage was build manually and it looked fine.

This fixes #2439 .

- [ ] Bug fix
- [ ] Improvement
- [ ] Typo fix
- [ ] Documentation update
- [x] Update of other files

- [ ] For the main program: My edits contain no tabs, indentation is five spaces and any line endings do not contain any blank chars
- [ ] I've read CONTRIBUTING.md and Coding_Convention.md 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,16 +10,13 @@ ARG INSTALL_ROOT
 
 # /etc/os-release provides $VERSION_ID below.
 # We don't need the openh264.repo and the non-oss repos, just costs build time (repo caches).
-# Also we need to remove the util_linux RPM to /really/ make sure busybox-util-linux gets installed.
-# And we need to run zypper update, see all PR #2424.
+
 RUN source /etc/os-release \
-  && rm -f /etc/zypp/repos.d/repo-openh264.repo /etc/zypp/repos.d/repo-non-oss.repo \
+  && zypper removerepo repo-openh264 repo-non-oss repo-update-non-oss \
   && export ZYPPER_OPTIONS=( --releasever "${VERSION_ID}" --installroot "${INSTALL_ROOT}" --cache-dir "${CACHE_ZYPPER}" ) \
   && zypper "${ZYPPER_OPTIONS[@]}" --gpg-auto-import-keys refresh \
-  && rpm -e util-linux --nodeps \
   && zypper "${ZYPPER_OPTIONS[@]}" --non-interactive install --download-in-advance --no-recommends \
        bash procps grep gawk sed coreutils busybox ldns libidn2-0 socat openssl curl \
-  && zypper up -y \
   && zypper "${ZYPPER_OPTIONS[@]}" clean --all
 ## Cleanup (reclaim approx 13 MiB):
 # None of this content should be relevant to the container:


drwetter:
@polarathene : better?

polarathene:
Yeah looks like the advice I gave in 2023 👍 

I personally have been sticking to Fedora as a base image of choice but I don't expect much to have changed with the Leap advice since then :)

## Reference - Comparison to `chisel` for the `builder` stage

**FWIW:** Switching from openSUSE Leap to Canonical's [`chisel`](https://github.com/canonical/chisel) ~~is over 20MB less in size for the `builder` stage (43MB vs 65MB)~~:

**EDIT:** My mistake, I was looking at the openSUSE Leap `builder` stage first `RUN` layer for `/root-fs` that included other changes outside of that location, and the other two `RUN` don't help reduce the weight of that stage until the final stage `COPY` 😓 (_where `/root-fs` becomes 46MB_). Thus nevermind this feedback below, it only saves a few MB.

```Dockerfile
# This is only necessary until an eventual official chisel image is published:
FROM alpine AS chisel
ARG CHISEL_VERSION=1.1.0
ARG TARGETARCH
ARG CHISEL_RELEASE="https://github.com/canonical/chisel/releases/download/v${CHISEL_VERSION}/chisel_v${CHISEL_VERSION}_linux_${TARGETARCH}.tar.gz"
RUN wget -qO - "${CHISEL_RELEASE}" | tar -xz --no-same-owner -C /usr/local/bin chisel

# Instead of COPY from `/root-fs`, you can instead use `FROM builder`
# as Chisel installed packages directly to `/` of a scratch base.
FROM scratch AS builder
ARG XDG_CACHE_HOME=/cache
SHELL ["/nu", "-c"]
# These bind mounts are only needed temporarily to run `chisel`
RUN --mount=type=cache,target=${XDG_CACHE_HOME}/chisel,id="chisel-cache" \
    --mount=type=bind,from=chisel,source=/etc/ssl/certs/ca-certificates.crt,target=/etc/ssl/certs/ca-certificates.crt \
    --mount=type=bind,from=chisel,source=/usr/local/bin/chisel,target=/chisel \
    --mount=type=bind,from=ghcr.io/nushell/nushell,source=/usr/bin/nu,target=/nu \
    /chisel cut --release ubuntu-24.04 --root / \
      base-files_base bash_bins procps_bins grep_bins gawk_bins sed_bins coreutils_bins busybox_bins openssl_bins curl_bins
# If have another `RUN` that should execute via `/bin/sh` or `/bin/bash`, change `SHELL`:
SHELL ["/bin/bash", "-c"]
```

**NOTES:**
- `/usr/share/doc` at the install root location will be under 1MB, but otherwise none of the other stuff needs to be cleaned up like with Leap `builder` stage.
- If `/etc/ssl/certs/ca-certificates.crt` is required, then add this slice too: [`ca-certificates_data`](https://github.com/canonical/chisel-releases/blob/e5db53e97e03c46944a84e7ab5abeb1b7cb53cd9/slices/ca-certificates.yaml#L7-L12)
  - **EDIT:** Actually this file doesn't seem to be present in the existing Leap image so should be a non-issue 😅
- `socat` is missing a slice presently.
  - A slice for this package was almost contributed indirectly [here](https://github.com/canonical/chisel-releases/pull/258/commits/1040ce8a986ae6d0e9f44e14b56eff70f4b6e480#diff-e5d3f94a7692fa1a4bbbdcc71ad9f60a2c51c4971e60e3bc17c3b37bf5a270a4), that would need to be upstreamed for a `socat_bins` slice to have parity with the OpenSUSE Leap base image.
  - `socat` is required for [testing STARTTLS injection vulnerabilities](https://github.com/testssl/testssl.sh/blob/a9a1295875dc3835cb8fd0a226e4e9395968e915/doc/testssl.1.md?plain=1#L241) (_mail service ports only_).
  - The `socat` package on Leap is only a 1MB weight difference.

**Worth switching?:**
- It might be worth keeping an eye on `chisel`, once it gets some UX improvements (_and that `socat` slice_) you could have a simpler, smaller image build (_**EDIT:** Actual image size reduction wasn't as big as I initially thought_).
- Compared to Fedora / openSUSE distros that offer more complex package managers (_and a far broader selection of packages_), `chisel` may not be as convenient / flexible. `chisel` is more of an optimization tool oriented towards minimal image builds akin to "distroless" images.

drwetter:
Thanks! And apologize to have this hung for so long.

Actually heard the first time about chisel. Thanks for the heads up. IIRC by the end of the year opensuse 15.6 should be EOL and after that it looks like there's a version 16 from which I don´t know what it'll be like. From what I've heard earlier the desktop version will be containerized -- whatever that means for this project.

polarathene:
> Thanks! And apologize to have this hung for so long.

No worries! I know how it is finding the time to spare for such 😅 

> From what I've heard earlier the desktop version will be containerized -- whatever that means for this project.

SUSE has other base images specifically for containers (BCI - Base Container Image) at [their own image registry](https://registry.suse.com/repositories?sort=featured&categories%5B%5D=bci).

It's been a while but I think the BCI images might differ a little with packages. So if there were an issue with your current base image in the future, the BCI images should be similar enough for you and I doubt they'd be dropping those.

If needed you could also switch to an alternative base image:
- Fedora (_more frequent release cadence + minimal lag in package versions, approx 6 months_)
- Debian (_packages tend to lag behind more than openSUSE Leap, usually releases a major version every 2 years?_)
- Ubuntu + `chisel` (_optimized for containers via `chisel`, very competitive in size, limited package selection, follows upstream Ubuntu releases every 6 months for package updates_)
- Alpine (_musl instead of glibc, frequent releases + minimal lag in package versions, previously encountered some compatibility and performance issues_)

Fedora was simple enough, with a similar approach to what openSUSE Leap is doing with `--installroot`. It is my preferred choice since I investigated in the original discussions / PR that contributed Leap to this project.

`chisel` seems great for minimal size, just needs more time to be friendlier to maintain / use.

drwetter:
Short comment here..

> Debian (packages tend to lag behind [..]

It's seems to me not relevant

polarathene:
I think we also considered Debian alongside Fedora and Leap. It wasn't as competitive at meeting the size goal though that was important for you.

A recent issue I had with Debian was Debian 12 was released around June 2023, and that came with SpamAssassin 4.0 package, which in March 2024 released 4.0.1, yet that did not arrive until Debian 12 in March 2025. Dates might be slightly off, but that was quite a long time for a bug fix to land.

Other software also similarly lagged behind in receiving updates with point releases to get fixes we wanted. We still use Debian as a base for now but that is one of my annoyances with it. You shouldn't have to wait 1-2 years for that, it's even longer for some packages when it comes to upgrading major/minor versions (_Leap had a similar observation IIRC, but was a bit more inconsistent_).

PR #2758, opened 2025-05-06
ci: `Dockerfile` - Support local and git builds with the openSUSE Leap image

polarathene:
## Describe your changes

This is roughly the equivalent of the [recent Alpine PR](https://github.com/testssl/testssl.sh/pull/2752) to bring the two to parity. Although this one has the more expected default of `dist-local` to be used with a `git clone`.

I've additionally:
- Shifted comments to be more inline with the `RUN` lines. These are omitted when parsed by `RUN`, unlike as shell script when `\` is involved. This is because all those lines for `RUN` will be collapsed to a single line when parsed and give to the shell to run.
- The symlink likewise is more explicit, matching the Alpine image.
- As that image also had `xxd` and BusyBox provides that, I've also add a symlink for that (_assuming the busybox xxd is compatible_).
- The cleanup step is no longer separate `RUN`, just a single `rm -r` command.

- [x] Improvement
- [x] Update of other files

### DIFF
--- a/Dockerfile.md
+++ b/Dockerfile.md
@@ -1,47 +1,66 @@
 ## Usage
 
-### From git directory
+Run the image with `testssl.sh` options appended (default is `--help`). The container entrypoint is already set to `testsl.sh` as the command for convenience.
 
-```
-docker build .
-```
-
-Catch is when you run without image tags you need to catch the ID when building
-
-```
-[..]
----> 889fa2f99933
-Successfully built 889fa2f99933
+```bash
+docker run --rm -it ghcr.io/testssl/testssl.sh:3.2 --fs github.com
 ```
 
-More comfortable is
+### Output files
 
-```
-docker build -t mytestssl .
-docker run --rm -t mytestssl example.com
-```
+Keep in mind that any output file (_`--log`, `--html`, `--json`, etc._) will be created within the container.
 
-You can also supply command line options like:
+Use a volume bind mount to a local host directory to access the files outside of the container. Set a working directory for the container and any options output prefix can then use a relative path, like this example for `--htmfile`:
 
-```
-docker run -t mytestssl --help
-docker run --rm -t mytestssl -p --header example.com
+```bash
+# Writes the HTML output to the host path: /tmp/example.com_p443-<date>-<time>.html
+docker run --rm -it -v /tmp:/data --workdir /data ghcr.io/testssl/testssl.sh:3.2 --htmlfile ./ example.com
 ```
 
-### From dockerhub or GHCR
+**NOTE:**
+- The UID/GID ownership of the file will be created by the container user `testssl` (`1000:1000`), with permissions `644`.
+- Your host directory must permit the `testssl` container user or group to write to that host volume. You could alternatively use [`docker cp`](https://docs.docker.com/reference/cli/docker/container/cp/).
 
-You can pull the image from dockerhub or ghcr.io and run:
+### From DockerHub or GHCR
 
-``docker run --rm -t drwetter/testssl.sh --fs example.com`` or ``docker run --rm -t ghcr.io/testssl/testssl.sh --fs example.com``
+You can pull the image from either of these registries:
+- DockerHub: [`drwetter/testssl.sh`](https://hub.docker.com/r/drwetter/testssl.sh)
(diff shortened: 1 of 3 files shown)


polarathene:
**UPDATE:**
- I've added the build advice below (simplified) to your `Dockerfile` docs
- `Dockerfile` docs revised/reorganized.
- Minor changes to comments in the Alpine image.

---

## Sizes

46MB base + 8MB from the `COPY` (_16MB with the `ADD`_). Thus 54MB vs 62MB final image sizes.

Sizes were inspected via:

```bash
docker run --rm -it -v /var/run/docker.sock:/var/run/docker.sock ghcr.io/wagoodman/dive:latest localhost/testssl.sh:3.2
```

Alpine variants are 27MB vs 35MB.

## Building

```bash
# `dist-local`
mkdir /tmp/testssl && cd /tmp/testssl
docker build --tag localhost/testssl.sh:3.2 --target dist-local .

# `dist-git` (still requires a local `Dockerfile`):
docker build --tag localhost/testssl.sh:3.2 --target dist-git --build-arg GIT_BRANCH=3.2 .
```

### Remote builds (`Dockerfile` + build context)

It is also possible to build without a `Dockerfile` locally (_but only really useful for the `dist-git` target_):

```bash
# Use a URL to a remote `Dockerfile` as the build context (or set it via `--file`):
docker build \
  --tag localhost/testssl.sh:3.2 --target dist-git --build-arg GIT_BRANCH=3.2 \
  https://raw.githubusercontent.com/testssl/testssl.sh/3.2/Dockerfile
```

For `dist-local` to support `COPY`, this can also be done with a remote build context instead of a local git clone (_although this is what happens behind the scenes via `docker build` instead_), but the image will be as large as `dist-git` since [`.dockerignore` is not supported with remote build contexts](https://github.com/docker/buildx/issues/3169) (_even if you have this file locally_).

```bash
# COPY will source build context from the remote repo + branch instead:
# - `.dockerignore` is not supported, the image will be as large as `dist-git`
# - `--file` could be omitted here as the default is to use a `Dockerfile` at the root of repo.
docker build \
  --target dist-local \
  --file https://raw.githubusercontent.com/testssl/testssl.sh/3.2/Dockerfile \
```

**NOTE:** Using `--file` is required if building an alternative `Dockerfile` instead (_Alpine image variant_):

```bash
# Demonstrating with the alternative Alpine image variant:
# Both targets will build the same due to remote build context
docker build \
  --target dist-local \

# For `dist-git` the build context isn't relevant, nothing is copied from it.
docker build \
  --target dist-git \
  --build-arg GIT_BRANCH=3.2
  .
```

**NOTE:** This kind of makes the `dist-git` target redundant?
- You're limited by the branch, whereas if you want the commit you'd need to manage the `git clone` yourself and perform a checkout, at which point you might as well just use `dist-local`.

Dockerfile.md
```diff
@@ -1,47 +1,66 @@
...
+This will produce a slightly larger image however as `.dockerignore` is not supported with remote build contexts.
+
+If you would like to build the Alpine image instead this way, just provide the alternative `Dockerfile` via `--file`:
+
+```bash
+docker build \
+  --tag localhost/testssl.sh:3.2-alpine \
+  --file https://raw.githubusercontent.com/testssl/testssl.sh/3.2/Dockerfile-alpine \
+  https://github.com/testssl/testssl.sh.git#3.2
+```
```
polarathene, line 66: Now that `Dockerfile.git` has become `Dockerfile-alpine`, perhaps the default `dist-git` should also be changed to `dist-local`?

This command presumably makes `dist-git` redundant, but I understand your reluctance to drop that from the 3.2 series and defer until 3.3.

As the file name has changed in the 3.2 series though, perhaps changing the default to `dist-local` is valid for consistency, and that would align better with these docs which don't mention using `--target` at all.

@jauderho may have an opinion on this as he contributed the `Dockerfile.git` file IIRC? Is this build command equivalent in functionality?

drwetter: As said, if you believe Dockerfile.alpine fits better, then feel free to change this. Maybe amend this PR accordingly or let me know if I should rather merge this first.

drwetter:
Thanks, @polarathene, Looks good!

See comment whether I should merge this first.

drwetter:
Ok, then we can take care of the renaming afterwards. Thanks for the PR

PR #2768, opened 2025-05-11
ci(fix): 3.2 - Docker support adjustments

polarathene:
## Describe your changes

This PR does some minor revisions to `Dockerfile.md` docs + the `docker-3.2.yml` workflow as detailed in their commits.

- The docs are now using [GH Alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) syntax for displaying notes.
- In addition to the `Dockerfile-alpine` to `Dockerfile.alpine` rename, it also corrects the CI workflow ([which is presently failing](https://github.com/testssl/testssl.sh/actions/workflows/docker-3.2.yml)) to use the updated `GIT_BRANCH` build arg instead of the previous `BUILD_VERSION` build arg (_introduced recently via_ https://github.com/testssl/testssl.sh/pull/2752).

---

**NOTE:**
- We could also consider removing the need for the build arg in CI by preferring the `dist-local` stage. This would be implicit if we swap the `dist-git` + `dist-local` stage order of appearance in the `Dockerfile` to match the openSUSE `Dockerfile` since the last stage is the default one to build.
- [`dist-git` is effectively redundant](https://github.com/testssl/testssl.sh/pull/2758#issuecomment-2853244152) and could be dropped in favor of the alternative shown in the `Dockerfile.md` docs (_which makes no mention of these two final stages, as there isn't a need to raise awareness_).

  That said the UX may not be as pleasant, so you could keep the `dist-git` stage for a niche use-case if you think it's worthwhile? 🤷‍♂️ 

  Proposed (without `dist-git`):

  ```bash
  docker build --tag localhost/testssl.sh:3.2 https://github.com/testssl/testssl.sh.git#3.2

  # Alpine variant requires `--file`:
  docker build \
  ```

  Proposed (_if keeping `dist-git`, but as an opt-in target instead_) with a local `Dockerfile` copy, but remote source:
  
  ```bash
  # NOTE: Only cloning for the `Dockerfile`
  
  # Optional - Set the `GIT_URL` build arg too if preferring a fork instead of upstream repo:
  docker build \
    --target dist-git \
    --build-arg GIT_BRANCH=3.2 \
    .
  ```

  The main benefit of the latter with a `dist-git` target is that the user doesn't have to think about the two different URL inputs (_shouldn't be much of a concern, they'd likely copy/paste from docs and it's either a one off or automated script, assuming you actually have any users that care about this support in the first place_).

  Either way, you could have automated CI builds at a single branch that changes `GIT_BRANCH` to official releases or some PR branch/fork, that's the main benefit, but as documented this would work fine with `dist-local` with a remote context URL to a repo + branch.

  If they really care about size they should prefer `dist-local` with manual `git clone` so that `.dockerignore` can be used to shave off another 8MB.

- [x] Bug fix
- [x] Improvement
- [x] Documentation update


drwetter:
Hi @polarathene ,

looks good, I'd like to merge it. One question why did you change this?

```
-        uses: docker/setup-qemu-action@v3.6.0
+       uses: docker/setup-qemu-action@v3
```

Which minor version 3 would be picked then? Normally the dependabot came and suggested to upgrade the minor version like this one: https://github.com/testssl/testssl.sh/pull/2702

polarathene:
> Which minor version 3 would be picked then?

It resolves to the latest in the 3.x series at the time the workflow is run.

You can pin by minor version if you prefer dependabot to bump those, I just noticed some of your actions were only pinning by major version so I adjusted for consistency with less maintenance hassle (_unless something goes wrong with a minor version update, then pinning with explicitly approved version bumps is more useful_).

I can revert that if you'd prefer?

drwetter:
> I can revert that if you'd prefer?

Nah, that's fine. Just didn't know for sure and I needed your word on this.

For the dist-local issue it's better to have an issue for this. ;-) Otherwise I forget it. It's best practice anyway not to mix points.

drwetter:
Houston, we have a problem:  https://github.com/testssl/testssl.sh/actions/runs/14969641933

image

polarathene:
Oh sorry about that, [their docs](https://github.com/docker/build-push-action?tab=readme-ov-file#inputs) say the type is a list which implied an array 🤷‍♂️ 

> image

Easy fix to just turn that back into a string then. In a project I maintain it turns out it's a multi-line string where I have values set like:

```yaml
build-args: |
  GIT_BRANCH=3.2
```

I can fix that if you like 👍

PR #2771, opened 2025-05-12
ci: Docker - Prefer `dist-local` stage

### LINKED ISSUE #2769: Preferring the `dist-local` stage ?
Taken from #2768 

----

**NOTE:**
- We could also consider removing the need for the build arg in CI by preferring the `dist-local` stage. This would be implicit if we swap the `dist-git` + `dist-local` stage order of appearance in the `Dockerfile` to match the openSUSE `Dockerfile` since the last stage is the default one to build.
- [`dist-git` is effectively redundant](https://github.com/testssl/testssl.sh/pull/2758#issuecomment-2853244152) and could be dropped in favor of the alternative shown in the `Dockerfile.md` docs (_which makes no mention of these two final stages, as there isn't a need to raise awareness_).

  That said the UX may not be as pleasant, so you could keep the `dist-git` stage for a niche use-case if you think it's worthwhile? 🤷‍♂️ 

  Proposed (without `dist-git`):

  ```bash
  docker build --tag localhost/testssl.sh:3.2 https://github.com/testssl/testssl.sh.git#3.2

  # Alpine variant requires `--file`:
  docker build \
  ```

  Proposed (_if keeping `dist-git`, but as an opt-in target instead_) with a local `Dockerfile` copy, but remote source:
  
  ```bash
  # NOTE: Only cloning for the `Dockerfile`
  
  # Optional - Set the `GIT_URL` build arg too if preferring a fork instead of upstream repo:
  docker build \
    --target dist-git \
    --build-arg GIT_BRANCH=3.2 \
    .
  ```

  The main benefit of the latter with a `dist-git` target is that the user doesn't have to think about the two different URL inputs (_shouldn't be much of a concern, they'd likely copy/paste from docs and it's either a one off or automated script, assuming you actually have any users that care about this support in the first place_).

  Either way, you could have automated CI builds at a single branch that changes `GIT_BRANCH` to official releases or some PR branch/fork, that's the main benefit, but as documented this would work fine with `dist-local` with a remote context URL to a repo + branch.

  If they really care about size they should prefer `dist-local` with manual `git clone` so that `.dockerignore` can be used to shave off another 8MB.

----

polarathene:
## Describe your changes

Changes the Alpine `Dockerfile` to also default to `dist-local` stage as the final image output, like the openSUSE Leap `Dockerfile` already does. The `dist-git` stage remains for now as opt-in, but I'm not sure if it has any notable advantage to any potential users, perhaps drop it in the 3.3 release series?

This will affect anyone relying on the `dist-git` default target (_which recently had a breaking change of `BUILD_VERSION` => `GIT_BRANCH` build arg_), they would now need to also be explicit about the `--target` stage when building. However just like mentioned in #2769 there is an alternative without `--target` that works effectively the same with `dist-local`:

```bash
docker build --tag localhost/testssl.sh:3.2 https://github.com/testssl/testssl.sh.git#3.2

# Alpine variant requires `--file`:
docker build \
```

Resolves: https://github.com/testssl/testssl.sh/issues/2769
Closes: https://github.com/testssl/testssl.sh/pull/2770

- [x] Improvement
- [x] Update of other files

### DIFF
--- a/.github/workflows/docker-3.2.yml
+++ b/.github/workflows/docker-3.2.yml
@@ -56,8 +56,6 @@ jobs:
           context: .
           file: Dockerfile.alpine
           platforms: linux/amd64,linux/386,linux/arm64,linux/arm/v7,linux/arm/v6,linux/ppc64le
-          build-args:
-            - GIT_BRANCH
           cache-from: type=gha, scope=${{ github.workflow }}
           cache-to: type=gha, scope=${{ github.workflow }}
           labels: ${{ steps.docker_meta.outputs.labels }}
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -12,12 +12,12 @@ CMD ["--help"]
 # Final image stage (add `testssl.sh` project files)
 # Choose either one as the final stage (defaults to last stage, `dist-git`)
 
-# 27MB Image (Local repo copy from build context, uses `.dockerignore`):
-FROM base-alpine AS dist-local
-COPY --chown=testssl:testssl . /home/testssl/
-
 # 35MB Image (Remote repo clone, cannot filter content through `.dockerignore`):
 FROM base-alpine AS dist-git
 ARG GIT_URL=https://github.com/testssl/testssl.sh.git
 ARG GIT_BRANCH
 ADD --chown=testssl:testssl ${GIT_URL}#${GIT_BRANCH?branch-required} /home/testssl
+
+# 27MB Image (Local repo copy from build context, uses `.dockerignore`):
+FROM base-alpine AS dist-local
+COPY --chown=testssl:testssl . /home/testssl/


polarathene:
@drwetter you may prefer this PR :)

drwetter:
Hi @polarathene :

I lost track a bit  whether it was me editing the file or the PR... long story short: Is change I needed to do fine?

polarathene:
> whether it was me editing the file or the PR...

I'm not sure what you are referring to. This PR removes the need for the CI build arg and makes the Alpine Dockerfile more consistent with the openSUSE Leap Dockerfile, they would both be `dist-local` stage by default which is what you'd typically expect.

In your 3.3 series you might want to remove `dist-git` as it's unlikely to be of value to anyone, your Dockerfile docs show how to do the equivalent without `dist-git`.

drwetter:
@polarathene: I had to edit the PR as there was a merge conflict. Could you just have a look that it's correct?

polarathene:
**TL;DR:** LGTM, feel free to merge.

> Could you just have a look that it's correct?

[This diff](https://github.com/testssl/testssl.sh/commit/0ab5e2d878ac6e0adc1ceeeff9b5580ec7ae6302#diff-80ccf8f4ffedfa99ec95209d7c7c524dcf8c5319606b61f4e71d67e0e8a03455)? It looks fine to me.

The only change to the `docker-3.2.yaml` workflow is removal of:

```yaml
build-args: GIT_BRANCH
```

Which is presumably what the merge conflict was regarding.

---

This PR is simple, it just swaps the position of `dist-git` and `dist-local` stages in the Alpine `Dockerfile`, that makes the `build-args: GIT_BRANCH` workflow setting redundant as `dist-local` stage is built instead of `dist-git` 👍

drwetter:
> Which is presumably what the merge conflict was regarding.

yes, I believe so. Was just confusing to me that this popped up. Nevermind. Thx!

PR #2850, opened 2025-07-18
Ensure HEAD requests on port 80 can actually go through, including timeout checks (only used by the Opossum test at the moment, fixes #2847)

### LINKED ISSUE #2847: [BUG] Using a port suffix for the URI causes some file descriptor errors
I am running version `commit a719c46bcb49caa32fd08e4a1301da48f6b013ae`.

I couldn't find any issues or PRs mentioning this exact problem. When using a URI like `host:port` some errors occur for certain tests, when it tries to set up a file descriptor. For demonstration purposes I'll just use a public domain and the default port of 443, in which case the problem also occurs merely because the port is present. The relevant section from the output:

```
 Testing vulnerabilities 

 Heartbleed (CVE-2014-0160)                not vulnerable (OK), no heartbeat extension
 CCS (CVE-2014-0224)                       not vulnerable (OK)
 Ticketbleed (CVE-2016-9244), experiment.  not vulnerable (OK), no session ticket extension
 Opossum (CVE-2025-49812)                  /usr/local/src/testssl.sh/testssl.sh: line 1940: github.com:443: Name or service not known
/usr/local/src/testssl.sh/testssl.sh: line 1940: /dev/tcp/github.com:443/80: Invalid argument
```

Line 1940 is `exec 33<>/dev/tcp/$node/80`, but `$node` is `github.com:443` so that won't work. :> I think the same problem **might** also occur on line 1849, given that it's exactly the same but other code uses `$nodeip` instead. I'm not sure under what circumstances that might trigger though.

`/usr/local/src/testssl.sh/testssl.sh --parallel github.com:443`

**Expected behavior**

Looking at the CVE for Opossum it seems to require an initial plaintext connection. In that case I'm actually not too sure if this particular test is very useful when running on a non-standard port, since there's a good chance port 80 won't be in use anyway (or it even belongs to a different web server on the same IP). Trying to guess the port is likely to be incorrect, and the specific host I actually want to test doesn't even use any plaintext ports.

Instead of causing a "strange" looking error, perhaps testssl.sh could simply detect the port suffix and emit a warning of its own? Then users could either skip the test, or provide an extra argument for the plaintext port. I'm just running with the default tests though so I'm not sure how to skip **just** the Opossum test, besides manually specifying all the other tests (requiring a lot of maintenance).

**Your system**
 - OS: `Debian GNU/Linux 12 (bookworm)`
 - Platform: `Linux 6.1.0-37-amd64 x86_64`
 - OpenSSL + bash: `Using OpenSSL 3.0.16 (Apr 15 2025)  [~94 ciphers]` and `Using bash 5.2.15`

GottemHams:
## Describe your changes

See #2847 for more details.

- [X] Bug fix
- [ ] Improvement
- [ ] Typo fix
- [ ] Documentation update
- [ ] Update of other files

- [X] For the main program: My edits contain no tabs, indentation is five spaces and any line endings do not contain any blank chars
- [X] I've read CONTRIBUTING.md and Coding_Convention.md 
- [X] I have tested this __fix__ or __improvement__ against >=2 hosts and I couldn't spot a problem

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -1933,11 +1933,33 @@ http_header_printf() {
      local errfile=$TEMPDIR/$NODE.$NODEIP.http_header_printf-err.log
      local -i ret=0
      local proto="" foo="" node="" query=""
+     local pid
 
      [[ $DEBUG -eq 0 ]] && errfile=/dev/null
 
      IFS=/ read -r proto foo node query <<< "$1"
-     exec 33<>/dev/tcp/$node/80
+
+     # This command may fail immediately if the server responds with "connection refused".
+     # To deal with this while also being able to interrupt the call, we can use 'wait' to check the real exit code if the process no longer exists.
+     # The connection is wrapped in a subshell so we can redirect stderr instead of polluting any test output.
+     bash -c "exec 33<>/dev/tcp/$NODEIP/80" 2>$errfile &
+     pid=$!
+     if ! ps $pid >/dev/null && ! wait $pid; then
+          [[ -n "$PROXY" ]] && return 3
+          return 1
+     fi
+
+     # On the other hand, if the process is still running then we'll need to check for a timeout, since the server doesn't seem to actively reject the connection
+     wait_kill $pid $HEADER_MAXSLEEP
+     if [[ $? -ne 0 ]]; then
+          # Killed, i.e. server didn't respond in time
+          [[ -n "$PROXY" ]] && return 3
+          return 1
+     fi
+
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks a lot for your work, @GottemHams  !  I prefer my PR going forward.

Just needs to be tested for cases which doesn't have port 80 opened. And it's a matter of balancing that with when does a block happen there. For the latter and atm I'd tend to not put it into the background. If somebody has good arguments otherwise, I am all ears.

Issue #479, opened 2016-09-25
DANE support


rugk:
See https://en.wikipedia.org/wiki/DNS-based_Authentication_of_Named_Entities

Test website: https://dane.sys4.de/

PR #2866, opened 2025-07-29
DNS HTTPS RR (RFC 9460) for 3.3dev

drwetter:
This is a fresh start for #2484 as the PR wasn't ready yet for 3.2 by the time it was released.

The info for the HTTPS RR shows up in the very beginning, i.e. in `service_detection()`. All keys are listed now in bold, values in a regular font.

`get_https_rrecord()` was introduced by copying and modifying `get_caa_rr_record()`.

There's a similar obstacle as with CAA RRs: older binaries show the  resource records binary encoded. Thus a new set of global vars is introduced HAS_*_HTTPS which check whether the binaries support decoding the RR directly.

For CAA there was a minor bug fixed when records were queried also when it was instructed to minimize/skip or use proxy only.

Todo:
- Add logic in QUIC
    - if RR is detected and not QUIC is possible
    - add time for QUIC detection when RR is retrieved
 - show full HTTPS RR record, at least when having a new DNS client
 - shorten the comments in `get_https_rrecord()`
 - man page
 - when ASSUME_HTTP is set and no services was detected: this needs to be handled

- [ ] Bug fix
- [ ] Improvement
- [ ] Typo fix
- [ ] Documentation update
- [x] Update of other files

- [] For the main program: My edits contain no tabs, indentation is five spaces and any line endings do not contain any blank chars


Delicates:
The reporting should also indicate whether or not the DNS resolver flags that the HTTPS record has been secured and verified by DNSSEC by checking for AD or AA flag in the DNS response.

drwetter:
> The reporting should also indicate whether or not the DNS resolver flags that the HTTPS record has been secured and verified by DNSSEC by checking for AD or AA flag in the DNS response

DNSSEC is from the developer stance a different topic and won't be addressed here.

Delicates:
> DNSSEC is from the developer stance a different topic and won't be addressed here.

No worries.

It just seems relevant and is fairly simple to do with something like: `delv +short +trust <hostname> HTTPS`.

But I guest it would be best to address DNSSEC systemically for all DNS lookups, not just `HTTPS`.

Would be a much more important pre-requisite for DANE (#479) than this.

drwetter:
Actually I looked after a long while into that and delv seemed to me the best choice as it checks whether DNSSEC really works instead of just looking for all releant records. Problem with that is just needs to be installed. But atm I believe that'll become then just a requirement. 

What me confused though is that on a host which deliberately has no DNSSEC capable forwarder (no "ad" returned from e.g. `drill -o AD org.`) delv returns `fully validated` for example.com . 

PS: See also https://github.com/testssl/testssl.sh/pull/1815

Delicates:
> Actually I looked after a long while into that and delv seemed to me the best choice as it checks whether DNSSEC really works instead of just looking for all releant records. Problem with that is just needs to be installed. But atm I believe that'll become then just a requirement.

`delv` is actually a part of BIND. So if the system already has `nslookup` and `dig`, then it should also have `delv` installed as part of the same package. Unless it is a really-really ancient decades-old BIND version, or unless the distribution for some reason breaks down BIND tools into separate packages.

> What me confused though is that on a host which deliberately has no DNSSEC capable forwarder (no "ad" returned from e.g. `drill -o AD org.`) delv returns `fully validated` for example.com .

The purpose of `delv` is to perform validation of all DNSSEC records by itself, regardless of what DNS resolver does. So it ignores presence or absence of AD flag returned by the resolver. If you read the `delv` man-page, it says:
```
Because delv does not rely on an external server to carry out validation, it can be used to check the validity of DNS responses in environments where local name servers may not be trustworthy.
```

`testssl.sh` can basically completely outsource DNSSEC validation to `delv` and not worry about whether the DNS resolver does it or not.

`delv` still relies on the DNS resolver to perform the queries though, which can be problematic if:
* the DNS resolver actively tampers with responses by stripping away all DNSSEC records (like some 4G/5G mobile operators do), it will not be able to validate anything so will report an `unsigned answer`.
* the testing is done inside a split-DNS environment where local private LAN DNS records are returned instead of public Internet-facing records.

A user could ask `delv` to use a specific resolver `@server` that might be more trusted than the local resolver, or one that will always return public Internet-facing records, for example Google DNS: `delv +short +trust @8.8.8.8 www.example.com HTTPS`.

Just need to do a bit of multi-line parsing:
* trust level is indicated in the first line with the leading `;` (even for `NXDOMAIN` status) unless resolution or validation fails
* CNAME records are in the leading lines before queried records (if they exist)
* queried records (could be multiple e.g. `TXT`) are printed last (if they exists)

`delv +nodnssec` might be easier to grep and analyse than `delv +short +trust`.

The possible trust levels could be:
* `; fully validated`
* `; unsigned answer`
* `; negative response, fully validated`
* `; negative response, unsigned answer`
* missing if `delv` failed to contact the resolver, or if it has encountered a broken DNSSEC such as `dnssec-failed.org`

The only frustration is that `delv` on execution seems to always return a successful return code, regardless of any errors/failures.

drwetter:
Thanks for the helpful heads up! See https://github.com/testssl/testssl.sh/issues/3034 

Parsing shouldn't be a big deal. Just a bit of work. Error code: sigh.

Fine tuning: The standard DNS resolver should be used as a default, as some environments block traffic to external resolvers. There should be an option though supplying an external server of choice, for DNSSEC.

PR #2885, opened 2025-09-15
HTTP Clock Skew Invalid Age header crash fix

Tristanhx:
## Describe your changes

HTTP_clock_skew breaks execution of rest of script if Age header contains non-numerical values. This commit adds a check via regex for this value and removes the variable if it is invalid as if it was not given.

The run_http_date() function tries to run

`difftime=$((HTTP_TIME + HTTP_AGE - NOW_TIME))`

When the Age header is missing bash will automatically substitute with 0, but we found it sometimes occurs that the age header contains an entire date like the Date header. This will lead to an invalid expression and the script will halt.

The fix is checking if the Age header (if it exists) is purely numerical. If it is not a new heading and message is added to the output (but not the file output) and the HTTP_AGE variable is unset. This prevents the error that halts the script and allows the script to continue as if the Age header was not given.

- [x] Bug fix
- [x] Improvement
- [ ] Typo fix
- [ ] Documentation update
- [ ] Update of other files

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -2768,6 +2768,14 @@ run_http_date() {
      if [[ ! -s $HEADERFILE ]]; then
           run_http_header "$1" || return 1
      fi
+
+     # make sure that HTTP_AGE is either missing or a valid value (only digits) for the later expression.
+     if [[ -n "$HTTP_AGE" && ! "$HTTP_AGE" =~ ^[0-9]+$ ]]; then
+          pr_bold " HTTP Age Header              "
+          out "Age header has invalid value: $HTTP_AGE (treated as 0)";
+          unset HTTP_AGE
+          outln
+     fi
      pr_bold " HTTP clock skew              "
      if [[ -n "$HTTP_TIME" ]]; then
           if "$HAS_OPENBSDDATE"; then


drwetter:
Good catch

testssl.sh
```diff
@@ -2768,6 +2768,14 @@ run_http_date() {
      if [[ ! -s $HEADERFILE ]]; then
           run_http_header "$1" || return 1
      fi
+
+     # make sure that HTTP_AGE is either missing or a valid value (only digits) for the later expression.
+     if [[ -n "$HTTP_AGE" && ! "$HTTP_AGE" =~ ^[0-9]+$ ]]; then
+          pr_bold " HTTP Age Header              "
+          out "Age header has invalid value: $HTTP_AGE (treated as 0)";
+          unset HTTP_AGE
```
drwetter, line 2776: unsetting HTTP_AGE is not equaly setting it to zero.

drwetter:
Not sure whether I like this fix. Probably it's better to set it to NaN (here or in  `run_http_header()`) and make sure the following statements can deal with that (see `difftime`).

You rather want me to fix that?

drwetter:
Thanks @Tristanhx , for the finding and the PR is much appreciated! 👍
👍
In the meantime I suggested a different fix though.

Issue #2083, opened 2022-01-10
Inconsistent / incorrect ROBOT attack results (2nd issue)


famzah:
I'm experiencing the same inconsistency in the results as in issue #1107, when repeatedly testing the same Debian server.

Here is an example for ports 465 (SSL-only) and 587 (STARTTLS):
```bash
# IP address of the server is 192.252.146.33

famzah@vbox64:~/testssl$ for i in {1..20} ; do ./testssl.sh --robot famzah.net:465|grep " ROBOT" ; done
 ROBOT                                     not vulnerable (OK)
 ROBOT                                     VULNERABLE (NOT ok)
 ROBOT                                     VULNERABLE (NOT ok) - weakly vulnerable as the attack would take too long

famzah@vbox64:~/testssl$ for i in {1..20} ; do ./testssl.sh --robot --starttls smtp famzah.net:587|grep " ROBOT" ; done
 ROBOT                                     not vulnerable (OK)
 ROBOT                                     VULNERABLE (NOT ok)
 ROBOT                                     VULNERABLE (NOT ok) - weakly vulnerable as the attack would take too long
```

I cannot reproduce this on port 25 (STARTTLS) where I consistently get "not vulnerable (OK)". This is the expected result as the server is using the patched OpenSSL by Debian. The server is running Debian Buster and OpenSSL is version "1.1.1d-0+deb10u7". Note: ports 465 & 587 are served by a different server implementation compared to the server for port 25. I cannot disclose more about the implementations.

You should be able to reproduce on your end easily. It takes a couple of attempts to get different results. Usually in less than 10 attempts you will encounter a different result.

It you can't reproduce, I have saved the screen output and the "/tmp" dir content for both "not vulnerable (OK)" and "VULNERABLE (NOT ok)" runs executed with `debug=3`. If you need the saved data, I will attach it here.

The question is -- Is "testssl.sh" encountering a bug and reports inconsistent results, or is it the server for ports 465 & 587 which does something weird and really is vulnerable.

Thank you.

---

Here is my environment:
```bash
famzah@vbox64:~/testssl$ ./testssl.sh -b 2>/dev/null | grep "from"
    testssl.sh       3.0.6 from https://testssl.sh/

famzah@vbox64:~/testssl$ ./testssl.sh -b 2>/dev/null | grep -A3 OpenSSL
 Using "OpenSSL 1.1.1f  31 Mar 2020" [~79 ciphers]
 on vbox64:/usr/bin/openssl
[...]

PR #2969, opened 2025-12-15
Mitigate inconsistent test results for ROBOT

drwetter:
As reported a longer while back in #2083 there were trailing bytes when receiving a TLS alert by the ROBOT check.

This PR corrects and thus normalizes the length of the TLS alert message to the correct value, supposed the length in the TLS alert is two bytes and it is an TLS alert.

Also this PR now uses a separate variable for the timeout. In 2ce0110e the timeout was changed by mistake as MAX_WAITSOCK was reduced from 10 to 5. For this check it is still 5 which seemed fine (TBC). Using a separate global variable however may offer some possibility for tuning the check when the latency to the target is high.

open:
- update the docs wrt ROBOT_TIMEOUT
- more checks

- [x] Bug fix
- [ ] Improvement
- [ ] Typo fix
- [ ] Documentation update
- [ ] Update of other files


drwetter:
This needs to be backported to 3.2

drwetter:
I merged this to mitigate the problem and move things forward. 

Not sure really about the timeout , @dcooper16 ? In practice and so far it seemed fine to me. The implementation though is not as easy to understand for me. If you have a minute or two, your input would be highly appreciated.

At a certain point of time I'd like to backport this to 3.2 . At that point though I'd like to be more certain whether I did the right thing.

PR #3049, opened 2026-06-06
Fix --mx host:port parsing and incorrect no-MX message (#2986)

### LINKED ISSUE #2986: confusing output
Note to self: e.g ``./testssl.sh --mx bund.de:25`` confuses the cmd line parser. It results in a message ``YYYYMMDD-HHMM  has no MX records(s)``

Observed so far in 3.3dev using Debian 13

potato-20:
Fixes #2986.

`--mx <host>:<port>` confused the parser: `--mx` sets `PORT=25` unconditionally, and the trailing `:port` on the domain was never stripped before the MX DNS lookup — so `host -t MX example.com:25` returned nothing, and the failure branch then printed `$1` (the run date) instead of the host, e.g. `20260606-1200 has no MX records(s)`.

This:
- strips a trailing `:<port>` off the domain before the `get_mx_record` lookup (bash parameter expansion, no extra fork) and uses it as the port under test;
- corrects the no-MX message to print the actual `$domain` and `MX record(s)` (typo).

Tested against multiple hosts with and without an appended `:port` (e.g. `--mx google.com:25` now resolves the MX and proceeds on port 25); plain `--mx example.com` behaviour is unchanged. `shellcheck -x -P utils --severity=error` clean and `bash -n` clean on the change; five-space/no-tabs convention followed.

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -23652,21 +23652,27 @@ draw_line() {
 
 run_mx_all_ips() {
      local fname_date="$1"
+     local domain="$2"
      local mxs mx
-     local mxport
+     local mxport=${3:-25}
      local -i ret=0
      local word=""
 
      STARTTLS_PROTOCOL="smtp"
+     # A port may be appended to the domain, e.g. "--mx example.com:587" (#2986).
+     # Strip it off before the MX DNS lookup and use it as the port to test.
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks!

PR #3060, opened 2026-06-11
Add HSTS preload list check via hstspreload.org API (--phone-out)

### LINKED ISSUE #1248: HSTS Browser Preloading Check ( Like SSL Labs )

potato-20:
## Describe your changes

Adds an HSTS preload list status check. When `--phone-out` is supplied, `run_hsts()` queries the hstspreload.org status API (`/api/v2/status`) for the tested host and reports whether the domain is `preloaded` / `pending` / `rejected` / `unknown`, combining that with the served-header state (marked-for-preload, same-domain, bulk) to produce an appropriate severity (OK / INFO / LOW / MEDIUM, or WARN if the API is unreachable).

Fixes #1248
Supersedes #1809 — this revives @tosticated's (Jim Blankendaal) original work, rebased onto current `3.3dev`. Since that PR is a draft on his fork it can't be pushed to, hence a fresh PR; credit is retained in `CREDITS.md`.

- [x] For the main program: my edits contain no tabs, indentation is five spaces and line endings contain no blank chars
- [x] I have tested this new feature against hosts that show it and hosts that don't
- [x] I made corresponding changes to the documentation and to `help()`
- [x] I added the contributor to CREDITS.md and the change to CHANGELOG.md

### Notes
- Addresses the review comments on #1809: the API-response matching is now pure bash (`[[ ... == *...* ]]`, no `grep` fork), the JSON quoting is handled inside `check_hsts_preloadlist_match()` so callers pass plain values, and the value arrays use `local -a`. The output decision table is kept as-is (per your feedback on #1809).
- Only known status/bulk values are matched against the API response, so no untrusted API input is reflected back to the user.
- Unit test added: `t/53_hsts_preload.t` (github.com listed → `preloaded`, example.com not → `unknown`), modeled on `t/52_ocsp_revoked.t`, core `Test::More` only.
- The rendered man pages `doc/testssl.1` and `doc/testssl.1.html` are **not** regenerated here (needs pandoc) — only the `doc/testssl.1.md` source is updated. Could you regenerate them with `make -C doc`?

Credit for the original implementation and the decision table goes to @tosticated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@
 * DNS via proxy improvements, also IPv6 support for proxy
 * Client simulation runs in wide mode which is even better readable
 * Added --reqheader to support custom headers in HTTP requests
+* `--phone-out` checks the HSTS preload list on https://hstspreload.org/
 * Deprecating --fast and --ssl-native (warning only but still av)
 * Compatible to GNU grep >=3.8, bash 5.x
 * Don't use external pwd command anymore
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -55,6 +55,7 @@ Full contribution, see git log.
   - maximum certificate lifespan of 398 days
   - ssl renegotiation amount variable
   - custom http request headers
+  - HSTS preload list lookup
 
 * Frank Breedijk
   - Detection of insecure redirects
--- a/doc/testssl.1.md
+++ b/doc/testssl.1.md
@@ -152,7 +152,7 @@ The same can be achieved by setting the environment variable `WARNINGS`.
 
 `--ids-friendly` is a switch which may help to get a scan finished which otherwise would be blocked by a server side IDS. This switch skips tests for the following vulnerabilities: Heartbleed, CCS Injection, Ticketbleed and ROBOT. The environment variable OFFENSIVE set to false will achieve the same result. Please be advised that as an alternative or as a general approach you can try to apply evasion techniques by changing the variables USLEEP_SND and / or USLEEP_REC and maybe MAX_WAITSOCK.
 
-`--phone-out` Checking for revoked certificates via CRL and OCSP is not done per default. This switch instructs testssl.sh to query external -- in a sense of the current run -- URIs. By using this switch you acknowledge that the check might have privacy issues, a download of several megabytes (CRL file) may happen and there may be network connectivity problems while contacting the endpoint which testssl.sh doesn't handle. PHONE_OUT is the environment variable for this which needs to be set to true if you want this.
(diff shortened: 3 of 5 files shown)


drwetter:
hi @potato-20 ,

first: thanks for reviving this! And awesome you amended that with a unit test. 

Quick shot: 
- yes, can run pandoc later. There are some glitches anyway, as I have to do that manually ---either on MacOS or Linux (and I always forgot which is the right one) as otherwise there will be a huge diff. That should be automated by github preferably *) . For a line though I often end up manually edit the HTML and roff file.
- UI: One thing which is incomprehensible to the user is "unknown" which comes directly from the API . Off the top of my head don't know what the right wording could be. Maybe if we change `HSTS preload list` to `HSTS preload API` we could use  something like  `no entry (=unknown) ` or just `no entry`. 

More later.



*) there was a discussion here but atm I cannot find it.

drwetter:
PS: if you rebase, the `.1` and `.html` file will be autogenerated

potato-20:
Fixed — replaced `"unknown"` with `"no entry"` (your suggestion), keeping the section header as `HSTS preload list` since the rename wasn't strictly needed to fix the comprehension issue. Also rebased onto current `3.3dev`; the man pages will auto-generate via #3064. Unit test updated to match.

drwetter:
Sigh, a day or two days ago I commented on this but I obviously didn't push the button . TL:DR : I'll merge it to a different branch and change some things. For some reason the pandoc autogeneration did not fly.

drwetter:
... continued in #3071 .

For the autorun of pandoc: For some reason it was skipped here. 

image


For me it looked like:

image

As the autorun was ready I got an e-mail.


Thus closing and waiting for the GH runners in #3071 to succeed. Thanks for reviving the stale PR! 👍

Issue #3061, opened 2026-06-15
[BUG / possible BUG] Huge delay on WSL (Arch and also NixOS)


lapo-luchini:
- [x] I am running version `e0af44a7be4c02da86c0f6e93ad9008841fc02b7` from the git repo 
- [ ] I am running ``testssl.sh -v | grep from``

**Before you open an issue please consult the [FAQ](https://github.com/testssl/testssl.sh/blob/3.3dev/FAQ.md) and check whether this is a known problem by searching the  [issues](https://github.com/testssl/testssl.sh/issues)**

- [x] couldn't find anything

**Command line to reproduce (or docker command)**

```
testssl.sh google.com
```

and in other console:

```
% ps auxf | grep -A20 testssl
lapo       76824  1.0  0.0  16048 12160 pts/5    S+   14:02   0:00  |           \_ bash /home/lapo/bin/testssl.sh google.com
lapo       76934  0.0  0.0  14796  9472 pts/5    S+   14:02   0:00  |               \_ /usr/sbin/openssl s_client -tls1
lapo       76935  0.0  0.0   6812  2432 pts/5    S+   14:02   0:00  |               \_ grep -aiq unknown option
```

(but it also happens with `testssl.sh -v` really!)

**Expected behavior**

An immediate response to `testssl.sh -v` and even when reporting remote websites, the header should appear pretty soon. (then the complete report can take the time it takes)

**Your system (please complete the following information):**
 - OS: `Arch Linux` (under Windows 11 WSL), but same on `NixOS 25.11 (Xantusia)` (WSL)
 - Platform: `Linux 6.6.87.2-microsoft-standard-WSL2 x86_64`
 - OpenSSL + bash: unknown (takes a looong time also to do `testssl.sh -b`

**Additional context**

I am willing to debug the issue… I already tried `DEBUG=5`, or `bash -vx` (it stops after redirecting stderr to 42), or `strace` but I failed to undestand th eproblem.

drwetter:
Hi @lapo-luchini ,

as the doc says you can run `./testssl.sh --debug NUMBER TARGET` then you see approximately where it's hanging. Sluggish upon `-b` / `-v` is rather strange. 

You can also prepend `SETX=true bash -x ./testssl.sh COMMAND` but there's more output.  Maybe a start for that is  `SETX=true bash -x ./testssl.sh -v`.

Can you run `time ./testssl.sh -v` and report back how long it takes?

lapo-luchini:
[...]

PR #3070, opened 2026-06-19
Avoid potentially long timeouts

lapo-luchini:
## Describe your changes

I'm adding a "timeout" guard to avoid excessive timeouts when localhost is stopping TCP RST packets, such as on WSL2 in mirrored mode. This is only used if detected and should never break existing systems as it only put a cap on wait time, but systems that have a faster results are effectively untouched.

Closes GitHub #3061.

- [x] Bug fix

### DIFF
--- a/testssl.sh
+++ b/testssl.sh
@@ -21323,8 +21323,8 @@ find_openssl_binary() {
 
      # In order to avoid delays due to lookups of the hostname "invalid." we just try to avoid using "-connect invalid."
      # when possible. The following does a check fopr that. For WSL we stick for now to the old scheme. Not sure about Cygwin
-     if [[ SYSTEM2 == "WSL" ]]; then
-          NXCONNECT=-connect $NXDNS
+     if [[ $SYSTEM2 == "WSL" ]]; then
+          NXCONNECT="-connect $NXDNS"
      else
           # If this connects and bails out with an error message, we do not need "-connect invalid."
           if $OPENSSL s_client 2>&1 </dev/null | grep -Eiaq 'Connection refused|connect error|Bad file descriptor'; then
--- a/testssl.sh
+++ b/testssl.sh
@@ -21153,6 +21153,7 @@ find_openssl_binary() {
      local openssl_location="" cwd=""
      local curve="" ossl_tls13_supported_curves
      local ossl_line1="" yr=""
+     local ossl_probe_timeout=""
      # FIXME: At the moment curves_ossl does not include any post-quantum key-exchange
      # groups (e.g., MLKEM512, MLKEM768, MLKEM1024, SecP256r1MLKEM768, X25519MLKEM768,
      # SecP384r1MLKEM1024, curveSM2MLKEM768). They do not need to be included since they are only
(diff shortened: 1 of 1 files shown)


drwetter:
Thanks for the PR!

However: It won't work as setting the variables will either all be set or unset which doesn't make sense. That's why I'd rather close this.

The thing with the missing variable reference (`SYSTEM` instead of `$SYSTEM`(: nice catch! 👍. Probably 3.2 has the same problem. Mind changing that for both branches?

And maybe amend the check `elif [[ -e "/mnt/c/Windows/System32/bash.exe" ]] ` like  `elif ( [[ -e "/mnt/c/Windows/System32/bash.exe" ]]  || grep -qi microsoft /proc/version ) ` like you suggested?

lapo-luchini:
> It won't work as setting the variables will either all be set or unset which doesn't make sense.

I don't understand this part, sorry. The checks are meant to be just a before… just limited in maximum time spent. But going to a timeout always means that the connection was attempted, and thus the output will never have "unknown option" (which happens instantaneously, before the connection is even attempted), so the results should be just as (individually) valid as before. 🤔

drwetter:
> so the results should be just as (individually) valid as before. 🤔

yes, sarcastically speaking they show all true (or maybe false?) which is for further tests "not really useful".

lapo-luchini:
I can't see the problem on my side. 🤔

Just to b sure I added locally some debug like this:

```sh
     $ossl_probe_timeout $OPENSSL s_client -ssl2 </dev/null 2>&1 | grep -aiq "unknown option" || HAS_SSL2=true
     $ossl_probe_timeout $OPENSSL s_client -ssl3 </dev/null 2>&1 | grep -aiq "unknown option" || HAS_SSL3=true
     $ossl_probe_timeout $OPENSSL s_client -tls1 </dev/null 2>&1 | grep -aiq "unknown option" || HAS_TLS1=true
     $ossl_probe_timeout $OPENSSL s_client -tls1_1 </dev/null 2>&1 | grep -aiq "unknown option" || HAS_TLS11=true
     $ossl_probe_timeout $OPENSSL s_client -tls1_2 </dev/null 2>&1 | grep -aiq "unknown option" || HAS_TLS12=true
     $ossl_probe_timeout $OPENSSL s_client -tls1_3 </dev/null 2>&1 | grep -aiq "unknown option" || HAS_TLS13=true
     $ossl_probe_timeout $OPENSSL s_client -no_ssl2 </dev/null 2>&1 | grep -aiq "unknown option" || HAS_NO_SSL2=true

     for n in HAS_SSL2 HAS_SSL3 HAS_TLS1 HAS_TLS11 HAS_TLS12 HAS_TLS13 HAS_NO_SSL2; do
          echo "$n ${!n}"
     done
```

and the output seems fine:

```
% ./testssl.sh -v

HAS_SSL2 false
HAS_SSL3 false
HAS_TLS1 true
HAS_TLS11 true
HAS_TLS12 true
HAS_TLS13 true
HAS_NO_SSL2 false

#####################################################################
  testssl.sh version 3.3dev from https://testssl.sh/dev/
  (9aeaede 2026-06-19 21:31:59)

  This program is free software. Distribution and modification under
  GPLv2 permitted. USAGE w/o ANY WARRANTY. USE IT AT YOUR OWN RISK!

#####################################################################

  Using OpenSSL 3.6.2 (Apr 20 2026)  [~69 ciphers]
  on cyberone-wsl:/usr/sbin/openssl
  built: Mon Apr 20 16:11:24 2026, platform: linux-x86_64
  Using bash 5.3.12
```

drwetter:
🤔

PR #3104, opened 2026-07-15
Improve handling of short-lived certificates (#3097)

ericcgu:
Closes #3097. Implements the short-lived certificate handling agreed in the issue thread with @drwetter and @petecooper.

## What is your pull request about?
- [ ] Bug fix
- [x] Improvement
- [ ] New feature (adds functionality)
- [ ] Breaking change
- [ ] Typo / spelling fix
- [ ] Documentation update
- [ ] Update of other files

## Problem
Certificates with a short validity period (e.g. Let's Encrypt's 6-day "shortlived" profile, now GA) always trip the `days2warn` expiry thresholds and are shown in **red** on the "Certificate Validity (UTC)" line, even seconds after issuance. That red is misleading for a certificate that is *intentionally* short-lived.

## Change
Detect short-lived certificates by their validity period (`notAfter - notBefore`) using a new `DAYS_VALID_SHORTLIVED` threshold (default **10 days**, per the CA/Browser Forum BR §1.6.1 "Short-lived Subscriber Certificate" definition, which also covers the LE 6-day profile). For such certs:

- show them as **good** (not red) with a `short-lived cert (N days)` remark so the short lifespan reads as intended rather than as an error, and
- only warn (**HIGH**) when **less than 24h** of validity is left, and only for certificates whose total lifetime exceeds 24h (a cert whose whole life is under 24h stays good until it expires rather than being flagged the entire time).

Regular certificates are unaffected and keep the existing `days2warn` behaviour. The threshold is overridable via the `DAYS_VALID_SHORTLIVED` env var, like `DAYS2WARN1/2`.

Per @drwetter's "step by step" request, this PR only handles the short-lived case. The broader per-profile threshold reconsideration (45-day `tlsserver`, upcoming 64-day/45-day `classic`, etc.) is intentionally left out for a separate discussion/issue.

## Testing
Verified end-to-end with local `openssl s_server`:

| Cert | Terminal output | JSON severity |
|------|-----------------|---------------|
| 5-day validity, ~5 days left | `short-lived cert (4 days)` (green) | `OK` |
| ~5-day validity, < 24h left | `short-lived cert, expires < 24h` (red) | `HIGH` |
| 12h validity, freshly issued | `short-lived cert (0 days)` (green) | `OK` |
| 12h validity, ~1h left | `short-lived cert (0 days)` (green) | `OK` |
| 90-day validity (regression) | `89 >= 60 days` (unchanged) | `OK` |

[...]

Issue #3097, opened 2026-07-13
[Feature request] Improve handling for short-lived certificates


petecooper:
**Please check this repo whether this is a known feature request**

None that I have found.

**Describe your feature request (if it's a technical feature)**

Improved handling of known short-lived certificates. 160-hour certificates from Let's Encrypt are GA:

https://letsencrypt.org/2026/01/15/6day-and-ip-general-availability

**If your feature request otherwise is related to a usage problem, please describe it**

Red (warning) text shown when a valid short-lived (e.g. 160-hour) certificate is tested:

<img width="951" height="575" alt="Image" src="https://github.com/user-attachments/assets/eb920647-caf5-4b71-b828-78f845054434" />

**Describe the solution you'd like**

* Detection of short-lived certificate.
* Adjust 'Certificate Validity (UTC)' to be less jarring if short-lived certificate is found.

**Which version of testssl.sh are you referring to?**

`testssl.sh version 3.2.4 from https://testssl.sh/`

ericcgu:
Thanks @petecooper 

1. Would you kindly clarify what exactly you mean by:
"Adjust 'Certificate Validity (UTC)' to be _**less jarring**_ if short-lived certificate is found."
Is this a removal of red highlight for short lived?

2. Is the flag or definition of short-lived **10** days or fewer?

@drwetter If approved, I can work on this.

petecooper:
> Is this a removal of red highlight for short lived?

Essentially, yes. If the renewal period is active for a cert of any lifespan, then a warning could be shown. Red text for a valid short-lived cert - in the example above it's literally minutes old - is visually jarring, and a presentational tweak would be greatly appreciated.

>Is the flag or definition of short-lived 10 days or fewer?

In the case of Let's Encrypt, it's 160-hour (or 6.67 days).

Thank you.

drwetter:
@ericcgu : sure, PR's are always welcome.

Maybe first we need to discuss how to flag which certificates. Atm I don't know whether the 6 day certificate is the only one which has a shorter time span than the 90 days from LE or the 200 days for ~everything else?

Suggestion:
* <24 h for 6 day certificate : red (maybe with a remark that it is a 6day cert, so that the user gets feedback it's propaby intended)
* everything else: can be reconsidered when you feel it would help

OK?

petecooper:
LE have [profiles](https://letsencrypt.org/docs/profiles/):

* "classic" is 90-day
* "tlsserver" is 45-day
[...]

PR #3106, opened 2026-07-15
Improve handling of short-lived certificates (#3097) [3.2 backport]

ericcgu:
Backport of the short-lived certificate handling to the **3.2** branch. The feature request (#3097) was filed against 3.2.4, so this brings the fix to the stable branch. The 3.3dev version is #3104.

## Problem
Certificates with a short validity period (e.g. Let's Encrypt's 6-day "shortlived" profile, now GA) always tripped the `days2warn` expiry thresholds and were shown in **red** on the "Certificate Validity (UTC)" line, even seconds after issuance. That red is misleading for a cert that is intentionally short-lived.

## Change
Detect short-lived certificates by their validity period (`notAfter - notBefore`) using a new `DAYS_VALID_SHORTLIVED` threshold (default **10 days**, per the CA/Browser Forum BR 1.6.1 "Short-lived Subscriber Certificate" definition, which also covers the LE 6-day profile). For such certs:

- show them as **good** (not red) with a `short-lived cert (N days)` remark, and
- warn (**HIGH**) only when less than 24h of validity is left, and only for certificates whose total lifetime exceeds 24h (a cert whose whole life is under 24h stays good until it expires rather than being flagged the entire time).

Regular certificates are unaffected and keep the existing `days2warn` behaviour. The threshold is overridable via the `DAYS_VALID_SHORTLIVED` env var (documented in `doc/testssl.1.md` next to `DAYS2WARN1/2`).

## Notes for reviewers
- Code is functionally identical to the 3.3dev change (commit 26a75cf7).
- The IPv6-proxy fix from that commit is **not** included here -- it already landed on 3.2 via #3101.
- 3.2's man/HTML docs are **Ronn-NG** generated (not pandoc). I hand-inserted the new entry into `doc/testssl.1` and `doc/testssl.1.html` to match that format rather than regenerating, which would reflow the whole document. Feel free to regenerate with the proper toolchain.
- `CHANGELOG.md` entry placed under "Features implemented / improvements in 3.2" without a version tag -- adjust if you'd prefer a `(3.2.x)` tag.


drwetter:
Thanks again, @ericcgu !

drwetter:
> The IPv6-proxy fix from that commit is not included here

That's appreciated! We'll try to address one thing at a time.

> 3.2's man/HTML docs are Ronn-NG generated (not pandoc)

Sigh, yes. Thank you. I believe the origin was an older Pandoc version on Linux. If we settle on Tool X and Version >=Y, we'll probably need to reformat that, once.

> CHANGELOG.md entry placed under "Features implemented / improvements in 3.2" without a version tag

That's fine. We didn't bother before. For this version we should start doing that, for 3.4 we could consider it

PR #3144, opened 2026-09-14
Avoid use of color functions

dcooper16:
`run_freak()` and `run_logjam()` were using the color functions `$magenta` and `$off` (from `set_color_functions()`) in the string `$addtl_warning`. This worked okay for the terminal output, but not for the HTML output or `fileout()`.

This PR fixes `run_freak()` and `run_logjam()` by using `pr_warning()` for the terminal and HTML output and plaintext for `fileout()`.

This PR also deletes the unused color functions from `set_color_functions()` and adds a note discouraging the use of the color functions that are still defined there.

- [x] Bug fix
- [ ] Improvement
- [ ] Breaking change: bug fix, feature or improvement that would cause existing output (especially JSON, CSV) to not work as expected before
- [ ] Typo / spelling fix
- [ ] Documentation update
- [ ] Update of other files

- [x] I've read [CONTRIBUTING.md](https://github.com/testssl/testssl.sh/blob/3.3dev/CONTRIBUTING.md)
- [x] My code follows [Coding_Convention.md](https://github.com/testssl/testssl.sh/blob/3.3dev/Coding_Convention.md) 
- [ ] If it's a bigger change: I added myself to [CREDITS.md](https://github.com/testssl/testssl.sh/blob/3.3dev/CREDITS.md) (alphabetical order of last name) and the change to [CHANGELOG.md](https://github.com/testssl/testssl.sh/blob/3.3dev/CHANGELOG.md)

## AI section
- [ ] I found a bug / an improvement using LLM version: `[e.g. GPT-A.B, Claude <NAME> A.B, Gemini A.B <NAME>, Qwen<B>-Coder, DeepSeek-<A>  etc.]`
- [x] My contribution does not include any AI-generated content
- [ ] My contribution includes AI-generated content, as disclosed below:
    - AI Tools: `[e.g. GitHub CoPilot, JetBrains Junie, VS Code plugin <NAME> etc.]`
    - LLMs and versions: `[e.g. GPT-A.B, Claude <NAME> A.B, Gemini A.B <NAME>, Qwen<B>-Coder, DeepSeek-<A>  etc.]`


drwetter:
Hi David,
thanks!
I agree with you that those functions / definitions should be avoided, as for general output we should  rather use the colorize functions which includes HTML output.

However 
- after applying the patch there's still `brown` and `yellow` defined (and `off`). Would those be the ones which are still in use after this patch?
- I am not sure whether we should remove the definitions completely because we could use that for debug purposes. or any other output which won't make it to a file , like throwing an cmdline error during interactive use or whatsoever.


Remark: The additional warnings in `run_logjam()` and `run_freak()` : good workaround . In the long run, should be still support `--ssl-native` ?

dcooper16:
>   * after applying the patch there's still `brown` and `yellow` defined (and `off`). Would those be the ones which are still in use after this patch?

I wrote this PR before working on a rewrite of `emphasize_stuff_in_headers()`. Now that #3145 has been merged, `brown`, `yellow`, and `off` are no longer needed. So, I updated this PR to remove them.

Now `set_color_functions()` just sets `TERM=xterm` for OpenBSD when `TERM` is `xterm-256`. I don't know whether it is still needed.

>   * I am not sure whether we should remove the definitions completely because we could use that for debug purposes. or any other output which won't make it to a file , like throwing an cmdline error during interactive use or whatsoever.

We could keep them, if you think they might be needed. However, the `tm_` functions (e.g., `tm_warning()`) seem better for sending output just to the terminal.

> Remark: The additional warnings in `run_logjam()` and `run_freak()` : good workaround . In the long run, should be still support `--ssl-native` ?

I would be okay with getting rid of `--ssl-native`. If I understand correct, it was originally for testing against systems that worked with OpenSSL, but not `tls_sockets()`. However, if I remember correctly, testssl.sh won't really work anymore if `tls_sockets()` can not connect. So, perhaps there is no longer a reason for `--ssl-native`.

drwetter:
I was just about to argue before your comment came in.

> We could keep them, if you think they might be needed. However, the tm_ functions (e.g., tm_warning()) seem better for sending output just to the terminal

Ok, makes sense.

Wrt OpenBSD: I'll check

drwetter:
> Wrt OpenBSD: I'll check

That seems to be safe to remove as well as long as we don't make use of tput and friends. 

I got an errors like this already @ startup:

```
tput: not enough arguments (3) for capability `setaf'
tput: not enough arguments (3) for capability `setaf'
tput: not enough arguments (3) for capability `setaf'
tput: not enough arguments (3) for capability `setaf'
tput: not enough arguments (3) for capability `setaf'
tput: not enough arguments (3) for capability `setaf'
tput: not enough arguments (3) for capability `setaf'
tput: not enough arguments (3) for capability `setaf'
```

if I use the `setaf` definitions and set `XTERM`  not to just `xterm` . This seems to be a [known problem](https://daemonforums.org/showthread.php?t=11594) I forgot. 

Looks like another argument for `tm_*()` functions.

